PR #22, opened 2012-11-29
Add collapse, interleave/interleave_longest, append/prepend.

abarnert:
Five new functions, all inspired by StackOverflow questions.

`collapse(iterable, basetype=None, levels=None)`

A multi-level `flatten`. Collapse a hierarchical iterator all the way down to strings and non-iterables (and instances of `basetype`, if given, which can be a type or a tuple), or down `levels` levels. People on SO keep asking "how do I flatten up to N levels instead of exactly N", or "how do I flatten iterators but not lists", or "why does my flatten screw up my strings". Rather than writing separate solutions for each, this is an all-singing-all-dancing-flatten-everything implementation.

`interleave(*iterables)` and `interleave_longest(*iterables)`

Like `flatten(zip(*iterables))` and `flatten(zip_longest(*iterables))`, except that `iterate_longest` skips missing entries instead of filling with None.

`append(iterable, value)` and `prepend(iterable, value)`

Yields from iterable, then yields value, or vice-versa. Only worth including because I tested every way I could think of to implement this, and using `chain(iterable, [value])` is by far the most efficient, which surprised someone on SO who kept arguing that it was a huge waste to build a length-1 list.

Unlike my previous pull requests, this time I found your unit tests and added tests for them. (But the "python3 setup.py test" doesn't work, because it tries to test the non-2to3 version.) I didn't go back and write tests for my other patches, because I'd have to cancel my pull requests, patch up each branch separately, submit new pull requests, and re-merge my fork's master, but if you want me to, I can.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -1,8 +1,9 @@
 from functools import partial, wraps
-from itertools import izip_longest, ifilter
+from itertools import chain, izip, izip_longest, ifilter
 from recipes import *
 
-__all__ = ['chunked', 'first', 'peekable', 'collate', 'consumer', 'ilen']
+__all__ = ['chunked', 'first', 'peekable', 'collate', 'consumer', 'ilen',
+           'append', 'prepend', 'interleave', 'interleave_longest', 'collapse']
 
 
 _marker = object()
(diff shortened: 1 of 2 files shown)


erikrose:
@bbayles FWIW, I've come around and am enthusiastic about including `append` and `prepend`. Thanks again for cleaning house!

bbayles:
Hmm, I'm good with the interleaves and probably `collapse`. But `append` and `prepend` seem less flexible than just using `chain(['whatever', 'i', 'want'], iterable)` and `chain(iterable, ['not', 'limited', 'to', 'one'])`?

erikrose:
Good point. It saves a pair if brackets, but it's not worth the loss of API compactness.

bbayles:
Superseded by PR #74; the saga continues there.

PR #24, opened 2013-04-07
enhanced chunked

xpysol:
I've found this recipe at python-ideas, where it was suggested by Peter Otten. You may want to time its performance against your current version of chunked(), but I think it should be faster. Also, Wolfgang Maier's proposal of the strict parameter to turn on error handling looks like a good idea to me.
This is the original thread at python-ideas:
http://thread.gmane.org/gmane.comp.python.ideas/20113


bbayles:
@xpysol: It's a long time later, but I think I'll close and decline this one. #47 simplified the implementation of `chunked`, and the `slide` function requires the length of the iterable to be known.

PR #25, opened 2013-05-10
flatten for n-nested lists

roryk:
I can't claim credit for this, I snagged it from StackOverflow a long time ago and use it all of the time.


bbayles:
My preference is for #22 over this one, so I'll close out this PR. Even so, many thanks.

roryk:
Thanks!

PR #26, opened 2013-06-11
New itertool: ipartition

Rosuav:
Not sure if this would be of interest to you, but it's something I knocked together in response to a python-list question and Tim Chase said it might be of use. It looks like you're using the MIT license? I'm okay with that, if it is.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -1,9 +1,10 @@
 from functools import partial, wraps
 from itertools import izip_longest
+from collections import deque
 from recipes import *
 
 __all__ = ['chunked', 'first', 'peekable', 'collate', 'consumer', 'ilen',
-           'iterate', 'with_iter']
+           'iterate', 'with_iter', 'ipartition']
 
 
 _marker = object()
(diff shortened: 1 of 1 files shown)


erikrose:
I like the idea. I'd like to…

1. Code it with less repetition and
2. See if it can be readily expanded to an arbitrary number of partitions, as I was thinking of adding a bucket() function that did that anyway. (You'd pass in a `key` function that returned any number of bucket identifiers, and `partition` (or `bucket`, which would be a better name for the larger version) might become indexable so you could get at all the buckets: `buckets = bucket(bool, iterable); print next(buckets[true])`.)

Crazy?

bbayles:
I like the `dict` in @cvrebert's version, but was planning on playing with it to see if we can avoid pulling the iterable into memory.

bbayles:
I think I've got something that could work in 054f4c79871ec6441644e41d2697941116b8cdad; I'll play some more and make a PR with any changes.

bbayles:
I'll close this issue in favor of #66, which has tests and such.

PR #27, opened 2013-06-18
Add the `sift()` itertool

nvie:
Sift creates two lazy subiterators that together consume the input.

```
>>> even, odd = sift(xrange(10), key=lambda x: x % 2 == 0)
>>> list(even)
[0, 2, 4, 6, 8]
>>> list(odd)
[1, 3, 5, 7, 9]
```

It's more efficient than looping over the input twice with two inverse filters.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -14,6 +14,7 @@ New Routines
 .. autofunction:: distinct_permutations
 .. autofunction:: first(iterable[, default])
 .. autofunction:: ilen
+.. autofunction:: intersperse
 .. autofunction:: iterate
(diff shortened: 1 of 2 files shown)


erikrose:
@bbayles Something weird may have happaned with this merge. I'm not sure if you meant to get sift() into the codebase or not, but it doesn't look like it landed there. I flag it in case this is a surprise for you. :-) Cheers!

bbayles:
Odd indeed! I didn't intend to merge in `sift`, and it looks like it's not there. I'll make an issue to re-examine it.

PR #37, opened 2014-04-06
new function - always iterable

jaraco:
The `always_iterable` function is one of my favorite functions. I use it time and time again in unexpected scenarios.

Often, when a parameter is allowed to be a string or None, but also an iterable, one has to set up a series of tests and transformation. Using always_iterable means the logic can be greatly simplified and more intuitive, simply iterating over the (possibly empty) result.

I've tried to communicate this value in the docstring.

I've also back-ported it to support Python 2.7 first without six.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -8,6 +8,7 @@ API Reference
 New Routines
 ============
 
+.. autofunction:: always_iterable
 .. autofunction:: chunked
 .. autofunction:: collate(*iterables, key=lambda a: a, reverse=False)
 .. autofunction:: consumer
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -252,8 +252,8 @@ def always_iterable(item):
     >>> always_iterable(None)
     ()
 
-    >>> always_iterable(range(10))
-    range(0, 10)
+    >>> always_iterable(xrange(10))
+    xrange(10)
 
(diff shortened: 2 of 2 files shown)


bbayles:
@jaraco, I'd like to resurrect this. With your blessing I'll make a new PR against the more recent master branch, and credit you in the next release?

jaraco:
Awesome. I'd be delighted.

pylang:
I notice `string_types = unicode, bytes` throws a `NameError` in Python 3.  This seems to work in Python 3.5 and 2.7:

```python
from collections import Iterable

def always_iterable(item):
    if item is None:
        item = ()
#     string_types = str, bytes
#     if isinstance(item, string_types) or not hasattr(item, '__iter__'):
    if isinstance(item, Iterable):
        item = item,
    return item

```

Tests

```python
assert always_iterable("foo") == ('foo',)
assert always_iterable(u"foo") == ('foo',)
assert always_iterable(b"foo") == (b'foo',)
```

REF http://stackoverflow.com/a/7197882/4531270

jaraco:
The [reference implementation](https://github.com/jaraco/jaraco.itertools/blob/master/jaraco/itertools.py#L832-L863) uses [six to detect the string types](https://github.com/benjaminp/six/blob/master/six.py#L41-L49).

PR #40, opened 2014-07-06
Better Cartesian product support for big data.

a904guy:
Adds two functions:

total_product(): Calculate total permutations of a Cartesian product.
Used to get an integer for iteration over with Cartesian
specific_product()

specific_product(): Extract single permutation of Cartesian product
iteration in an consistent speed regardless of list/set sizes. 
This is better then the current intertools.product() generator/yield when dealing with
(big data research) massive amounts of sets/permutations.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -18,6 +18,8 @@ New Routines
 .. autofunction:: one
 .. autoclass:: peekable
 .. autofunction:: with_iter
+.. autofunction:: total_product
+.. autofunction:: specific_product
 
 
 Itertools Recipes
--- a/docs/versions.rst
+++ b/docs/versions.rst
@@ -2,6 +2,9 @@
 Version History
 ===============
 
+2.4
+    * Added ``total_product`` and ``specific_product``. (Thanks, Andy Hawkins of BombSquad!)
+
 2.3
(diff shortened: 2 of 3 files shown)


bbayles:
@a904guy, I'm leaning toward closing this. If you provide a motivating example (why would I want the 40,000th lexicographic arrangement?) I could be persuaded to change my mind. Also, the terminology seems a bit confusing - "permutation" and "product" are both used in the description.

a904guy:
Simple, it's a fast quick way to pull the 34,921 arrangement without having to compile the entire 40k permutations into memory, so that you can resume processes on large data sets from the last nth piece of data, or that individual pieces can be given out to task managers or job launchers. 

As far as the terminology, it can be called whatever, considering product is a different methodology, I'd agree it needs better wording but to say there is no practical reason for a low resource quick way to pluck from a large data set is a bit presumptuous.

bbayles:
Thanks for writing. Your example helps - this is for being able to split up tasks that operate on data sets that result from a cartesian product of inputs. It's not for Project Euler-like [puzzles](https://projecteuler.net/problem=24), which is what I had thought of when I read your code.

PR #44, opened 2014-07-31
Add unique_from_each() function

bbayles:
Some months after answering [this](http://stackoverflow.com/questions/17035577/unique-features-between-multiple-lists/17035741#17035741) question at StackOverflow, I had a need for a function that would give me the unique items from a list of lists.

Here are some examples:

``` python
>>> unique_from_each("mississippi", "missouri")
[['p', 'p'], ['o', 'u', 'r']]
>>> unique_from_each("mississippi", "missouri", "russian")
[['p', 'p'], ['o'], ['a', 'n']]
```

`mississippi` and `missouri` both have an `m`, some instances of `i`, and some instances of `s`. The former doesn't have an `o`, a `u`, or an `r`, though. And the latter doesn't have a `p`, let alone two.

If we add `russian` to the mix, there's no additional overlap with `mississippi`, but it shares `r` and `u` with `missouri`. It's the only one with an `n`.

Other notes:
- For my application I needed order and duplicates to be preserved
- This will exhaust any iterators given to it, but that's typical in itertools
- This does a linear search of the concatenation of all the input iterables; could make that a set to speed up searching for long inputs if desired

It seems itertools-y, so I'd like to submit it to be considered for more-itertools.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -18,7 +18,7 @@ New Routines
 .. autofunction:: iterate
 .. autofunction:: one
 .. autoclass:: peekable
-.. autofunction:: unique_from_each
+.. autofunction:: unique_to_each
 .. autofunction:: with_iter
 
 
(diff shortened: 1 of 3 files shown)


bbayles:
I think I still like this addition. Open to alternate implementations, but we'll try this one.

PR #49, opened 2014-09-02
Fix: variable `p` overrides previous definition

nvie:
In Python 2.x, the loop variable from a list comprehension bleeds into the calling context, overriding local `p` in this case.


bbayles:
Fixed in #59 - thanks, @nvie!

PR #50, opened 2014-09-02
Add new itertool: `side_effect()`

nvie:
Passes through the iterable, but invokes a function for each item iterated over.

Can be useful for lazy counting, logging, collecting data, or updating progress bars, anything un-pure.

`fn` must be a function that takes a single argument.  Its return value will be discarded.

Examples:

```
>>> from more_itertools import take, consume, side_effect
>>> iterable = range(10)
>>> iterable = side_effect(print, iterable)
>>> consume(take(3, iterable))
0
1
2
```

Example for collecting data as it streams through:

```
>>> iterable = range(10)
>>> c = set()
>>> iterable = side_effect(c.add, iterable)
>>> consume(take(3, iterable))
>>> sorted(c)
[0, 1, 2]
```

**EDIT:** I've updated the function to support emitting side effects for chunks. This is useful because you can still process the result as a continuous stream, making chaining much easier.  It basically saves you from the ceremony of otherwise having to do this:

```
>>> iterable = range(100)
>>> iterable = chunked(iterable, 30)
>>> iterable = side_effect(do_something_with_chunk, iterable)
>>> iterable = flatten(iterable)
>>> list(iterable)
[0, 1, 2, 3, ..., 99, 100]
```

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -18,6 +18,7 @@ New Routines
 .. autofunction:: iterate
 .. autofunction:: one
 .. autoclass:: peekable
+.. autofunction:: side_effect
 .. autofunction:: with_iter
 
 
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -336,7 +336,7 @@ def intersperse(e, iterable):
     raise StopIteration
 
 
-def side_effect(fn, iterable):
+def side_effect(fn, iterable, chunk_size=None):
     """
     Passes through the iterable, but invokes a function for each item
     iterated over.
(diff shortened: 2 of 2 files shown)


erikrose:
I like this idea, especially for debugging. Do we need to integrate chunked() into it, though, or can we simply compose it with chunked, like `side_effect(print, chunked(iterable))`?

bbayles:
I'm positive on this one also; agree re: compose with `chunked`. I figured I'd try to write some tests and see how it feels.

nvie:
Yeah, I liked the idea of using composition to achieve the chunking, but you also have to keep in mind that this changes the structure of the iterable. So the proper way of composing a side-effect that does not change its input will have to look like this:

```python
iterable = flatten(side_effect(print, chunked(iterable)))
```

I found that to be a bit too verbose to be useful/desirable in practice, but I get the argument/pushback. Other ideas welcome :)

erikrose:
Good point. You're right—that's unpalatable.

bbayles:
@nvie, if you're up to adding tests and resolving merge conflicts, I'll review and merge.

Or, If you'd rather, I can start a new branch and do those things in a separate PR.

nvie:
> Or, If you'd rather, I can start a new branch and do those things in a separate PR.

@bbayles Feel free to do it.  I won't get to it this week anyway.

bbayles:
OK, I'll close this and we'll take this back up in PR #79. Thanks!

PR #51, opened 2014-09-10
Switch to pytest (because nose hides problems)

nvie:
Disclaimer: there is a reason, other than personal taste, for this change.
#### The problem with nose

When I was running the test suite on Python 3 (using the tox + nose setup), I was quite puzzled why the tests were passing even though [this line is present in `more.py`](https://github.com/erikrose/more-itertools/blob/master/more_itertools/more.py#L4), and `izip_longest` isn't in the `itertools` package anymore in Python 3.

After diving into it a bit more, I found that adding this line _would_ break the tests.

```
from itertools import izip_longest
print(izip_longest)  # use the imported function to expose the problem
```

Strangest, the tests don't fail with an `ImportError`, but instead with an `NameError` on the `print(izip_longest)` line. Following the stack trace, nose seems to be doing some import magic which I haven't fully grasped yet. Strangely enough, the `chunked` tests (which depend on `izip_longest`) _don't_ fail: o_O ?
#### Using pytest instead

This patch changes the test runner to `py.test` instead, which does not showcase said behaviour and fails for Python 3 in the expected way:

```
more_itertools/more.py:4: in <module>
    from itertools import izip_longest
E   ImportError: cannot import name 'izip_longest'
```

Note that I've simply wrapped the `eq_`, `ok_`, and `assert_raises` helpers for now. If you agree with this patch, I'll make sure to rewrite all the invocations directly, replacing them by plain `assert`s, and get rid of the wrapper functions.

Also, I would love to fix the now-exposed Python 3 errors, but I'll wait until after we've decided on this pull request.

@erikrose What do you think?

### DIFF
--- a/docs/testing.rst
+++ b/docs/testing.rst
@@ -2,23 +2,15 @@
 Testing
 =======
 
-more-itertools uses nose for its tests. First, install nose::
-
-    pip install nose
-
-Then, run the tests like this::
-
-    nosetests --with-doctest
-
-
-Multiple Python Versions
-========================
-
-To run the tests on all the versions of Python more-itertools supports, install
-tox::
+To run the tests on all the versions of Python more-itertools supports,
+install tox::
 
     pip install tox
 
 Then, run the tests::
 
     tox
+
+To run it on a single env, use::
+
+    tox -e py27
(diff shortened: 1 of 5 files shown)


bbayles:
I'm inclined to close this in favor of moving away from 2to3. I'd like to get to a place where testing is possible with built-in tools rather than absolutely needing to rely on nose or pytest, but even so I think this PR will be superseded. Thanks!

PR #56, opened 2015-09-19
Add ichunked function

mlyundin:
Please review. Maybe this function will be helpful

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -4,7 +4,7 @@ from functools import partial, wraps
 from itertools import izip_longest
 from recipes import *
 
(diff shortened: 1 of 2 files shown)


bbayles:
Unfortunately, this implementation has an issue with chunk sizes that evenly divide the iterable:

```python
>>> list(list(c) for c in ichunked(range(4), 4))
2: [[0, 1, 2, 3], []]
```

So I think I'll probably close out this PR. However, I'm pretty convinced that splitting an iterator into smaller iterators (not lists) is a nice thing to have, so I'll open an issue about this.

PR #61, opened 2016-11-05
Replace tox with Travis

bbayles:
This PR replaces tox with Travis. Travis will do automatic testing against Python 2.7, 3.3, 3.4, and 3.5.

I didn't add Python 2.6 to the test list, since I'll be dropping Python 2.6 from `setup.py` [soon](https://twitter.com/ErikRose/status/794369557934866433).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,20 @@
+sudo: false
+
+language: "python"
+
+python:
+    - "2.7"
+    - "3.3"
+    - "3.4"
+    - "3.5"
+
+install:
+    - "pip install -U nose"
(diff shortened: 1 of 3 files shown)


erikrose:
Though tox is nice to have around for testing offline (on airplanes, etc.). You can make them work together, like we did in https://github.com/erikrose/parsimonious.

bbayles:
I'll switch to that method; I didn't know about `tox-travis`.

bbayles:
Neat; that worked!

erikrose:
Brilliant!

PR #54, opened 2015-05-01
Windowed

abarnert:
Adds a `windowed` function that generalizes `pairwise` to generate adjacent triples, quadruplets, etc.:

```
    >>> take(4, windowed(count(), 3))
    [(0, 1, 2), (1, 2, 3), (2, 3, 4), (3, 4, 5)]
```

PR #47, opened 2014-08-26
Simplify chunked() implementation

nvie:
What it says on the tin.

In particular it avoids the use of the unnecessary complex tricks like iterable cloning, izipping spreaded args, marker filling, and marker removal.

PR #62, opened 2016-11-05
Drop 2to3

bbayles:
This PR moves `more-itertools` to a single codebase for Python 2 and Python 3. I've used `six` judiciously (with Erik's [endorsement](https://twitter.com/ErikRose/status/794369557934866433)).

The commits show me slowly taking out the 2to3 scaffolding and then removing it. I'll leave a few comments on less-than-obvious changes.

### DIFF
--- a/more_itertools/recipes.py
+++ b/more_itertools/recipes.py
@@ -13,7 +13,7 @@ import operator
 from random import randrange, sample, choice
 
 from six import PY2
-from six.moves import filterfalse, map, zip, zip_longest
+from six.moves import filterfalse, map, range, zip, zip_longest
 
 __all__ = ['take', 'tabulate', 'consume', 'nth', 'quantify', 'padnone',
            'ncycles', 'dotproduct', 'flatten', 'repeatfunc', 'pairwise',
@@ -268,7 +268,7 @@ def iter_except(func, exception, first=None):
     Like __builtin__.iter(func, sentinel) but uses an exception instead
     of a sentinel to end the loop.
 
-        >>> l = range(3)
+        >>> l = [0, 1, 2]
         >>> list(iter_except(l.pop, IndexError))
         [2, 1, 0]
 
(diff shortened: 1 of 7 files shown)


more_itertools/recipes.py
```diff
@@ -200,7 +204,10 @@ def roundrobin(*iterables):
     """
     # Recipe credited to George Sakkis
     pending = len(iterables)
-    nexts = cycle(iter(it).next for it in iterables)
+    if PY2:
```
bbayles: This is sort of an eyesore, but seemed better than `getattr(it, '__next__', getattr(it, 'next'))`

more_itertools/recipes.py
```diff
@@ -287,7 +294,7 @@ def random_product(*args, **kwds):
         ('b', '2', 'c', '2')
 
     """
-    pools = map(tuple, args) * kwds.get('repeat', 1)
+    pools = [tuple(pool) for pool in args] * kwds.get('repeat', 1)
```
bbayles: This is cribbed from the Py3 itertools docs.

setup.py
```diff
@@ -26,6 +19,7 @@
     author_email='erikrose@grinchcentral.com',
     license='MIT',
     packages=find_packages(exclude=['ez_setup']),
+    install_requires=['six>=1.10.0'],
```
bbayles: This is the latest version of `six`; I'd also consider specifying >= 1.0.0.

erikrose:
For safety, let's do either of those and <2.0.

more_itertools/recipes.py
erikrose: It looks like it should be equivalent. Do you see why they did that?

erikrose:
Lgtm so far!

more_itertools/recipes.py
bbayles: Yeah - Py3 `map` doesn't return a multiply-able `list` like Py2's does. Could wrap the `map(tuple ,args)` in a `list()`, but the comprehension is cleaner.

erikrose:
Ah, right. Everything got lazier. Carry on!

bbayles:
OK, I think this is where we want it.

After this I'd like to get a version of #47 in, and then a release?

bbayles:
Whoops, I meant #54 (sliding window) and not #47. Then a release.

erikrose:
If you're waiting for me to okay this PR, consider it okayed! You've been doing great work!

Issue #32, opened 2014-03-08
Include implementations from jaraco.util.itertools


jaraco:
I just now discovered more-itertools. Thanks for that!

I'd like to leverage this project to supersede the implementations I have in [jaraco.util.itertools](https://bitbucket.org/jaraco/jaraco.util/src/156cbbc862aea5737eb188f9b1e0d5224d3b5002/jaraco/util/itertools.py?at=default) ([docs](http://pythonhosted.org/jaraco.util/jaraco.util.html#module-jaraco.util.itertools)). I see in the README the invite to send pull requests, but since this effort would be a somewhat large undertaking, I'd like to preface with some questions and concerns.

1. Would more-itertools likely accept most or all of jaraco.util.itertools?
2. Do you see any obvious conflicts or undesirable parts of that module (so I can exclude them from the pull request(s)?
3. Would you prefer individual pull requests for each tool, or would a large pull request be adequate?
4. jaraco.util.itertools relies heavily on doctests (for documentation and testing). Can more-itertools leverage those tests and documentation?

jaraco:
Please tag wont fix.

Issue #29, opened 2013-09-23
New itertool: window


buchanae:
How about adding a sliding window itertool?

There are multiple implementations that trade off on things like memory consumption/speed/etc, but here's a version that's worked fine for me

```
from collections import deque

def window(seq, n=2):
    it = iter(seq)
    win = deque((next(it, None) for _ in xrange(n)), maxlen=n)
    yield win
    append = win.append
    for e in it:
        append(e)
        yield win
```

bbayles:
Fixed in #63. Thanks!

PR #63, opened 2016-11-06
Sliding window function

bbayles:
This PR adds a `windowed` function, which produces a sliding window over an iterable. It's a generalization of `pairwise`, which does a window of length 2.

```python
>>> all_windows = windowed([1, 2, 3, 4, 5], 3)
>>> next(all_windows)
(1, 2, 3)
>>> next(all_windows)
(2, 3, 4)
>>> next(all_windows)
(3, 4, 5)
```

---

This seems to be a popular request: we have it in issue #29, in PR #32, and again in PR #54.

There are various implementations floating around - @buchanae uses one with a `deque`, @abarnert extends the method of `pairwise`, Raymond Hettinger's [initial version](https://github.com/python/cpython/blob/f6dca3649114dc488b8df02328e6962d8ae9f6ad/Lib/test/test_itertools.py#L441-L450) took a different approach, and there are lots of attempts in [this StackOverflow thread](http://stackoverflow.com/questions/6822725/rolling-or-sliding-window-iterator-in-python).

I like the simplicity of the `tee` versions, but they have the disadvantage of emitting nothing when the window's width exceeds the length of the iterable.

I think when you specify that you want a window of width `n` you expect back `n` items, so I pad with a `fillvalue` (default `None)`, a la `itertools.zip_longest`.

```python
# fillvalue not specified
>>> all_windows = windowed([1, 2, 3, 4, 5], 6)
>>> next(all_windows)
(1, 2, 3, 4, 5, None)

# fillvalue specified
>>> all_windows = windowed([1, 2, 3, 4, 5], 6, fillvalue='')
>>> next(all_windows)
(1, 2, 3, 4, 5, '')
```

This allows you to do things like `a, b, c = next(all_windows, 3)` without needing to worry about the input iterable.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -19,6 +19,7 @@ New Routines
 .. autofunction:: one
 .. autoclass:: peekable
 .. autofunction:: unique_to_each
+.. autofunction:: windowed
 .. autofunction:: with_iter
 
 
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -1,5 +1,6 @@
 from __future__ import print_function
 
+from collections import deque
 from functools import partial, wraps
 
 from six import iteritems
@@ -8,7 +9,7 @@ from .recipes import take
 
 __all__ = ['chunked', 'first', 'peekable', 'collate', 'consumer', 'ilen',
            'iterate', 'with_iter', 'one', 'distinct_permutations',
-           'intersperse', 'unique_to_each']
+           'intersperse', 'unique_to_each', 'windowed']
 
 
 _marker = object()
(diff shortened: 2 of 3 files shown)


erikrose:
Actually, is there a motivation to return tuples, which I generally reserve for anonymous structs and for things that need hashability? I see we return lists from most other routines.

bbayles:
Changed the double dash. Re: tuples vs. lists, `pairwise` emits tuples:

```python
>>> list(pairwise('abcd'))
0: [('a', 'b'), ('b', 'c'), ('c', 'd')]
```

erikrose:
Wfm. I suppose you can argue that the most common use of overlapping windows would be for some kind of position-dependent struct.

PR #64, opened 2016-11-08
Bump to version 2.3

bbayles:
Now that we have `windowed`, I'd like to get it out to PyPI for the world to enjoy. I've updated the version strings and changelog in this PR.

@erikrose, if you're good with sending this to PyPI, I can create the GitHub release.

FWIW, in [redis-collections](https://github.com/honzajavorek/redis-collections/blob/master/.travis.yml) we have Travis do the PyPI push - that might be a good thing here?

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -48,7 +48,7 @@ copyright = u'2012, Erik Rose'
 # built documents.
 #
 # The short X.Y version.
-version = '2.2'
+version = '2.3'
 # The full version, including alpha/beta/rc tags.
 release = version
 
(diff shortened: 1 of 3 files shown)


erikrose:
Fantastic. I'm happy to do the PyPI release for now. If you end up continuing to work on more-itertools, I'll give you the keys as well. Thus far I've resisted giving Travis the ability to write to PyPI. I think it's risky to concentrate in one place the ability to push so many packages into the world; it makes too juicy of a target. So I'm going to add one more polish PR tweaking some metadata, and, if that looks good to you, we can cut a release!

PR #66, opened 2016-11-09
Bucket (was "Separate" and "Partition")

bbayles:
This PR builds on the ideas in #26 and adds a `separate` function. By default it will split an input iterable into two - items that have `bool(item) == True` and items that have `bool(item) == False`. A dictionary with `True` and `False` keys is returned:

```python
>>> iterable = [0, '', 1, 'x', None, [1]]
>>> D = separate(iterable)
>>> list(D[False])
[0, '', None]
>>> list(D[True])
[1, 'x', [1]]
```

You can customize the keys of the returned dictionary as well as the function that operates on the items:

```python
>>> iterable = [0, 1, 2, 3, 4, 5, 6, 7, 8]
>>> D = separate(iterable, keys=(0, 1, 2), fn=lambda x: x % 3)
>>> list(D[0])  # Divisible by 3
[0, 3, 6]
>>> list(D[1])  # Remainder 1 when divided by 3
[1, 4, 7]
>>> list(D[2])  # Remainder 2 when divided by 3
[2, 5, 8]
```

Since we can't know what the return values of the function will be until we run it on all of the items, you must specify up-front what you think they'll be (obviously `bool` only returns `True` and `False`). I considered having an "other" key in the dictionary for items that don't match any `keys`, but decided against it.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -483,6 +483,7 @@ class separate(object):
                     item = next(self._it)
                     item_value = self._key(item)
                     if item_value == value:
+                        yield item
                         break
                     self._cache[item_value].append(item)
 
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -464,37 +464,17 @@ class bucket(object):
 
     The original iterable will be advanced and its items will be cached until
     they are used by the child iterables. This may require significant storage.
-
-    By default, attempting to select a value that does not match any items in
-    the iterable will exhaust the iterable, caching all the values.
-    If a set of *values* is provided, items that don't match any will not be
-    cached, and attempting to select them will generate a KeyError.
-
-    >>> iterable = [0, 1, 2, 3, 4, 5, 6, 7, 8]
-    >>> mod_3 = lambda x: x % 3  # Remainder when divided by 3
-    >>> values = {0, 1}  # Only keep items that equal 0 or 1 (mod 3)
-    >>> s = bucket(iterable, key=mod_3, values=values)
-    >>> list(s[0])
-    [0, 3, 6]
-    >>> list(s[1])
-    [1, 4, 7]
-    >>> list(s[2])
-    Traceback (most recent call last):
-    ...
-    KeyError: 2
+    Be aware that attempting to select a bucket that no items correspond to
+    will exhaust the iterable and cache all values.
 
     """
 
-    def __init__(self, iterable, key, values=()):
+    def __init__(self, iterable, key):
         self._it = iter(iterable)
         self._key = key
         self._cache = defaultdict(deque)
-        self._values = set(values)
 
     def __contains__(self, value):
-        if self._values and (value not in self._values):
-            return False
-
         try:
             item = next(self[value])
         except StopIteration:
@@ -524,12 +504,8 @@ class bucket(object):
                     if item_value == value:
                         yield item
                         break
-                    elif self._values and (item_value not in self._values):
-                        continue
                     else:
                         self._cache[item_value].append(item)
 
     def __getitem__(self, value):
-        if self._values and (value not in self._values):
-            raise KeyError(value)
         return self._get_values(value)
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -499,10 +499,3 @@ class separate(object):
 
     def __getitem__(self, value):
         return self._get_values(value)
-
-    def __next__(self):
-        return next(self._it)
-
-    def next(self):
-        # For Python 2 compatibility
-        return self.__next__()
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -11,7 +11,7 @@ from .recipes import take
 
 __all__ = ['chunked', 'first', 'peekable', 'collate', 'consumer', 'ilen',
            'iterate', 'with_iter', 'one', 'distinct_permutations',
-           'intersperse', 'unique_to_each', 'windowed', 'separate']
+           'intersperse', 'unique_to_each', 'windowed', 'bucket']
 
 
 _marker = object()
@@ -447,13 +447,13 @@ def windowed(seq, n, fillvalue=None):
         yield tuple(window)
 
 
-class separate(object):
+class bucket(object):
     """
-    Wraps an iterable and returns an object that separates the iterable
+    Wraps an iterable and returns an object that buckets the iterable
     into child iterables based on the *key* function.
 
     >>> iterable = ['a1', 'b1', 'c1', 'a2', 'b2', 'c2', 'b3']
-    >>> s = separate(iterable, key=lambda s: s[0])  # Select by first character
+    >>> s = bucket(iterable, key=lambda s: s[0])  # Select by first character
     >>> a_iterable = s['a']
     >>> next(a_iterable)
     'a1'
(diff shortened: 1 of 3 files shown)


bbayles:
I'll probably rename this; `partition` is now used in the [recipes](https://docs.python.org/3/library/itertools.html) for Python 3. See #67.

erikrose:
Plus there's `str.partition`, which does something different.

bbayles:
I went with `separate()` instead of `partition`.

more_itertools/more.py
```diff
@@ -424,3 +424,61 @@ def windowed(seq, n, fillvalue=None):
     for item in it:
         append(item)
         yield tuple(window)
+
+
+def separate(iterable, keys=None, fn=None):
```
erikrose: (Repeating this because it got hidden by the partition→separate rename.) Is there any reason we can't determine the keys dynamically and thus obviate the need for an arg? Then we could rename `fn` to `key` and look like `sort()` et al. Obviously, we'd have to return a custom object that implements `__getitem__` rather than a dict....

erikrose:
Sorry, GH reviews (or my use of them) are kind of clumsy for the moment. :-)

erikrose:
Yes, it still requires buffering the whole thing, but the call no longer requires `keys`. There might even be situations in which `keys` could not be predetermined.

bbayles:
I think I prefer the up-front specification of keys, because I think without that we're basically back to the nice and simple version with `defaultdict(list)` from [this comment](https://github.com/erikrose/more-itertools/pull/26#issuecomment-19247802).

Is this the sort of behavior you have in mind?

```python
>>> iterable = range(100)
>>> key = lambda x: 10 * (x // 10)  # Round down to nearest 10
>>> s = Separator(key)  # Wrap the iterable
>>> next(s[10])  # Everything up to 10 is buffered?
10
>>> next(s[10])
11
>>> next(s[100])  # Everything is read in now, we didn't find the value, so... some exception?
KeyError
>>> next(s[20])
20
```

erikrose:
The defaultdict approach has the disadvantage of sucking the whole iterable into RAM, which definitely won't go so well with infinite iterables. ;-) But the behavior you outline is what I'm suggesting.

Also, that's a really clever way of rounding down to the nearest 10.

bbayles:
I see your point re: infinite iterables, but what would you want to happen in this case?

```python
>>> from itertools import count  # Here's an infinite iterable
>>> s = Separator(count(), key=lambda x: x)
>>> s['a']  # Wait forever and run out of memory?
```

erikrose:
Wait forever and run out of memory. ;-) I'm not sure what else is to be done, unless we stick an arbitrary limit on it. Any greedy approach will immediately run out of RAM on *any* infinite iterable; this will only on user error or pathological input.

bbayles:
That behavior seems like a dealbreaker to me. You can't do `item in s`, because it will run out of memory. You can't try to catch `KeyError` because it will come out of memory, etc.

I'm OK with closing this out - we have the initial idea for `partition` in `recipes.py` now. I think there's something neat and potentially useful in the "specify keys up front" version, but do concede that it's not a very common thing to see.

erikrose:
I don't know that it should be a dealbreaker, as it has precedent: this simple construct spins forever as well. In fact, it spins at such a low level that you can't even ctrl-C out of the interpreter:

```python
from itertools import count
'a' in count()
```

When using `separate`, you have to be careful to know the range of your function and code according to it. I think that's fine, as I've had many situations where `separate` (or `bucket`, as I seem to keep calling it) would have been a useful routine. You can infinite-loop or exhaust memory in a one-liner any number of ways, so that doesn't kill it for me, if we document well.

bbayles:
OK, I can see where you're coming from. I think we can have the best of both worlds! The latest commits implement the essentials of what I sketched out above:

Default behavior:
```python
>>> s = separate(count(), key=lambda x: 10 * (x // 10))
>>> next(s[10])
10
>>> next(s[10])
11
>>> next(s[100])
100
>>> next(s[100])
101
```

With values pre-specified:
```python
>>> s = separate(count(), key=lambda x: 10 * (x // 10), values={10, 20, 30})
>>> next(s[10])
10
>>> next(s[10])
11
>>> next(s['a'])  # We didn't run out of memory!
...
KeyError: 'a'
```

This could be useful if the types of values you care about are sparse in the iterable and you don't want to store everything in between when iterating.

more_itertools/more.py
```diff
@@ -445,3 +445,81 @@ def windowed(seq, n, fillvalue=None):
...
+    >>> list(s[1])
+    [1, 4, 7]
+    >>> list(s[2])
+    Traceback (most recent call last):
+    ...
+    KeyError: 2
+
+    """
+
+    def __init__(self, iterable, key, values=tuple()):
```
erikrose: `tuple()` can be more concisely spelled `()`.

more_itertools/more.py
```diff
@@ -445,3 +445,81 @@ def windowed(seq, n, fillvalue=None):
...
+        try:
+            item = next(self[value])
+        except StopIteration:
+            return False
+        else:
+            self._cache[value].appendleft(item)
+
+        return True
+
+    def _get_values(self, value):
```
erikrose: I could use a docstring for this, as I try to reason about it.

more_itertools/more.py
```diff
@@ -445,3 +445,81 @@ def windowed(seq, n, fillvalue=None):
...
+                yield self._cache[value].popleft()
+            else:
+                while True:
+                    item = next(self._it)
+                    item_value = self._key(item)
+                    if item_value == value:
+                        yield item
+                        break
+                    if self._values and (item_value not in self._values):
+                        continue
```
erikrose: I'm not sure what the answer is, but is it better to act as an implicit `filter` here as this does, or to raise some kind of error as in "I wasn't expecting this value!"? The [concrete application I've been using in my head to vet this](https://github.com/erikrose/peep/blob/master/peep.py#L868-L872) would have benefitted from the latter. How about ones you've got in mind?

erikrose:
Also, oops: we've become inconsistent in our naming. We used to use nouns, like chunked, first, and peekable, but now we've slipped in some verbs, like collate and intersperse. Not really sure how best to recover from that.

erikrose:
Aside from those comments, this lgtm!

erikrose:
(I'm partial to the name "bucket" (which imply "bucketed" or "buckets" as a noun) if you don't have a strong preference. If you do, go with yours.)

bbayles:
Re: nouns and verbs, I think that ship sailed in `recipes`. Take, Tabulate, Consume, Pad None, Flatten...

Re: naming, I'm fine with `bucket` if you like it; `separate` was my second choice anyway.

If you're cool with the state of this, could you change your GitHub "PR Review" status from "requests changes?" (I'm still getting used to the PR review workflow)

erikrose:
Never mind; you're right!

erikrose:
Lgtm! (I'm still figuring out the review workflow as well and didn't mean to start some elaborate process. But now Start Some Elaborate Process is the default when I hit command-Return. :-))

more_itertools/more.py
bbayles: Hmm, I suppose if you wanted my behavior (which is to be able to bucket the iterable by some known values without having to worry about running out of memory due to non-matching values) could be solved by first running the iterable through `filter` / `ifilter`. So I'd be OK with zapping the `values` functionality and reminding people of this option in the docs?

bbayles: Re: raising an exception for unexpected values... not sure. I think I'm going to fall back to simplicity.

erikrose:
You know, that's a really excellent idea. Itertools are always meant to be composed, after all.

erikrose:
It's going to be so nice not rewriting this next time I need it. :-)

Issue #11, opened 2012-08-04
more_itertools.collate is too similar to heapq.merge


kmike:
Hi there,

Nice idea to create more_itertools package! But it seems more_itertools.collate is duplicating heapq.merge from standard library with minor differences. Do you really need it?

julianpistorius:
Update: Python 3.5 added `key` and `reverse`:

https://docs.python.org/3/library/heapq.html#heapq.merge

Thankfully `more-itertools` is there for those of us stuck in 2.7 land. :)

bbayles:
I like the "call out to `heapq.merge()` if we can" method; I'll do that in a PR soon (1274fe20b6234df3917dfe3748399a36c431bebc).

bbayles:
Fixed in #69. Thanks @kmike and @julianpistorius !

PR #69, opened 2016-11-10
Make collate() use heapq.merge() where possible

bbayles:
Re: #11, this PR changes `collate()` such that it returns from `heapq.merge()` if:
* Neither the `key` nor the `reverse` arguments are specified, or
* The user is on Python 3.5 or later (which supports `key` and `reverse`)

Otherwise the normal `collate()` routine is run. As discussed in the issue, `heapq.merge()` is faster for a few reasons.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -2,6 +2,8 @@ from __future__ import print_function
 
 from collections import deque
 from functools import partial, wraps
+from heapq import merge
+from sys import version_info
 
 from six import iteritems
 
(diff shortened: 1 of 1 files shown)


more_itertools/more.py
```diff
@@ -166,17 +185,17 @@ def collate(*iterables, **kwargs):
...
+        return merge(*iterables)
 
-    min_or_max = partial(max if reverse else min, key=lambda a_b: a_b[0])
-    peekables = [peekable(it) for it in iterables]
-    peekables = [p for p in peekables if p]  # Kill empties.
-    while peekables:
-        _, p = min_or_max((key(p.peek()), p) for p in peekables)
-        yield next(p)
-        peekables = [x for x in peekables if x]
+    if (version_info[0] >= 3) and (version_info[1] >= 5):
```
erikrose: For further efficiency, we could avoid the version check on every invocation by doing it once, at import time.

erikrose:
Otherwise, looks okay to me.

more_itertools/more.py
bbayles: I had that at one point, but I put this back in the kwarg check to keep it clear what's going on. I'll try it a different way.

PR #71, opened 2016-11-11
Indexing and slicing of peekables

bbayles:
Re: Issue #4, this PR adds indexing and slicing support for peekables.

As described in the issue, you can think of this as analogous to calling `list()` on the remaining items in the iterable and indexing/slicing the result. The peekable will not be advanced, however.

You can even index and slice infinite iterables:
```python
... from more_itertools import peekable
... 
... p = peekable(count())
... p[10:21:2]
0: [10, 12, 14, 16, 18, 20]
>>> next(p)
0
>>> next(p)
1
```

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -70,8 +70,6 @@ def first(iterable, default=_marker):
         return default
 
 
-
-
 class peekable(object):
     """
     Wrapper for an iterator to allow lookahead.
@@ -161,15 +159,14 @@ class peekable(object):
         return self.__next__()
 
     def _get_slice(self, index):
-        start = 0 if index.start is None else index.start
+        start = index.start
         stop = index.stop
-        step = index.step
 
-        if (start < 0) or ((stop is not None) and (stop < 0)):
+        if (
+            ((start is not None) and (start < 0)) or
+            ((stop is not None) and (stop < 0))
+        ):
             raise ValueError('Negative indexing not supported')
-
-        if start == stop:
-            return []
 
         cache_len = len(self._cache)
 
@@ -186,7 +183,7 @@ class peekable(object):
                 except StopIteration:
                     break
 
-        return list(self._cache)[start:stop:step]
+        return list(self._cache)[index]
 
     def __getitem__(self, index):
         if isinstance(index, slice):
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -171,7 +171,7 @@ class peekable(object):
         cache_len = len(self._cache)
 
         if stop is None:
-            self._cache.extend(islice(self._it, None))
+            self._cache.extend(self._it)
         elif stop >= cache_len:
             self._cache.extend(islice(self._it, stop - cache_len))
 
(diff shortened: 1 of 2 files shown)


more_itertools/more.py
```diff
@@ -129,24 +140,57 @@ def peek(self, default=_marker):
...
+            ((stop is not None) and (stop < 0))
+        ):
+            raise ValueError('Negative indexing not supported')
+
+        cache_len = len(self._cache)
+
+        if stop is None:
+            while True:
+                try:
+                    self._cache.append(next(self._it))
```
erikrose: I wonder if we could do this with .extend and islice() and save the overhead of spinning around in the while loop.

more_itertools/more.py
```diff
@@ -129,24 +140,57 @@ def peek(self, default=_marker):
...
+                except StopIteration:
+                    break
+        elif stop >= cache_len:
+            for i in range(stop - cache_len):
+                try:
+                    self._cache.append(next(self._it))
```
erikrose: Same here, really

more_itertools/more.py
```diff
@@ -129,24 +140,57 @@ def peek(self, default=_marker):
...
+                except StopIteration:
+                    break
+
+        return list(self._cache)[index]
```
erikrose: Does it matter? Are the results of a slice required to be themselves sliceable?

bbayles: It matters because that's how we're returning the slice, yeah?

bbayles:
Not sure whether it actually saves any overhead, but `islice` does look much nicer!

erikrose:
In general, staying out of the interpreter loop helps. Using itertools primitives help a LOT, because they turn straight into C stack frames.

erikrose:
Let me know if you want any more opinions on this. I don't see any showstoppers. It'll be exciting to have a well-maintained version of this instead of a bunch of copy pasta on Stack Overflow.

I'm not sure what the right decision is about the datatype to return from `__getitem__(slice)`. Lists are convenient but use a ptr's worth of mem for every item, while https://docs.python.org/2/reference/datamodel.html?highlight=getitem#object.__getitem__ suggests slices should be "of the same type" as the parent container but may be inapplicable since it's talking about sequences. Having typed that, I find I've just argued myself into returning an islice, since I bet it performs faster and smaller and it can be transformed into a list by the caller if needed. Whatever you decide, I trust your judgment.

bbayles:
I almost buy the argument for returning an `islice` iterable, but (a) I'm using peekable in the first place because I don't want my iterator-y to be so iterator-y; I asked for a slice and now I have to iterate over it to materialize it? (b) I've never encountered something that I've used slice notation on that returned an iterator.

So I think I'll stick with list and appreciate your trust of my judgement. I like this addition, and am pleased to solve the oldest open issue.

bbayles:
Wait, I checked what `range` does, and you're right - this should return an iterator. I was so sure of myself 5 minutes ago.

bbayles:
Wait, scratch that again - if this returns an `islice` of `self._cache`, then `self._cache` could change by the time you finish iterating over it.

Back to `list`, and I'm merging before I change my mind again.

erikrose:
Ha! Sounds good. We can always flop around more before the next release.

PR #74, opened 2016-11-13
Add interleave, interleave_longest, and collapse

bbayles:
This PR builds on #22, adding @abarnert's `interleave`, `interleave_longest`, and `collapse` functions.

I've updated them to work with Python 3 natively, and addressed some of the comments from the original PR.

Re: [this comment thread](https://github.com/erikrose/more-itertools/pull/22#r3275958) about `collapse`, the situation is a bit more complicated than it seems.

The function currently special-cases text types. The thread suggested that `basestring` (or `six.string_types`) could be used as the default for the `basetype` argument. However, this will cause infinite recursion without a special case, as `iter(s)` will return an iterator even when `s` is a single character string.

My feeling is that @abarnert was basically right about what people want (strings treated as a unit, not just another iterable) and that this should continue to be the behavior, but I'm open to alternate suggestions.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -571,9 +571,8 @@ def interleave_longest(*iterables):
     >>> list(interleave_longest([1, 2, 3], [4, 5], [6, 7, 8]))
     [1, 4, 6, 2, 5, 7, 3, 8]
     """
-    dummy = object()
-    i = chain.from_iterable(zip_longest(*iterables, fillvalue=dummy))
-    return filter(lambda x: x is not dummy, i)
+    i = chain.from_iterable(zip_longest(*iterables, fillvalue=_marker))
+    return filter(lambda x: x is not _marker, i)
 
 
 def collapse(iterable, basetype=None, levels=None):
@@ -595,6 +594,7 @@ def collapse(iterable, basetype=None, levels=None):
             basetype and isinstance(node, basetype)):
             yield node
             return
+
         try:
             tree = iter(node)
         except TypeError:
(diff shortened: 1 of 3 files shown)


more_itertools/more.py
```diff
@@ -545,3 +550,62 @@ def _get_values(self, value):
...
+
+    >>> list(collapse([[1], 2, [[3], 4], [[[5]]], 'abc']))
+    [1, 2, 3, 4, 5, 'abc']
+    >>> list(collapse([[1], 2, [[3], 4], [[[5]]]], levels=2))
+    [1, 2, 3, 4, [5]]
+    >>> list(collapse((1, [2], (3, [4, (5,)])), list))
+    [1, [2], 3, [4, (5,)]]
+    """
+    def walk(node, level):
+        if (
```
bbayles: Ah, but this style is in PEP 8! See last examples in the [indentation section](https://www.python.org/dev/peps/pep-0008/#id17). I generally run my commits through [pycodestyle](https://github.com/PyCQA/pycodestyle) to check for PEP 8 violations.

more_itertools/more.py
```diff
@@ -545,3 +550,62 @@ def _get_values(self, value):
...
+            ((levels is not None) and (level > levels)) or
+            isinstance(node, string_types) or
```
bbayles: I'll agree on renaming `basetype`.

more_itertools/more.py
erikrose: Crazypants! ;-) I wonder if those are new additions. Carry on!

erikrose:
Lgtm!

Issue #53, opened 2014-12-27
Peek without peekable


themiurgo:
Hello,

sometimes I need to peek at the first element of an iterable. I usually implement it like this:

```
def peek(iterable):
    iterator = iter(iterable)
    item = next(iterator)
    return item, itertools.chain([item], iterator)

element, my_list = peek(my_list)
```


Would this be an interesting addition to more-itertools? If so, I'll send a pull request sometime in the future.

Cheers.

mathieulongtin:
I managed to make peekable 3 times faster, but it's still way slower than this, and it fits a lot of my use cases for peekable. Thanks.

```
In [63]: data = range(10000)

In [64]: %timeit sum(peekable(iter(data)))
The slowest run took 4.38 times longer than the fastest. This could mean that an intermediate result is being cached
10 loops, best of 3: 26.6 ms per loop

In [65]: %timeit sum(peek(data)[1])
1000 loops, best of 3: 242 µs per loop

In [66]:
```

themiurgo:
Glad it is useful! If you ever want it as a drop-in function and want to avoid rewriting it, it is included in the `toolz` library: https://github.com/pytoolz/toolz/pull/244

mathieulongtin:
Didn't know about toolz, looks great. Thanks. 

I made slight modification to peek so that it can return a default instead of raise StopIteration:

```
_marker = object()
def peek(iterable, default=_marker):
      iterator = iter(iterable)
      try:
          item = next(iterator)
          return item, itertools.chain([item], iterator)
      except StopIteration:
          if default == _marker:
              raise
          return default, iterator
```

bbayles:
I'm trying to decide on what to do with this issue. On the one hand: `toolz` already has this, `more_itertools` already has a serviceable-but-slower method of doing the same thing. On the other hand, this procedure is _way_ faster than the existing one.

I did find a 2x-3x speedup for the existing `peekable` class - I'll make a PR for that shortly. However, that's still much slower than this method.

I'm inclined to (a) rename this function from `peek` to `look_ahead`, and (b) extend it to return an arbitrary number of elements from the head of the iterator.

erikrose:
I think your inclination is fine. Another fun (and short) name is `spy`.

bbayles:
Fixed in #76 (and #76) - many thanks @themiurgo and @mathieulongtin.

PR #75, opened 2016-11-15
Iterate through peekables more quickly

bbayles:
Re: [this discussion in issue #53](https://github.com/erikrose/more-itertools/issues/53#issuecomment-260524759), this PR modifies `peekable.__next__` to make (destructive) iteration through `peekable()`-wrapped iterators faster.

On some tests (like this one) the difference is 2.5x:
```python
>>> for func in (peekable_old, peekable_new):
...     def stmt():
...         iterable = iter(range(10000))
...         sum(func(iterable))
... 
...     result = timeit(stmt, number=1000)
...     print(result)
8.013842238986399
3.322781088994816
```

The simplification is that we don't need to mess with the cache if it's empty (likely to be the case if we're iterating through in a loop), and if it's not empty we can pop off its leftmost item and return that.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -150,9 +150,10 @@ class peekable(object):
         return self._cache[0]
 
     def __next__(self):
-        ret = self.peek()
-        self._cache.popleft()
-        return ret
+        if self._cache:
+            return self._cache.popleft()
+
+        return next(self._it)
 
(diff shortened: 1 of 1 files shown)


erikrose:
That's an excellent payoff for adding 1 branch. :-) lgtm!

PR #76, opened 2016-11-15
Add look_ahead function

bbayles:
Re: Issue #53, this adds a "peek without `peekable()`" function called `look_ahead()`. It returns the first `n` items from an iterable, plus a copy of the iterable that has not been advanced.

Iterating over `look_ahead()`-returned iterable is _much_ faster than iterating over a `peekable()`-wrapped iterable.
```python
>>> from more_itertools import look_ahead, peekable
... from timeit import timeit
... 
... def old():
...     iterable = iter(range(10000))
...     p = peekable(iterable)
...     assert sum(peekable(iterable)) == 49995000
... 
... def new():
...     iterable = iter(range(10000))
...     head, p = look_ahead(iterable, 0)
...     assert sum(look_ahead(iterable)[1]) == 49995000
... 
... print(timeit(old, number=1000))
... print(timeit(new, number=1000))
8.047204594011419
0.25736623600823805
```

That's over 30x faster than the master branch, and still at least 15x faster than the optimized `peekable` in PR #75.

Many thanks to @themiurgo and @mathieulongtin for their contribution to this.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -3,15 +3,18 @@ from __future__ import print_function
 from collections import defaultdict, deque
 from functools import partial, wraps
 from heapq import merge
+from itertools import chain
 from sys import version_info
 
 from six import iteritems
 
 from .recipes import islice, take
 
-__all__ = ['chunked', 'first', 'peekable', 'collate', 'consumer', 'ilen',
-           'iterate', 'with_iter', 'one', 'distinct_permutations',
-           'intersperse', 'unique_to_each', 'windowed', 'bucket']
+__all__ = [
+    'chunked', 'first', 'peekable', 'collate', 'consumer', 'ilen', 'iterate',
+    'with_iter', 'one', 'distinct_permutations', 'intersperse',
+    'unique_to_each', 'windowed', 'bucket', 'look_ahead'
+]
 
 
 _marker = object()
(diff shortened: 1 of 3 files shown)


more_itertools/more.py
```diff
@@ -545,3 +548,47 @@ def _get_values(self, value):
...
+        ['a']
+        >>> list(iterable)
+        ['a', 'b', 'c', 'd', 'e', 'f', 'g']
+
+    You may use unpacking to retrieve items instead of lists:
+
+        >>> (head,), iterable = look_ahead('abcdefg')
+        >>> head
+        'a'
+        >>> (first, second), iterable = look_ahead('abcdefg', 2)
```
erikrose: I love this API: you get can your data singly or listly without needing to set an option.

more_itertools/more.py
```diff
@@ -545,3 +548,47 @@ def _get_values(self, value):
 
     def __getitem__(self, value):
         return self._get_values(value)
+
+
+def look_ahead(iterable, n=1):
```
erikrose: Slight preference for the name `spy` for brevity and similarity with `peek`. Feel free to overrule.

erikrose:
Otherwise, lgtm!

bbayles:
I liked `spy`, good thought.

Issue #39, opened 2014-05-10
Add function to chunk an iterable based on a predicate


astronouth7303:
Add a generator that produces a generator for each chunk, where each chunk is defined by a "is first item" or "is last item" predicate.

Basic code for "is first item" idea:

``` python

def chunk_by_first(seq, pred):
    """chunk_by_first(iterable, callable) -> list, ...
    Breaks up an iterable based on a predicate: If true, break before that item.
    """
    buf = []
    for i in seq:
        if pred(i) and len(buf):
            yield buf
            buf = []
        buf.append(i)
    yield buf
```

(This implementation is of course undesirable if chunks are of considerable size. But should concisely express the idea.)

bbayles:
@astronouth7303, I'm in favor of something (or things) like this getting into `more_itertools`. The original version that emits lists I think is good - attempting to split an iterator into sub-iterators is perilous (as you've found, and as I found in #72 mor recently).

I'd like a name that distinguishes this from `chunked` (which we're using to indicate equal-sized subsequences) and `partition` (which we're using to mean "split in two pieces"). I think `split` might work, though there is a precedent for that in the `str` builtin. Better ideas are welcome.

Below I have a `split_before` (basically the same as your original) and a `split_after` (which instead breaks after the predicate is True).

---


```python
def split_before(iterable, pred):
    buf = []
    for item in iter(iterable):
        if pred(item) and buf:
            yield buf
            buf = []
        buf.append(item)
    yield buf

def split_after(iterable, pred):
    buf = []
    for item in iter(iterable):
        buf.append(item)
        if pred(item) and buf:
            yield buf
            buf = []
    if buf:
        yield buf
```

These can have neat applications:
```python
>>> from gmpy2 import is_prime
>>> iterable = range(2, 24)
>>> list(split_before(iterable, is_prime))
[[2],
 [3, 4],
 [5, 6],
 [7, 8, 9, 10],
 [11, 12],
 [13, 14, 15, 16],
 [17, 18],
 [19, 20, 21, 22],
 [23]]
>>> list(split_after(iterable, is_prime))
[[2],
 [3],
 [4, 5],
 [6, 7],
 [8, 9, 10, 11],
 [12, 13],
 [14, 15, 16, 17],
 [18, 19],
 [20, 21, 22, 23]]
```

PR #78, opened 2016-11-17
Add split_before and split_after functions

bbayles:
Re: Issue #39, this PR adds two new functions - `split_before()` and `split_after()`. These break up a list based on a user-supplied condition. There are two functions, as doing this with a keyword argument is pretty awkward, and there is precedent for that sort of thing in the `str` methods like `split` and `rsplit`.

Many thanks to @astronouth7303 for the idea and the implementation of `split_before()`. I like these examples in particular:
```python
>>> list(split_before('OneTwo', lambda s: s.isupper()))
[['O', 'n', 'e'], ['T', 'w', 'o']]
>>> list(split_after(iterable, is_prime))
[[2],
 [3],
 [4, 5],
 [6, 7],
 [8, 9, 10, 11],
 [12, 13],
 [14, 15, 16, 17],
 [18, 19],
 [20, 21, 22, 23]]
```

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -676,7 +676,7 @@ def split_after(iterable, pred):
     """Yield lists of items from *iterable*, where each list ends with an
     item where callable *pred* returns ``True``:
 
-        >>> list(split_after('one1two2', lambda s: s.isnumeric()))
+        >>> list(split_after('one1two2', lambda s: s.isdigit()))
         [['o', 'n', 'e', '1'], ['t', 'w', 'o', '2']]
 
         >>> list(split_after(range(10), lambda n: n % 3 == 0))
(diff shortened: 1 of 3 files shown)


more_itertools/more.py
```diff
@@ -695,3 +696,44 @@ def sliced(seq, n):
...
+    item where callable *pred* returns ``True``:
+
+        >>> list(split_before('OneTwo', lambda s: s.isupper()))
+        [['O', 'n', 'e'], ['T', 'w', 'o']]
+
+        >>> list(split_before(range(10), lambda n: n % 3 == 0))
+        [[0, 1, 2], [3, 4, 5], [6, 7, 8], [9]]
+
+    """
+    buf = []
```
erikrose: Has anybody toyed with using `itertools.takewhile` for speeding either of these routines?

bbayles: I fiddled with it, but since `takewhile` snarfs the last item before it breaks you'd need to peek ahead and I couldn't make that attractive.

erikrose:
Cool beans.

PR #79, opened 2016-11-18
`side_effect()`, round two.

bbayles:
This PR updates #50 from @nvie, adding the `side_effect()` function. I've included tests and changed the docstring a bit as well.


nvie:
Looks good to me.  Thanks, @bbayles!

PR #80, opened 2016-11-18
Avoid leaving a reference in chunked()

bbayles:
Re: Issue #13, this PR modifies `chunked()` in order to make garbage collection of retrieved chunks possible. The method comes from @piskvorky's code. See also [my](https://github.com/erikrose/more-itertools/issues/13#issuecomment-260221700) [comments](https://github.com/erikrose/more-itertools/issues/13#issuecomment-261435348) re: performance.

I couldn't get `sys.getrefcount()` to show me that there was a difference, but maybe someone else has an idea for how to test?

PR #81, opened 2016-11-18
Add sliced function

bbayles:
#5 is the last open issue without a PR. Here I add a `sliced()` function, which is like `chunked()` but for sequences that support slicing - see the [discussion](https://github.com/erikrose/more-itertools/issues/5#issuecomment-7404666) for reasoning behind adding anew function.

~~This PR uses the same trick as #80 does to avoid leaving a reference to the retrieved slice in the generator  function.~~

The open PRs will probably all conflict with each other, but once merged I'll do a changelog and such for a new release.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -15,7 +15,7 @@ __all__ = [
     'chunked', 'first', 'peekable', 'collate', 'consumer', 'ilen', 'iterate',
     'with_iter', 'one', 'distinct_permutations', 'intersperse',
     'unique_to_each', 'windowed', 'bucket', 'spy', 'interleave',
-    'interleave_longest', 'collapse'
+    'interleave_longest', 'collapse', 'sliced'
 ]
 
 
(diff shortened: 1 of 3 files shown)


more_itertools/more.py
```diff
@@ -650,3 +650,34 @@ def walk(node, level):
...
+    append = slice_holder.append
+    pop = slice_holder.pop
+
+    i = 0
+    while True:
+        append(seq[i:i + n])
+        if not slice_holder[0]:
+            return
+        yield pop()
+        i += n
```
coady: A `Sequence` implies `Sized`, so `range` with `step` would be the easier approach:
```
(seq[i:i + n] for i in range(0, len(seq), n))
```

If unsized sliceables *must* be supported, `itertools.count` can substitute for `range`.  And then to avoid this lingering reference problem, the loop could be terminated with `itertools.takewhile`.
```
itertools.takewhile(bool, (seq[i:i + n] for i in itertools.count(0, n)))
```

bbayles: I like your second approach. I intentionally avoided using `len` here.

bbayles:
Many thanks for the suggestion, @coady.

Issue #13, opened 2012-08-18
Memory-happy `chunked`


piskvorky:
This ticket is a minor feature suggestion, not a bug/issue.

I've been using chunked for years, the difference being, the lists (chunks) that are yielded are often huge. The scenario is numerical computing. 

To allow releasing the memory more quickly, I don't keep a reference to the yielded object in `chunked`, so that it can be easily garbage collected (once all other outside references are gone).

Code:

``` python
def chunked(iterable, chunksize):
    """
    Return elements from the iterable in `chunksize`-ed lists. The last returned
    list may be smaller (if length of collection is not divisible by `chunksize`).

    >>> list(chunked(xrange(10), 3))
    [[0, 1, 2], [3, 4, 5], [6, 7, 8], [9]]
    """
    it = iter(iterable)
    while True:
        wrapped_chunk = [list(itertools.islice(it, chunksize))]
        if not wrapped_chunk[0]:
            break
        # memory opt: wrap the chunk and then pop(), to avoid leaving behind a reference
        yield wrapped_chunk.pop()
```

bbayles:
The performance hit is noticeable on some test runs - 10% - 15% difference.

This seems like a job that `weakref` could help with, but since our chunks are lists, that's more difficult.

```python
>>> from more_itertools.more import chunked, chunked_2
... from more_itertools import consume
... from string import ascii_lowercase
... from timeit import timeit
... 
... # chunked is original; chunked_2 has the wrapping/popping
... for f in (chunked, chunked_2):
...     def stmt():
...         iterable = ascii_lowercase * 100
...         n = 25
...         all_chunks = f(iterable, n)
...         consume(all_chunks)
...     
...     result = timeit(stmt, number=10000)
...     print(f.__name__, result)
chunked 1.3155306179833133
chunked_2 1.474021002999507
```

bbayles:
I can basically match the performance of the original by making a few hack-ish modifications:

```python
def chunked(iterable, chunksize):
    it = iter(iterable)
    wrapped_chunk = []  # Pre-specify the chunk wrapper
    append = wrapped_chunk.append  # Pull the dot lookup out of the loop
    pop = wrapped_chunk.pop  # Ditto
    while True:
        append(list(itertools.islice(it, chunksize)))  # Use `list(islice(...))` instead of `take(...)`
        if not wrapped_chunk[0]:
            break
        yield pop()
```

I'll make a PR with this change. Many thanks, @piskvorky.

PR #82, opened 2016-11-19
Simplified chunked.

coady:
Minor optimization as well.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -38,12 +38,7 @@ def chunked(iterable, n):
     the client.
 
     """
-    iterable = iter(iterable)
-    while True:
-        chunk = take(n, iterable)
-        if not chunk:
-            return
-        yield chunk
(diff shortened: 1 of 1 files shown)


bbayles:
@coady , thanks for the PR. This is a really nice one-liner, and actually does boost performance slightly in my tests.

I have PR #80 open for `chunked()` as well. I think this actually might help with the problem in issue #13 also. I will confirm and close that one.

bbayles:
Very impressed by this. To spell out the trick for posterity:
* This uses the 2-argument version of [iter](https://docs.python.org/3/library/functions.html#iter), which takes a callable and a sentinel. `iter()` will repeatedly call the callable's `__next__()` method until `sentinel` is returned.
* The callable in this case is `take()`, augmented with arguments via `partial()`.
* When `take()` exhausts the iterable, it will return `[]`, which is the sentinel, and the `iter()` call breaks.

PR #84, opened 2016-11-19
Simplified one.

coady:
Tuple unpacking is lazily evaluated and raises `ValueError`.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -321,10 +321,10 @@ def one(iterable):
     ...
     ValueError: too many values to unpack (expected 1)
 
-    >>> one([])
+    >>> one([])  # doctest: +IGNORE_EXCEPTION_DETAIL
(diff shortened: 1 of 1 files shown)


bbayles:
Thanks for the PR. As you can see, this syntax doesn't work on all supported versions of Python.

This method was actually used originally, but changed in d579071032c6bee1cbb5eaac174fda9c3f10fefa for compatibility's sake.

bbayles:
Oh, maybe the only issue is that the text of the ValueError string changed? That might make thsi nicer.

bbayles:
Yeah, I think that's the problem. If you change the second example in the docstring to this:

```python
    >>> one([])  # doctest: +IGNORE_EXCEPTION_DETAIL
    ...
    ValueError
```

then I will merge this.

bbayles:
Excellent, thanks!

PR #85, opened 2016-11-21
Simplified and optimized unique_to_each.

coady:


more_itertools/more.py
```diff
@@ -423,19 +423,10 @@ def unique_to_each(*iterables):
...
-            elements_to_indices.setdefault(element, set()).add(i)
-
-    for element, indices in iteritems(elements_to_indices):
-        if len(indices) != 1:
-            for i in indices:
-                while element in pool[i]:
-                    pool[i].remove(element)
-
-    return pool
+    counts = Counter(chain.from_iterable(map(set, pool)))
```
bbayles: We left `Counter` out originally because we were supporting Py2.6, but we're not at this point, so I think this is probably good.

bbayles:
Thanks for the PR. This looks promising! Could you post some `timeit` tests on this change? I'd also be interested to see if changing `filter` and `map` for their comprehension equivalents is any better.

That is, something like:
```python
def unique_to_each(*iterables):
    pool = [list(it) for it in iterables]
    counts = Counter(chain.from_iterable((set(it) for it in pool)))
    uniques = {element for element in counts if counts[element] == 1}

    return [list((x for x in it if x in uniques)) for it in pool]
```

bbayles:
I ran the tests in question; @coady's method is indeed faster than both the original and my proposed comprehension version.

```python
>>> from more_itertools.more import (
...     unique_to_each,  # original
...     unique_to_each_2,  # coady
...     unique_to_each_3,  # bbayles
... )
... from timeit import timeit
... 
... 
... for func in (unique_to_each, unique_to_each_2, unique_to_each_3):
...     def stmt():
...         func("mississippi", "missouri", "minnesota")  # sub in other tests as needed
... 
...     result = timeit(stmt, number=50000)
...     print(result)
1.2068535090002115
0.9786465030047111
1.0241529259947129
```

PR #89, opened 2016-12-03
Add line break to fix RTD docs

bbayles:
This PR fixes a docstring that looked OK on PythonHosted, but is incorrectly indented on RTD:

rtd-shot


erikrose:
"::" at the end of the preceding line is another, shorter way of accomplishing this; it lets you dispense with the backticks.

bbayles:
So that's what that does...!

Issue #88, opened 2016-12-03
Request: Faster `all_equal` for strings


pylang:
**Observation**

I understand the code for `all_equal` derives from the [itertools recipe](https://docs.python.org/3/library/itertools.html#itertools-recipes).   This legacy implementation has the benefit of working with generic iterables.  However, I came across this [SO post](http://stackoverflow.com/a/14321721/4531270), which shows an elegant implementation for the same operation on strings, that is, verify all letters in a string are equal.

```python
import itertools

import more_itertools

s = "aaaa"

%timeit more_itertools.all_equal(s)
1000000 loops, best of 3: 1.1 µs per loop

%timeit s == s[0] * len(s)
1000000 loops, best of 3: 438 ns per loop
```

We see the SO algorithm in this case is 2x-3x faster for strings.

**Request**

Can the SO algorithm be included in `more_itertools.all_equal` so that if a string is passed as the argument, this faster algorithm is preferred? 

For example:

```python
def all_equal(iterable):
    """
    Returns True if all the elements are equal to each other.
    
    Uses a faster implementation for strings. 
    http://stackoverflow.com/a/14321721/4531270
    
        >>> all_equal('aaaa')
        True
        >>> all_equal('aaab')
        False
        >>> all_equal([1,1,1,1])
        True
        >>> all_equal([1,1,1,0])
        False
        
    """
    if isinstance(iterable, str):
        s = iterable
        return s == s[0] * len(s)
    g = itertools.groupby(iterable)
    return next(g, True) and not next(g, False)
```

**Tests**

My local tests confirm these results:

```python
# New algorithm
all_equal("aaaa")
# True
all_equal("aaab")
# False

# Legacy algorithm
all_equal([1,1,1,1])
# True
all_equal([1,1,1,0])
# False
```
**Performance**

There is some improvement in speed over legacy with continued benefits for longer strings.

```
s = "a"*100000

# Legacy implementation
%timeit -n 1000 more_itertools.all_equal(s)
1000 loops, best of 3: 1.09 ms per loop
    
# Proposed implementation
%timeit -n 1000 all_equal(s)
1000 loops, best of 3: 9.64 µs per loop
```

bbayles:
Are your labels backward for the second set of performance numbers? It seems your "Proposed implementation" is much slower in that section.

pylang:
Thank you.  I corrected with values from a separate machine.

bbayles:
[...]

PR #90, opened 2016-12-06
Additional tests for all_equal recipe

bbayles:
Re: Issue #88, @pylang suggested adding some tests for the corner cases of `all_equal()`. I agree that this is a good idea; I've added tests for the "empty" and "one item" edge cases.


erikrose:
Lgtm

Issue #91, opened 2016-12-11
FEATURE: Stepped Sliding Window option


pylang:
**Abstract**

This feature request proposes extending functionality in `more_itertools.windowed` by producing windows separated by a given step.

Sliding windows are known for advancing a fixed length from one adjacent item to the next (e.g. `step=1`) continuously.  Can an option be made for implementing larger step sizes?
        
**Example**

```python
>>> from more_itertools import windowed

>>> iterable = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
        
# Present: Continuous Sliding
>>> all_windows = windowed(iterable, 3)
>>> list(all_windows)
[(1, 2, 3), (2, 3, 4), (3, 4, 5), (4, 5, 6), (5, 6, 7), (6, 7, 8), (7, 8, 9), (8, 9, 10)]

# Proposed: Steps
>>> stepped_windows = windowed(iterable, 3, step=2)
>>> list(stepped_windows)
[(1, 2, 3), (3, 4, 5), (5, 6, 7), (7, 8, 9)]
```

By default, the iterator returns windows spanning the length of the given iterable.  Therefore, sliding may stop abruptly for windows that are not equally divisible (e.g. leaving out item `10`).  However, if all items in the iterable are desired, there can be an option for including the tail and wrapping the results back to the head, as requested in this [SO Post](https://stackoverflow.com/questions/34359157/group-items-of-a-list-with-a-step-size-python):

```python       
# Proposed: Steps and Wrapping 
>>> stepped_windows = windowed(iterable, 3, step=2, wrap=True)
>>> list(stepped_windows)
[(1, 2, 3), (3, 4, 5), (5, 6, 7), (7, 8, 9), (9, 1, 2)]
```
**Further Considerations**

I believe I have a working prototype for this implementation with passing unittests and comparable performance times that I can post here if this is a plausible feature and you are interested in further discussion.

bbayles:
Thanks for the writeup. I'd be interested to see your implementation and timing. I'm inclined to think `step` might be useful, but the `wrap` option seems a bit much - can you give a motivating use case?

pylang:
As a brief response: sure, a bit later I will post what I have so far.  Secondly, the motivation comes form the stack overflow post.  The idea of `wrap` attempts to address the question, "what should a 'stepped' window do with remaining items in the generator?"  Ideas are welcome.

bbayles:
[...]

PR #92, opened 2016-12-11
Add step keyword to windowed()

bbayles:
Re: issue #91, this PR adds an optional `step` keyword to the `windowed()` function. It advances the window the specified number of places before emitting.

```python
>>> iterable = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
>>> list(windowed(iterable, 4, step=2))  # Each window will start +2 from the last start
[(1, 2, 3, 4), (3, 4, 5, 6), (5, 6, 7, 8), (7, 8, 9, 10)]
```

---

By default `step` is `1`, and the function behaves as it did before.

If `step >= n`, it's possible to slide past the edge of the iterable. In that case `fillvalue` will be inserted such that the size of the returned tuple is always `n`:
```python
>>> iterable = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
>>> list(windowed(iterable, 3, step=4))
[(1, 2, 3), (5, 6, 7), (9, 10, None)]
```

When `step == n`, this acts like a version of `chunked()` that pads the end with `fillvalue`:
```python
>>> iterable = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
>>> list(windowed(iterable, 4, step=4))
[(1, 2, 3, 4), (5, 6, 7, 8), (9, 10, None, None)]
```

---

The implementation is a bit different from the one in [this StackOverflow answer](http://stackoverflow.com/a/13408251/353839), which can emit an item full of fill values when `step > n`:

```python
>>> list(list(x) for x in so_version(range(1, 10 + 1), size=4, step=5))
[[1, 2, 3, 4], [6, 7, 8, 9], [None, None, None, None]]
```

In the issue we discussed wrapping around the iterable, but absent from real-world example from signal processing or something I'm inclined to leave it out.

Many thanks to @pylang for the suggestion.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -468,6 +468,8 @@ def windowed(seq, n, fillvalue=None, step=1):
     if n == 0:
         yield tuple()
         return
+    if step < 1:
+        raise ValueError('step must be >= 1')
 
     it = iter(seq)
     window = deque([], n)
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -488,9 +488,10 @@ def windowed(seq, n, fillvalue=None, step=1):
         if i % step == 0:
             yield tuple(window)
 
-    # If there are items left, fill with the given value
-    if i % step:
-        for _ in range(i, step):
+    # If there are items from the iterable in the window, pad with the given
+    # value and emit them.
+    if (i % step) and (step - i < n):
+        for _ in range(step - i):
             append(fillvalue)
         yield tuple(window)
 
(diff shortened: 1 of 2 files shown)


bbayles:
My plan is to merge this tomorrow.

pylang:
Just a reminder, you might consider reviewing some unittests from issue #91 to add before merging.   Many thanks.

bbayles:
Thanks. I think those cases are covered!
* The issue's `test_windowed_raise` is covered by `test_fillvalue`
* The issue's `test_windowed_step_1` is covered by `test_basic`
* The issue's `test_windowed_step_2` is covered by `test_step`
* The issue's `test_windowed_step_3` is covered by `test_step`

I will add you to the credits before the next release.

Issue #55, opened 2015-09-03
Support chunking with a fixed number of chunks


mozbhearsum:
Something I do relatively often is chunk data into a certain number of chunks, rather than chunks of a certain size. I find this useful in situations where I'm parallelizing work that depends on a bottleneck (eg, database, vcs server), and I want to ensure I don't overload it if a massive amount of work comes in. I have an implementation of this in https://github.com/bhearsum/chunkify, but it seems like it might fit well into chunked.

It would require an API break, with an interface such as:
def chunked(data, chunk_size=None, total_chunks=None)

...where one (and only one) of chunk_size or total_chunks is required. chunkify also supports only returning a specific chunk, which is a bit more efficient for large lists, but not crucial.

bbayles:
Thanks for the suggestion. I'm thinking wontfix for this, however. We have [grouper](https://github.com/erikrose/more-itertools/blob/ec7acde9e16aef7d84282d784633acc152f71e22/more_itertools/recipes.py#L187-L195), which is pretty darned close, yeah?

bbayles:
Reviving this issue - I was wrong! `grouper()` and `chunked()` are almost identical (save for padding); I was thrown by the example for `grouper()`.

I'm thinking of a function that will evenly distribute items from an iterable into multiple sub-iterables. Something like:

```python
>>> group_1, group_2 = distribute(2, [1, 2, 3, 4, 5, 6])
>>> list(group_1)
[1, 3, 5]
>>> list(group_2)
[2, 4, 6]
```

When there are 2 groups, the 0th item goes in the 0th group; the 1st item goes in the 1st group; the 2nd item goes in the 0th group; the 3rd item goes in the 1st group... The downside is that the sub-iterables don't have the same order as the parent, but this way we don't need to know the length of the parent.

I'll make a PR for this; apologies for the mistaken dismissal!

PR #93, opened 2016-12-13
New padded() itertool

bbayles:
This PR adds `padded()`, a wrapper that pads an iterable to at least `n` items (or to the next multiple of `n` items). This is a generalization of `padnone()` from the itertools recipes. I found myself wanting this when working on #55.

If you want to make sure your iterable has at least `n` items, you can ensure that missing values are substituted with a fill value:
```python
>>> list(padded([1, 2, 3], fillvalue='?', n=5))
 [1, 2, 3, '?', '?']
```

If you want to make sure your iterable has at least `k * n` items, you can ensure that as well:
```python
>>> list(padded([1, 2, 3, 4], n=3, multiple=True))
[1, 2, 3, 4, None, None]
```

With just the iterable specified, this will act just like `padnone()` and yield `None` indefinitely after the iterable is exhausted.

---

Is `multiple` a good keyword argument for padding to the nearest multiple of `n`? I considered `mod` as well.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -3,7 +3,7 @@ from __future__ import print_function
 from collections import Counter, defaultdict, deque
 from functools import partial, wraps
 from heapq import merge
-from itertools import chain, count, islice, takewhile
+from itertools import chain, count, islice, repeat, takewhile
 from sys import version_info
 
 from six import iteritems, string_types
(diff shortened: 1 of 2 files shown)


pylang:
I follow the first example, but I don't follow the second.  Particularly, how is `n=3` yielding a result of length 6?  And what is `multiple` doing exactly?

bbayles:
In the second example, `multiple` is `True`, so we pad to the next multiple of 3. Since there are 4 items, the next multiple is 6. So the padding produces two `None` items.

I'm open to suggestions for a better name for `multiple`.

pylang:
I'm not suggesting this is better, but `next_multiple` seems more clear.

bbayles:
Great suggestion; I've changed the keyword to `next_multiple` .

erikrose:
Lgtm. I especially like the name `padded`.

pylang:
Will this eventually replace and deprecate `padnone`?

bbayles:
I think not. The recipes in the Python documentation are a key part of `more-itertools`. In #20/#73 we improved one, but generally I think we should leave them in as-is.

pylang:
Ah yes.  I forgot it's recipe.

PR #94, opened 2016-12-15
New distribute() itertool

bbayles:
Re: Issue #55, this PR adds `distribute()`, a new itertool that splits an iterable into a fixed number of chunks. The items are distributed among the chunks as evenly as possible.

The `i`th item in the iterable goes in the `(i % n)`th chunk, so we have:

```python
>>> group_1, group_2 = distribute(2, [1, 2, 3, 4, 5, 6])
>>> list(group_1)
[1, 3, 5]
>>> list(group_2)
[2, 4, 6]
```

This allows the items to be distributed evenly without knowing the length of the input iterable.

When `n` doesn't match the length of the input iterable, the sizes of the chunks won't be identical:

```python
>>> children = distribute(3, [1, 2, 3, 4, 5, 6, 7])
>>> [list(c) for c in children]
[[1, 4, 7], [2, 5], [3, 6]]
```

The input iterable is `tee`d, so behind the scenes the items from the input iterable are being cached if you iterate through one returned sub-iterator before the others.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -3,7 +3,7 @@ from __future__ import print_function
 from collections import Counter, defaultdict, deque
 from functools import partial, wraps
 from heapq import merge
-from itertools import chain, count, islice, takewhile
+from itertools import chain, count, islice, takewhile, tee
 from sys import version_info
 
 from six import iteritems, string_types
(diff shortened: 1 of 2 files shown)


pylang:
Nice.  This seems to be a way to interleave between a set of chunks.

**Observations**

I get different behavior for `n=1` where consuming the assigned generator with `list()` doesn't evaluate the results (compared to the docstring example for `n=2`).  
```python
>>> g = distribute(1, [1, 2, 3, 4, 5, 6])
>>> list(g)
[<generator object distribute.<locals>._iterator at 0x0000000004C11468>]       # expect numbers
```


**Questions**
1.  What does distributing to `n=1` chunks actually mean?  I would expect this to return itself.   
Is this what you want, or instead ignore distributing one chunk and raise an error if `n < 2`?  
2.  Would it be more memory-friendly to return a generator expression instead of a list comprehension, i.e. `return (_iterator(index) for index in range(n))`?

bbayles:
__Re: n=1 case__: If you want to get back the original for `n=1`, just put a comma after the assignment:
```python
>>> group_1, = distribute(1, [1, 2, 3, 4, 5, 6])
>>> list(group_1)
1: [1, 2, 3, 4, 5, 6]
```

`distribute()` must return an iterable of iterables, so you can use unpacking if you hard-code `n`.

---

__Re: return a generator expression__: I don't think this is necessary. `n` is unlikely to be large, and each iterable in the returned list is itself pretty compact.

Also, this would prevent slicing the returned value. I could see returning a `tuple` instead of a `list`.

bbayles:
Planning to re-read and merge this tomorrow.

PR #95, opened 2016-12-18
Allow negative indexing and arbitrary slicing for peekable()

bbayles:
`peekable()`-wrapped iterables supports indexing and slicing as of PR #71. They raise an exception when using negative indexes or slices with negative bounds, but are otherwise supposed to behave like normal sequences when indexed/sliced.

However, I missed a case: when using a slice with start > stop and step < 0, the peekable doesn't slice properly:
```python
>>> iterable = list(range(10))
>>> p = peekable(iterable)
>>> iterable[8: 1: -3]
[8, 5, 2]
>>> p[8: 1: -3]
[]
```

This PR fixes that, and in the process removes the restriction on negative indexes and slices with negative bounds.

I think I got this right, but off-by-one errors are easy to stumble into here!

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -189,7 +189,7 @@ class peekable(object):
         cache_len = len(self._cache)
         if index < 0:
             self._cache.extend(self._it)
-        elif index + 1 >= cache_len:
+        elif index >= cache_len:
             self._cache.extend(islice(self._it, index + 1 - cache_len))
 
         return self._cache[index]
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -181,7 +181,14 @@ class peekable(object):
         if isinstance(index, slice):
             return self._get_slice(index)
 
-        return self._get_slice(slice(index, index + 1, None))[0]
+        cache_len = len(self._cache)
+
+        if index < 0:
+            raise ValueError('Negative indexing not supported')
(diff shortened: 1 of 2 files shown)


more_itertools/more.py
```diff
@@ -107,6 +107,10 @@ class peekable(object):
         >>> next(p)
         'b'
 
+    Negative indexes are supported, but be aware that they will cache the
```
bbayles: Obviously don't try to use a negative index on an infinite iterator.

more_itertools/more.py
```diff
@@ -181,7 +188,13 @@ def __getitem__(self, index):
         if isinstance(index, slice):
             return self._get_slice(index)
 
-        return self._get_slice(slice(index, index + 1, None))[0]
+        cache_len = len(self._cache)
+        if index < 0:
```
erikrose: Looks good to me. I twiddled around with it but couldn't find any off-by-ones. I *feel* like this if/elif wants to be made one with the similar one in _get_slice(), but it's not obvious that it would result in a simplification. If you don't immediately say "Of course! Like this!", then let's declare victory and merge as-is. Nice work, yet again!

bbayles: Ah, yes. I had those two things together originally, but pulled them apart in 70c7767, since a `None` slice boundary has different meaning depending on the sign of step.

PR #97, opened 2016-12-22
New stagger() and zip_offsets() itertools

bbayles:
This addition is adapted from @joshbode's [ActiveState recipe](https://code.activestate.com/recipes/577852-iterator-offsetter/?in=user-4179046).

__Updated__, see thread below.

---

~~`offset_groups()`~~ `stagger()` lets you specify a set of offsets by which to lead or lag the iterable. It will yield groups of items, the `i`th of which corresponds to the `i`th offset. This is a sort of generalization of a sliding window.

For example, to get a sliding window with both lookback and lookahead:
```python
>>> list(stagger('abcdefg', offsets=(-2, -1, 0, 1)))
[(None, None, 'a', 'b'),
 (None, 'a', 'b', 'c'),
 ('a', 'b', 'c', 'd'),
 ('b', 'c', 'd', 'e'),
 ('c', 'd', 'e', 'f'),
 ('d', 'e', 'f', 'g')]
```

The `longest` parameter will continue until the last item of the iterable is the first item of the last group:
```python
>>> list(stagger('abcdefg', offsets=(0, 1, 2), longest=True))
[('a', 'b', 'c'),
 ('b', 'c', 'd'),
 ('c', 'd', 'e'),
 ('d', 'e', 'f'),
 ('e', 'f', 'g'),
 ('f', 'g', None),
 ('g', None, None)]
```

`fillvalue` can be used in place of `None` as well. `offsets` is `(-1, 0, 1)` by default to get `(before, during after`) groups:
```python
>>> for before, during, after in stagger('abcd', longest=True, fillvalue='?'):
...     print(before, during, after, sep='\t')
?	a	b
a	b	c
b	c	d
c	d	?
d	?	?
```

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -874,6 +874,9 @@ def stagger(iterable, offsets=(-1, 0, 1), longest=False, fillvalue=None):
         >>> list(stagger([0, 1, 2, 3], longest=True))
         [(None, 0, 1), (0, 1, 2), (1, 2, 3), (2, 3, None), (3, None, None)]
 
+    By default, ``None`` will be used to replace offsets beyond the end of the
+    sequence. Specify *fillvalue* to use some other value.
+
     """
     children = tee(iterable, len(offsets))
 
@@ -888,6 +891,9 @@ def zip_offset(*iterables, **kwargs):
 
         >>> list(zip_offset('0123', 'abcdef', offsets=(0, 1)))
         [('0', 'b'), ('1', 'c'), ('2', 'd'), ('3', 'e')]
+
+    This can be used as a lightweight alternative to SciPy or pandas to analyze
+    data sets in which somes series have a lead or lag relationship.
 
     By default, the sequence will end when the shortest iterable is exhausted.
     To continue until the longest iterable is exhausted, set *longest* to
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -37,6 +37,7 @@ __all__ = [
     'unique_to_each',
     'windowed',
     'with_iter',
+    'zip_offset',
 ]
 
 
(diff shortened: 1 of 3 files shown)


pylang:
This is a nice feature.  It looks similar to `windowed`,  where `offsets` is a verbose form of `n`.   I see this as a `windowed_offset`.

joshbode:
Not sure if this is an issue, but `zip(*iterables)` is not lazy in Python 2 (though it is in Python 3)
which is why I went for `izip(*iterables)` in the original recipe.

## Python 2
```python
In [1]: import itertools

In [2]: x = range(10)

In [3]: iterables = itertools.tee(x, 3)

In [4]: zipped = zip(*iterables)

In [5]: zipped
Out[5]:
[(0, 0, 0),
 (1, 1, 1),
 (2, 2, 2),
 (3, 3, 3),
 (4, 4, 4),
 (5, 5, 5),
 (6, 6, 6),
 (7, 7, 7),
 (8, 8, 8),
 (9, 9, 9)]
```

## Python 3
```
In [4]: zipped = zip(*iterables)

In [5]: zipped
Out[5]: <zip at 0x106522988>

In [6]: next(zipped)
Out[6]: (0, 0, 0)

In [7]: next(iterables[1])
Out[7]: 1

In [8]: next(zipped)
Out[8]: (1, 2, 1)
```

joshbode:
scratch that - looks like you're using `zip/zip_longest` from `six.moves` which is `itertools.izip/itertools.izip_longest` anyway.

bbayles:
You're quite right; it takes some effort to be lazy in this case.

bbayles:
Also, I'm open to a different name for this itertool. The original was `offsetter`; I went with `offset_groups`.

`windowed_offset` is suggested in this thread. I think of a "window"`as being continuous, which is not necessarily the case with this, so I think I'll steer clear of that.

It makes me thing of a "comb" function and the process of convolution from signal processing, but `comb_slide` is a weird name.

joshbode:
Another possibility: `stagger` as in a "staggered start" for a race/event (like a marathon), though the analogy breaks down since there's more than one racer, but only one `tee`d iterable here.

Which makes me wonder if a useful generalisation to this function would be just that: a function to apply a list of offsets against a list of distinct iterables (rather than clones like here) - the function in this PR is then a special case.

bbayles:
Re: name, I like `stagger`! I asked my colleague, who is too cool for Github, and he also liked it. So I'll push that change shortly.

Re: generalization, this is a good idea. This one is actually easy to name - `zip_offset()`.

My only issue is that it's really tough to write a docstring that covers the general case as well as the specific case concisely.

My thought is to make two functions: `stagger()` and `zip_offset()` - the former just calls the latter. Any objections?

joshbode:
Excellent :)

My thinking was that the generalised function could be called `stagger` and that the specialised cloning version would call it, i.e. `stagger(tee(iterable, len(offsets)), offsets=offsets)`

more_itertools/more.py
```diff
@@ -12,11 +12,32 @@
 from .recipes import take
 
 __all__ = [
-    'bucket', 'chunked', 'collapse', 'collate', 'consumer',
-    'distinct_permutations', 'distribute', 'first', 'ilen',
-    'interleave_longest', 'interleave', 'intersperse', 'iterate', 'one',
-    'padded', 'peekable', 'side_effect', 'sliced', 'split_after',
-    'split_before', 'spy', 'unique_to_each', 'windowed', 'with_iter',
+    'bucket',
```
bbayles: Some hate this style; others get tired of re-formatting things... :parasol_on_ground:

bbayles:
Mmhmm; I'm attached to `stagger()` being the cloning method and `zip_offset()` being the general method.

Very glad for the collaboration here; I really like how this has ended up!

joshbode:
Yeah! I don't feel too strongly about it being one way or the other, in any case :)

joshbode:
Plus, `itertools` might just be my favourite package in the standard library, so it is good to see this effort which I wasn't familiar with.

pylang:
What implementations do you have in mind for `stagger()` and `zip_offset()`?

bbayles:
Applications? As Josh said in an e-mail:
> I find myself reimplementing this function in any new language I'm using (did pretty much the same thing in Julia a few weeks ago) - for me, this function is key to doing simple lead/lag-type series analysis without resorting to something like pandas.

You can use `stagger()` to compute things like the [number of points on surface of tetrahedron](http://oeis.org/A005893):
```python
... from more_itertools import stagger, take
... 
... squares = (n * n for n in count())
... points = (a + b for a, b in stagger(squares, offsets=(0, 2)))
... list(take(10, points))
[4, 10, 20, 34, 52, 74, 100, 130, 164, 202]
```

(though there are easier ways to do that...) I could also see it being useful for, e.g. [Project Euler 263](https://projecteuler.net/problem=263)...

pylang:
Please disregard.  I see now the implementations are already committed.  Thank you.

more_itertools/more.py
```diff
@@ -834,3 +855,68 @@ def _iterator(index):
             i = (i + 1) % n
 
     return [_iterator(index) for index in range(n)]
+
+
+def stagger(iterable, offsets=(-1, 0, 1), longest=False, fillvalue=None):
+    """Yield tuples whose elements from are offset from *iterable*.
```
pylang: "are an offset from"

bbayles: Good catch, thanks.

pylang:
At the moment `zip_offset` and `staggered` are quite similar.  It took me a while to distinguish the difference.  For clarity, perhaps make in `staggered` this is a specialized case of `zip_offset` applied to one iterable.

bbayles:
They look reasonably distinct to me when rendered, especially since the first sentence of `zip_offset()`'s docstring says "`zip` the input *iterables* together" - what do you think?

image

image

pylang:
I've been staring at this for a while.  I'll have to come back with fresh eyes.  Thanks for the reply.

pylang:
The rendering does help.  For me, I think I would use `zip_offset` alone.

erikrose:
Looks like a good consensus! My only suggestion is to include in the docstring one example of a use case for it, perhaps the one from Josh's email. I like to do that for any of our more complicated routines.

bbayles:
Yep, good idea.

PR #98, opened 2016-12-22
Simply intersperse()

bbayles:
This PR changes the implementation of `intersperse()` to take out the Python loop and replace it with C itertools. My testing shows it to be 15%-20% faster on a 1000-item iterable.

---

The reason this implementation works is that interspersing is close to `zip`-ing.

If we create an infinite iterable interspersed element `e` with `repeat`, we get this when we `zip` it with the our iterable:
```python
>>> from itertools import repeat
...
... iterable = [0, 1, 2, 3]
... e = '?'
... filler = repeat(e)
... list(zip(filler, iter(iterable)))
[('?', 0), ('?', 1), ('?', 2), ('?', 3)]
```

If we `flatten` that, we get almost what we want - the only difference is an extra `e` at the beginning:
```python
>>> from itertools import chain
...
... list(chain.from_iterable([('?', 0), ('?', 1), ('?', 2), ('?', 3)]))
['?', 0, '?', 1, '?', 2, '?', 3]
```
We can chop that off with `islice`:
```python
>>> from itertools import islice
...
... list(islice(['?', 0, '?', 1, '?', 2, '?', 3], 1, None))
[0, '?', 1, '?', 2, '?', 3]
```

Which is what we want.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -9,7 +9,7 @@ from sys import version_info
 from six import string_types
 from six.moves import filter, map, zip, zip_longest
 
-from .recipes import take
+from .recipes import flatten, take
 
 __all__ = [
     'bucket', 'chunked', 'collapse', 'collate', 'consumer',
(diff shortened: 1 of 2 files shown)


pylang:
zipping an iterator advanced with `next()`.  Nice!

bbayles:
Your comment inspired me to check if doing `next(...)` on the `flatten`-ed iterable is faster. It is indeed! Another 5%-10% boost.

erikrose:
Not only a speed improvement but a conciseness one as well, especially in the tests. Lgtm!

pylang:
Score!

PR #102, opened 2016-12-29
Simplified distribute.

coady:


bbayles:
Nicely done, as always!

PR #104, opened 2017-01-04
Updates for 2.5.0

bbayles:
Five new itertools and some optimizations are in - I think it's time for a release.

This PR bumps the version to 2.5.0 (we went to three digit versions in 2.4.1; I figured I'd continue that).

Many thanks to all listed here - keep the ideas coming.


bbayles:
Excellent. I'll push out to PyPI with a wheel.

PR #105, opened 2017-01-08
New divide() itertool

bbayles:
Sort of re: issue #55, this PR adds `divide()` a counterpart to `distribute()` that maintains order.

I thought about doing this along with `distribute()`, but didn't like having to know the length of the input iterable / store the whole thing to find it.

However, I needed "chunk into N parts, maintaining order" yesterday. I found myself reaching for `more-itertools`, and regretted my decision.

I've made this exactly match `distribute()` in that it returns a list of iterables, such that one can be swapped for the other. (If this were on its own I might yield slices instead)

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -1,4 +1,4 @@
-from __future__ import division, print_function
+from __future__ import print_function
 
 from collections import Counter, defaultdict, deque
 from functools import partial, wraps
@@ -846,7 +846,7 @@ def distribute(n, iterable):
         >>> [list(c) for c in children]
         [[1], [2], [3], [], []]
 
-    This function uses ``itertools.tee``, and may require significant storage.
+    This function uses ``itertools.tee`` and may require significant storage.
     If you need the order items in the smaller iterables to match the original
     iterable, see ``divide()``.
 
@@ -986,7 +986,7 @@ def divide(n, iterable):
         >>> [list(c) for c in children]
         [[1], [2], [3], [], []]
 
-    This function will exhaust the iterable before returning, and may require
+    This function will exhaust the iterable before returning and may require
     significant storage.
     If order is not important, see ``distribute()``, which does not first pull
     the iterable into memory.
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -761,7 +761,7 @@ def split_before(iterable, pred):
 
     """
     buf = []
-    for item in iter(iterable):
+    for item in iterable:
         if pred(item) and buf:
             yield buf
             buf = []
@@ -781,7 +781,7 @@ def split_after(iterable, pred):
 
     """
     buf = []
-    for item in iter(iterable):
+    for item in iterable:
         buf.append(item)
         if pred(item) and buf:
             yield buf
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -1,4 +1,4 @@
-from __future__ import print_function
+from __future__ import division, print_function
 
 from collections import Counter, defaultdict, deque
 from functools import partial, wraps
@@ -20,6 +20,7 @@ __all__ = [
     'consumer',
     'distinct_permutations',
     'distribute',
+    'divide',
     'first',
     'ilen',
     'interleave_longest',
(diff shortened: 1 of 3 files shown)


more_itertools/more.py
```diff
@@ -957,3 +960,48 @@ def sort_together(iterables, key_list=(0,), reverse=False):
...
+        [[1, 2, 3], [4, 5], [6, 7]]
+
+    If the length of the iterable is smaller than n, then the last returned
+    iterables will be empty:
+
+        >>> children = divide(5, [1, 2, 3])
+        >>> [list(c) for c in children]
+        [[1], [2], [3], [], []]
+
+    This function will exhaust the iterable before returning, and may require
```
erikrose: No comma because not a compound sentence

more_itertools/more.py
```diff
@@ -957,3 +960,48 @@ def sort_together(iterables, key_list=(0,), reverse=False):
...
+    significant storage.
+    If order is not important, see ``distribute()``, which does not first pull
```
erikrose: Let's either skip a line between this and the above or else join them into a single paragraph.

more_itertools/more.py
```diff
@@ -957,3 +960,48 @@ def sort_together(iterables, key_list=(0,), reverse=False):
...
+        raise ValueError('n must be at least 1')
+
+    seq = tuple(iterable)
+    q, r = divmod(len(seq), n)
+
+    ret = []
+    for i in range(n):
+        start = (i * q) + (i if i < r else r)
+        stop = ((i + 1) * q) + (i + 1 if i + 1 < r else r)
+        ret.append(iter(seq[start:stop]))
```
erikrose: How important do you think `iter()` is here? I suppose it depends on how you expect the caller to use the result. If you expect him to run `for` over the items in the returned list, `iter` should be unnecessary, since `for` expects "iterable objects", not iterators themselves (i.e., it calls `__iter__` itself). I just noticed one or two other places in the code where I suspect we call `iter` unnecessarily, like in `collapse`. Thoughts?

more_itertools/more.py
erikrose: We actually do the same thing in distribute().

more_itertools/more.py
```diff
@@ -1,4 +1,4 @@
-from __future__ import print_function
+from __future__ import division, print_function
```
erikrose: Did you actually end up using the `/` operator? I don't see it.

more_itertools/more.py
bbayles: I like to break markdown and reST at sentence boundaries, since when rendered single-breaks get joined. But I'll change it for direct source viewing.

more_itertools/more.py
bbayles: I did this intentionally to match what `distribute` returns, which are un-materialized iterator objects.

Re: collapse, I think [this `iter()`](https://github.com/erikrose/more-itertools/blob/51ecf3ea59784ffd8725c7d42968c112e788b539/more_itertools/more.py#L680) is important, because it is what raises the subsequent `TypeError` for leaf nodes, right?

`split_before()` and `split_after()` might not need their `iter()`s, though.

bbayles: I did this intentionally to match what `distribute` returns, which are un-materialized iterator objects.

Re: collapse, I think [this `iter()`](https://github.com/erikrose/more-itertools/blob/51ecf3ea59784ffd8725c7d42968c112e788b539/more_itertools/more.py#L680) is important, because it is what raises the subsequent `TypeError` for leaf nodes, right?

`split_before()` and `split_after()` might not need their `iter()`s, though.

more_itertools/more.py
bbayles: I couldn't remember whether `//` was a first-class citizen in Py2.7; it seems it is.

erikrose:
Brandon Rhodes likes to break his text down even further, into phrases. :-)

more_itertools/more.py
erikrose: Hmm. It's true `distribute` returns iterators that don't materialize their entire runs in memory. But here in `divide`, `seq` is a tuple, and slicing a tuple immediately yields another (fully materialized) tuple. Passing it to `iter` doesn't make it less material; it just does the work of calling `__iter__` on it, turning it from an iterable to an iterator. And `distribute` returns an iterator, not an iterable, so can I see doing this for consistency. If that's what you're shooting for, then LGTM.

I see what you mean in `collapse`. That makes sense.

Issue #106, opened 2017-01-15
Request: pushback


diazona:
I'd like to suggest adding a wrapper that allows pushing a value back on to an iterator, so that the next call to `next(it)` will return the pushed value before the next element from the underlying iterable. I find myself wanting this from time to time (usually in parsing applications), and I could have sworn it was implemented somewhere standard, but I looked around and couldn't find it. Would this be a good addition to more-itertools?

I do have code to offer, but I'm posing this as an issue instead of a pull request because I have a dilemma. I've come up with two implementations, one [as a generator function](https://github.com/diazona/more-itertools/tree/pushback_function)
```python
def pushback(iterable, maxlen=None):
    iterable = iter(iterable)
    # add 1 to account for the append(None)
    stack = deque(maxlen=maxlen + 1 if maxlen is not None else None)
    while True:
        if stack:
            e = stack.pop()
        else:
            e = next(iterable)
        sent = yield e
        if sent is not None:
            stack.append(sent)
            stack.append(None) # dummy value to return from send()
```
and the other [as a class](https://github.com/diazona/more-itertools/tree/pushback_class)
```python
class pushback:
    def __init__(self, iterable, maxlen=None):
        self.iterable = iter(iterable)
        self.stack = deque(maxlen=maxlen)
    def __iter__(self):
        return self
    def __next__(self):
        return self.stack.pop() if self.stack else next(self.iterable)
    def send(self, value):
        self.stack.append(value)
```
The function implementation is about twice as fast in my preliminary tests (using IPython)
```
In [13]: %timeit list(pushback_function(range(10)))
100000 loops, best of 3: 5.45 µs per loop
In [14]: %timeit list(pushback_class(range(10)))
100000 loops, best of 3: 10.8 µs per loop
```
On the other hand the class implementation is conceptually cleaner, and also does not need to be "primed" by calling `next(it)` before sending in a value with `it.send(x)`.

Now, in most cases, you can prime the generator iterator without losing an item by running `it.send(next(it))`, and that could be done in a wrapper function to make it transparent to client code. But only the class implementation allows pushing in front of an empty iterable (admittedly a rather pathological use case):
```python console
>>> it = pushback([])
>>> it.send(10)
>>> list(it)
[10]
```
So my point is: if this is something you want for more-itertools, which implementation to use? Or is there a way to "fix" one of them to make it strictly better than the other, that I'm not seeing? (Or does this whole thing already exist and I wasted an evening?)

[...]

PR #107, opened 2017-01-16
Adding a prepend() method to peekable

diazona:
As discussed in #106. This does not use `reversed()`, so `p.prepend(a, b, c)` is equivalent to
```python
p.prepend(c)
p.prepend(b)
p.prepend(a)
```
but that can be changed.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -210,12 +210,16 @@ class peekable(object):
 
             >>> p = peekable([])
             >>> next(p)
-            StopIteration:
+            Traceback (most recent call last):
+              ...
+            StopIteration
             >>> p.prepend(1)
             >>> next(p)
             1
             >>> next(p)
-            StopIteration:
+            Traceback (most recent call last):
+              ...
+            StopIteration
 
         """
         self._cache.extendleft(items)
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -175,6 +175,11 @@ class peekable(object):
                 return default
         return self._cache[0]
 
+    def prepend(self, *items):
+        """Schedule items to be the next ones returned from ``next()`` or
+        ``self.peek()``."""
+        self._cache.extendleft(items)
+
     def __next__(self):
         if self._cache:
             return self._cache.popleft()
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -208,25 +208,6 @@ class peekable(object):
             >>> list(p)
             [11, 12, 1, 2, 3]
 
-        If you want to prepend elements such that the last one in the argument
-        list is the first to be returned, use ``reversed()``,
-
-            >>> p = peekable([1, 2, 3])
-            >>> p.prepend(*reversed((10, 11, 12)))
-            >>> list(p)
-            [12, 11, 10, 1, 2, 3]
-
-        or use separate calls to ``prepend()`` with one element each:
-
-            >>> p = peekable([1, 2, 3])
-            >>> p.prepend(10)
-            >>> p.prepend(11)
-            >>> p.prepend(12)
-            >>> list(p)
-            [12, 11, 10, 1, 2, 3]
-
-        This last example could also be wrapped up with a `for` loop.
-
         It is possible, by prepending items, to "resurrect" a peekable that
         previously raised ``StopIteration``.
 
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -95,7 +95,7 @@ def first(iterable, default=_marker):
 
 
 class peekable(object):
-    """Wrap an iterator to allow lookahead.
+    """Wrap an iterator to allow lookahead and prepending elements.
 
     Call ``peek()`` on the result to get the value that will next pop out of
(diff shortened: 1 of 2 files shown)


more_itertools/more.py
```diff
@@ -175,6 +197,55 @@ def peek(self, default=_marker):
...
+        ``self.peek()``. The items will be returned in first-in first-out order:
+
+            >>> p = peekable([1, 2, 3])
+            >>> p.prepend(10, 11, 12)
+            >>> next(p)
+            10
+            >>> list(p)
+            [11, 12, 1, 2, 3]
+
+        If you want to prepend elements such that the last one in the argument
```
bbayles: Could you remove this block? Starting with "If you want to prepend elements..." and ending with "...could be wrapped in a `for` loop". I don't think we need to remind people that to reverse something that they should use `reverse`... :business_suit_levitating: 

Also, the example just above makes it clear how things work.

more_itertools/more.py
```diff
@@ -175,6 +197,55 @@ def peek(self, default=_marker):
...
+            >>> p.prepend(1)
+            >>> next(p)
+            1
+            >>> next(p)
+            Traceback (most recent call last):
+              ...
+            StopIteration
+
+        """
+        self._cache.extendleft(reversed(items))
```
bbayles: Could you make sure `items` is wrapped by `list`? This will not work with, say, `peekable([]).prepend((x for x in '123'))` because `reversed` won't be able to operate.

more_itertools/more.py
bbayles: Could you remove this block? Starting with "If you want to prepend elements..." and ending with "...could be wrapped in a `for` loop". I don't think we need to remind people that to reverse something that they should use `reverse`... :business_suit_levitating: 

Also, the example just above makes it clear how things work.

more_itertools/more.py
bbayles: Also, could you add a test for this case?

more_itertools/tests/test_more.py
```diff
@@ -177,6 +177,147 @@ def test_slicing_reset(self):
...
+        actual = list(peekable(expected))
+        eq_(actual, expected)
+
+    # prepend() behavior tests
+
+    def test_prepend(self):
+        """Test calling ``prepend()`` at various points interspersed with calls
+        to ``next()``"""
+        it = peekable(range(5))
+        it.prepend(10)
```
bbayles: Could these go in a loop? I'm usually OK with violating DRY in tests, but I think this would be easier to understand in a loop. That, or do this test with a smaller iterable?

more_itertools/more.py
diazona: 1. I actually had the same thought you did (assuming you meant `prepend(*(x for x in '123'))`), and I tested that interaction. It does work, presumably because argument unpacking consumes the iterable and turns it into a tuple. The only case where this causes trouble with `reversed()` is if the caller wants to reverse the order of the elements in their iterable before passing it into `prepend()`, in which case they'll have to use `reversed(list((x for x in '123')))` or equivalent, but that's not something that we can solve in the implementation of `prepend()`.
2. Sure. Actually this is exactly what `test_prepend_iterable` was supposed to handle, but I should probably use a generator rather than a `range` to make sure there are no "secret" optimizations going on.

bbayles: Touché - I should have checked!

more_itertools/more.py
diazona: Well, I put it in there precisely because I do think it's useful to remind people of this trick - people who might otherwise complain about `prepend()` not working in the other direction, or not even having an option to do so. It's obvious that you can reverse an iterable with `reversed()`, but not so obvious that you can reverse the order of an argument list with the same technique. I'm not strongly _opposed_ to removing this block, but it's not something I put in just on a whim either. I'm trying to anticipate how people who haven't read this discussion might respond to the implementation choices we're making.

more_itertools/tests/test_more.py
diazona: Will do, good suggestion.

more_itertools/more.py
bbayles: I'd still rather not spell it out - note how the [the deque docs](https://docs.python.org/3/library/collections.html#collections.deque.extendleft) handle this. Since the ordering might be surprising they mention it, but leave the handling of it to the reader.

If you still think it's important, could you pick one method - either `reverse` or `prepend`-ing single items in a loop? If we're going to explain it, then I'd rather do it one way than two (PEP 20 styles).

diazona: Eh, good point. OK I'll remove that. I'll add another commit shortly, which should address everything.

diazona:
I think this latest commit should address your comments, but let me know if there's anything else or if I missed any required changes.

bbayles:
No more objections from me. I'll leave this open for comment for a day or two and then merge. I'll also add you to the credits before the next release. Many thanks!

diazona:
Sounds good! Thanks to you too.

PR #108, opened 2017-01-19
always_iterable(), round two

bbayles:
This PR builds on #37, and brings in `always_iterable()`. Since we're using `six` now, I've updated it to use that. I've also tweaked the docstring and added unit tests.

I'll close the older PR and merge this, provided there are no objections. Many thanks to @jaraco.


more_itertools/more.py
```diff
@@ -1004,3 +1004,53 @@ def divide(n, iterable):
...
+        >>> item_sum(10)
+        10
+        >>> item_sum([10, 20])
+        30
+
+    """
+    if obj is None:
+        return ()
+
+    string_like_types = (text_type, binary_type)
```
bbayles: We had [some discussion](https://github.com/erikrose/more-itertools/pull/37#issuecomment-39687015) about this way back when. I've got this using `text_type` and `binary_type` here, so we'll get `(unicode, str)` on Python 2 and `(str, bytes)` on Python 3.

more_itertools/more.py
```diff
@@ -1004,3 +1004,53 @@ def divide(n, iterable):
...
+    """
+    Given an object, always return an iterable.
+
+    If the object is not already iterable, return a tuple containing containing
+    the object::
+
+        >>> always_iterable(1)
+        (1,)
+
+    If the object is ``None``, return an empty iterable::
```
bbayles: I almost removed this special case in order to simplify the docstring... "Given an object, always return an iterable. If the object is not already iterable, wrap it in a tuple and return it. Otherwise, return the object."

I could argue both ways for whether `None` should return `(None,)` or whether it should return `()`. Anybody feel strongly?

jaraco: In my experience, empty for None is almost always what you want. I've never encountered a case where one would want None in a tuple. I'll dig up some examples.

bbayles: I found some examples (usually bare `return` clauses) where this would be helpful; I think I'm fine with the idea as you contributed it.

bbayles:
Glad to get this in; @thanks jaraco.

Issue #109, opened 2017-02-03
Tool suggestion: identify elements adjacent to those matching a predicate


diazona:
I just came up with another function that I don't _think_ already exists and might make a useful addition to more-itertools: identifying elements in an iterable that match a predicate or are adjacent to those matching a predicate. As a stupid (but simple) example, suppose I want to detect which letters are vowels or adjacent to vowels in a word. The design I have in mind returns a tuple for each element in the iterable with a boolean indicating whether it is or is adjacent to a "selected" element, as well as the element itself.
```python console
>>> list(adjacent(lambda c: c in 'aeiou', 'thursday'))
[(False, 't'), (True, 'h'), (True, 'u'), (True, 'r'), (False, 's'), (True, 'd'), (True, 'a'), (True, 'y')]
```
In my application it's important to know where the (equivalents of the) vowel-centered groups end and begin, in addition to knowing which elements are in those groups, so I pass the result of `adjacent()` through `groupby()`. (This is akin to choosing context lines in a differ.) If I wanted an iterable of just `['h', 'u', 'r', 'd', 'a', 'y']` in that example, I could use instead use `filter()` and `map()` or a generator, i.e. `(e[1] for e in adjacent(...) if e[0])`. I think this flexibility is important.

The question I want to bring up before sending in a pull request is how to generalize this. My base implementation is the following:
```python
def adjacent(predicate, iterable):
    i1, i2 = tee(iterable)
    selected = chain([False], map(predicate, i1), [False])
    adjacent = map(any, windowed(selected, 3))
    return zip(adjacent, i2)
```
(designed to avoid calling `predicate()` more than once per item) It's easy enough to change the number of elements of "context" by increasing the second argument to `windowed()`, and that would be a straightforward generalization. Is it also worthwhile to support arbitrary "masks" by using `stagger()` instead of `windowed()`? E.g. passing `offsets=(-1, 1)` to "mark" only elements which are before or after those which satisfy the predicate, leaving out the ones which do satisfy the predicate themselves? Or `offsets=(0,1,2)` to "mark" items which satisfy the predicate and the two that follow them?

bbayles:
My first impression was that this was a straightforward application of `windowed`. However, after playing with it for a bit in the interpreter, I couldn't come up with a more elegant method than yours; this is tricky to get right without calling `predicate` more frequently.

So, I'm open to a PR for this. For generalizations, I'm good with an optional `n` argument (I guess verify that it's at least 3?).

[...]

PR #110, opened 2017-02-08
adjacent() and groupby_transform()

diazona:
From issue #109, this PR implements `adjacent()` and `groupby_transform()`.

It's mostly straightforward, except that the one possibly-contentious design decision I made was to have `groupby_transform()` use default key and value transformations of `itemgetter(0)` and `itemgetter(1)` respectively, so you can call
```python
groupby_transform(zip(keys, values))
```
to group one iterable by the elements of another. That seems like the use case where this tool is going to be by far the most useful. And it's not like we lose any functionality, since without this special case, the only sensible behavior for default arguments (as far as I can imagine) is to just pass through to `groupby()`, in which case people should just use `groupby()` directly.

That being said, it does seem rather weird, especially given the name of the function. And it's not unimaginable that this special-case behavior could complicate things in some cases. And yes, "explicit is better than implicit", though I could counter with "practicality beats purity" ;-) Anyway, the point is, I'm fine with it if you think it's better to omit this special case and just have the default arguments emulate `groupby()`, and make people explicitly pass `itemgetter(0)` and `itemgetter(1)` whenever they want to group one iterable by another. I just think if we do that, it's a missed opportunity to simplify potentially a lot of client code.

Oh, and I'm also not thrilled with the test coverage, though I couldn't really think of any other useful tests to add.

(I'm also not attached to either of these function names.)

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -1178,12 +1178,13 @@ def groupby_transform(iterable, keyfunc=None, valuefunc=None):
     If ``None`` or an identity function is passed for *valuefunc*, the behavior
     is identical to ``itertools.groupby()``.
     """
-    if valuefunc is None:
-        if keyfunc is None:
+    if valuefunc is None and keyfunc is not None:
+        for key, group in groupby(iterable, keyfunc):
+            yield key, group
+    else:
+        if valuefunc is None and keyfunc is None:
             # itemgetter is about 2.2x as fast as lambda functions in some tests
             keyfunc = itemgetter(0)
             valuefunc = itemgetter(1)
-        else:
-            return groupby(iterable, keyfunc)
-    for key, group in groupby(iterable, keyfunc):
-        yield key, map(valuefunc, group)
+        for key, group in groupby(iterable, keyfunc):
+            yield key, map(valuefunc, group)
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -3,7 +3,7 @@ from __future__ import print_function
 from collections import Counter, defaultdict, deque
 from functools import partial, wraps
 from heapq import merge
-from itertools import chain, count, islice, repeat, takewhile, tee
+from itertools import chain, count, groupby, islice, repeat, takewhile, tee
 from operator import itemgetter
 from sys import version_info
 
@@ -13,6 +13,7 @@ from six.moves import filter, map, zip, zip_longest
 from .recipes import flatten, take
 
 __all__ = [
+    'adjacent',
     'always_iterable',
     'bucket',
     'chunked',
@@ -23,6 +24,7 @@ __all__ = [
     'distribute',
     'divide',
     'first',
+    'groupby_transform',
     'ilen',
     'interleave_longest',
     'interleave',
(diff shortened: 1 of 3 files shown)


bbayles:
Could you fix up the Python 2 error? I'll review more thoroughly when I get some time, but Travis is complaining of a legitimate issue.

diazona:
Seems OK now. Sorry about that, I always seem to forget to test in Python 2. (Admittedly this was a tricky one)

bbayles:
Thanks! I'll have some comments and change requests, as usual. :scissors:

more_itertools/more.py
```diff
@@ -1106,3 +1108,83 @@ def always_iterable(obj):
...
+    i1, i2 = tee(iterable)
+    padding = [False] * distance
+    selected = chain(padding, map(predicate, i1), padding)
+    adjacent = map(any, windowed(selected, 2 * distance + 1))
+    return zip(adjacent, i2)
+
+def groupby_transform(iterable, keyfunc=None, valuefunc=None):
+    """
+    A wrapper for ``itertools.groupby()`` which computes the values in the group
+    by applying *valuefunc* to the items in the iterable.
```
bbayles: Could you word this a little differently? I think it's more accurate to say that it groups the items by applying `keyfunc`, but applies `valuefunc` to them after.

bbayles: With that in mind, and as you guessed, I'd rather not have the `itemgetters` be the defaults here. I'm fine if they go in the example, but I think this function should be basically be "`groupby`, but you can transform the values after grouping."

bbayles: Also with that in mind, I'd simplify the implementation to this, perhaps?

```
def groupby_transform(iterable, keyfunc=None, valuefunc=None):
    valuefunc = (lambda x: x) if valuefunc is None else valuefunc
    for key, group in groupby(iterable, key=keyfunc):
        transformed_group = (valuefunc(item) for item in group)
        yield key, transformed_group
```

I think this is minimal.

more_itertools/more.py
```diff
@@ -1106,3 +1108,83 @@ def always_iterable(obj):
...
+
+        >>> grouper = groupby_transform(range(10),
+        ...                             lambda x: int(x / 5), lambda x: x + 2)
```
bbayles: Could you switch to `x // y` instead of `int(x / y)`?

more_itertools/more.py
```diff
@@ -1106,3 +1108,83 @@ def always_iterable(obj):
         return obj,
 
     return obj
+
+def adjacent(predicate, iterable, distance=1):
+    """
+    Returns an iterable over ``(bool, item)`` tuples where the ``item`` is drawn
```
bbayles: Nitpicks: I think Erik prefers no line break after `"""` for docstrings, and [PEP 257](https://www.python.org/dev/peps/pep-0257/#id16) mystifyingly wants things in the imperative mood ("Return..." and not "Returns...")

more_itertools/more.py
```diff
@@ -1106,3 +1108,83 @@ def always_iterable(obj):
...
+        >>> list(adjacent(lambda x: x % 4 == 2, range(6)))
+        [(False, 0), (True, 1), (True, 2), (True, 3), (False, 4), (False, 5)]
+
+    The *distance* used to determine what counts as adjacent can be configured:
+
+        >>> list(adjacent(lambda x: x % 4 == 2, range(6), 2))
+        [(True, 0), (True, 1), (True, 2), (True, 3), (True, 4), (False, 5)]
+
+    In this case any elements within 2 positions of the one that satisfies the
+    predicate are paired with ``True``. *distance* can be any nonnegative
```
bbayles: Could you break a paragraph after the "In this case..." sentence? Also, I'd probably leave out the example of this being the same as `map` and `zip`; if the reader wants to prove it to themselves we shouldn't deprive them of the exercise.

more_itertools/more.py
```diff
@@ -1106,3 +1108,83 @@ def always_iterable(obj):
...
+    value, see ``groupby_transform()``.
+    """
+    # Allow distance=0 mainly for testing that it reproduces results with map()
+    if distance < 0:
+        raise ValueError('distance must be at least 0')
+
+    i1, i2 = tee(iterable)
+    padding = [False] * distance
+    selected = chain(padding, map(predicate, i1), padding)
+    adjacent = map(any, windowed(selected, 2 * distance + 1))
```
bbayles: Could you not use the same variable name as the function here?

more_itertools/more.py
```diff
@@ -1106,3 +1108,83 @@ def always_iterable(obj):
...
+    ``adjacent()`` is designed to avoid calling *predicate* more than once per
+    item in the *iterable*, and thus is suitable for computationally expensive
+    *predicate*s.
+
+    To group consecutive elements which are associated with the same boolean
+    value, see ``groupby_transform()``.
+    """
+    # Allow distance=0 mainly for testing that it reproduces results with map()
+    if distance < 0:
+        raise ValueError('distance must be at least 0')
```
bbayles: I don't think this line is tested? I'm going to turn on coverage reports eventually.

more_itertools/more.py
diazona: Ah, sure. I forgot that `//` works in Python 2 too.

more_itertools/more.py
diazona: I suppose one _could_ do this:
```python
((k, map(valuefunc or lambda x: x, g) for k, g in groupby(iterable, keyfunc))
```
but that seems practically code-golfed to me ;-) It definitely does get simpler without special-casing the input arguments, though.

Do you specifically prefer generator syntax over `map()` for applying `valuefunc` to the group? Since we're using six, it's an iterable map either way, not the Python 2 `map()` which creates a list.

Since this is a pretty significant change to that one function, I'd consider doing the change on a separate branch, so the development history isn't cluttered with this version of `groupby_transform` that uses `itemgetter()`. Any preference on whether I do that, or just tack the change on to this branch?

diazona:
I'll get to work on those changes. I've got a busy weekend coming up so it may take a little while.

more_itertools/more.py
bbayles: I'm fine with just adding to this branch. And nice one-liner!

bbayles:
Many thanks for the alterations. I'll add these to the API docs and changelog before the next release.

PR #111, opened 2017-02-12
Add coverage tracking

bbayles:
This PR adds coverage with Coveralls.io . I've changed the Travis configuration file to use standard testing, since Tox doesn't seem to play nicely (probably something with directories).

Looks like we've got [100%](https://coveralls.io/builds/10112035); I'll add a badge once this is merged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ python:
     - "3.6"
 
 install:
-    - "pip install -U coveralls nose"
+    - "pip install -U coveralls nose six"
 
 script:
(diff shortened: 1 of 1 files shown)


bbayles:
(I turned off coveralls comment tracking after making the PR)

erikrose:
Nice!

I wonder if we can `pip install .` or something so we don't have to repeat the more-itertools dependencies (like six, atm) in the Travis config.

bbayles:
Yes, that worked - good call!

PR #112, opened 2017-02-13
With file

yardsale8:
In the spirit of ``with_iter`` and ``side_effect``, I have implemented ``with_file`` to handle context management of output files.  

For example, you can wrap a file handler in an iterator to log output to a file.

     from more_itertools import consume, with_file
     consume(print(x, file=output) for output in with_file(open("temp.tmp","w")) for x in range(3))

The pull request includes a ``nose`` test that passes under versions 2.7.12, 3.2.6, 3.3.6, 3.4.5, 3.5.2.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -40,6 +40,7 @@ __all__ = [
     'stagger',
     'unique_to_each',
     'windowed',
+    'with_file',
     'with_iter',
     'zip_offset',
 ]
@@ -380,6 +381,21 @@ def iterate(func, start):
         yield start
         start = func(start)
 
+
+def with_file(file_handle):
+    """Wrap a file context manager in a ``with`` statement, so it closes once consumed.
+
+    For example, this will close the file when the iterator is consumed::
+
+        >>> from more_itertools import consume
+        >>> consume(print(line, file=outfile)
+                    for outfile in with_file(open("temp.tmp", "w"))
+                    for line in range(3))
+
+
+    """
+    with file_handle as f:
+        yield f
 
 def with_iter(context_manager):
     """Wrap an iterable in a ``with`` statement, so it closes once exhausted.
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -387,10 +387,10 @@ def with_file(file_handle):
 
     For example, this will close the file when the iterator is consumed::
 
-        >>> from more_itertools import consume
-        >>> consume(print(line, file=outfile)
-                    for outfile in with_file(open("temp.tmp", "w"))
-                    for line in range(3))
+        from more_itertools import consume
+        consume(print(line, file=outfile)
+                for outfile in with_file(open("temp.tmp", "w"))
+                for line in range(3))
 
 
     """
--- a/more_itertools/tests/test_more.py
+++ b/more_itertools/tests/test_more.py
@@ -1,8 +1,8 @@
-from __future__ import division, unicode_literals
+from __future__ import division, unicode_literals, print_function
 
 from contextlib import closing
 from functools import reduce
-from io import StringIO
+from io import StringIO, BytesIO
 from itertools import chain, count, permutations
 from unittest import TestCase
 
@@ -11,7 +11,8 @@ import six
 from six.moves import filter, range
 
 from more_itertools import *  # Test all the symbols are in __all__.
-
+import sys
+PY3 = sys.version_info[0] == 3
 
 class CollateTests(TestCase):
     """Unit tests for ``collate()``"""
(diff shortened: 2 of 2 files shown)


bbayles:
Thanks for the PR. I'm a little puzzled, because I don't think I understand the use case - could you help me?

From the examples it seems like this aides in doing I/O operations in a comprehension or in an iterable-consuming expression. Is that the goal? Isn't this clearer?
```python
with open('temp.tmp', 'w') as output:
    [print(x, file=output) for x in range(3)]
```

It seems odd to save a line (not really, with the import) and introduce a strange "iterate over the file handle" concept.

Also, in the test, isn't [closing](https://github.com/erikrose/more-itertools/pull/112/files#diff-228d30171f84b762c55cb716df56ec37R373) doing the work here?

yardsale8:
I like the option of composing lazy expressions.  I really favor `with_iter` for this reason, as I can assure context management with an expression.  You are probably right about the unorthodox method of resolving the context management.

I came up with this generator while trying to use `side_effect` to log to a file.  This function isn't as much about saving a line as removing the need to use the `with` statement, which is more difficult to compose with other expressions.  I couldn't think of a way to properly handle the context management without writing another generator, like

    def with_log(handle, seq):
        with handle as log:
            print_log = lambda x: print(x, file=log)
            yield from side_effect(print_log, seq)  

which, in this case, defeats the purpose of the `side_effect` abstraction as I have mostly rewritten the logic.  Maybe I am missing something, is there a way to use `side_effect` to write to a log file without having to write another generator?  

Would adding `with_log` be a more welcome addition?

As for the test, I somewhat blindly modeled that test after `test_with_iter`, which also comes with `closing`.  I can rework the test if needed.

yardsale8:
Another solution might be to add an optional parameters `file=None` to `side_effect` to allow context management there.  I am imagining using `func = partial(func, **{'file':file})` inside a `with` statement if `file is not None`.  This way the user could design logging function that, for example, interact with the chunk size and writes to a file.

I will make a couple more pull requests based on these ideas, and you can decide if any of them seem suitable for `more_itertools`

Regardless, thanks for managing this module, I really appreciate having easy access to tools for explicit side-effects like `side_effect` and `consume`.

bbayles:
Thanks for the note. Would this accomplish what you want? Your idea about `side_effect` inspired this; it seems like a sensible generalization:

```python
def side_effect(
    func,
    iterable,
    chunk_size=None,
    func_args=None,
    func_kwargs=None,
    file_obj=None,
):
    """Invoke *func* on each item in *iterable* (or on each *chunk_size* group
    of items) before yielding the item.

    `func` must be a function that takes an item from the iterable as its first
    argument. Specify subsequent arguments by passing an
    *func_args* (an iterable) and *func_kwargs* (a mapping).
    Any returned values are discarded.

    If *file_obj* is given, it will be closed after iterating. This can be
    useful if the side effect is operating on

    `side_effect` can be used for logging, updating progress bars, or anything
    that is not functionally "pure."

    Emitting a status message:

        >>> from more_itertools import consume
        >>> func = lambda item: print('Received {}'.format(item))
        >>> consume(side_effect(func, range(2)))
        Received 0
        Received 1

    Operating on chunks of items:

        >>> pair_sums = []
        >>> func = lambda chunk: pair_sums.append(sum(chunk))
        >>> list(side_effect(func, [0, 1, 2, 3, 4, 5], 2))
        [0, 1, 2, 3, 4, 5]
        >>> list(pair_sums)
        [1, 5, 9]

    Writing to a file-like object:

        >>> from io import StringIO
        >>> from more_itertools import consume
        >>> f = StringIO()
        >>> consume(
        ...     side_effect(print, [1, 2], func_kwargs={'file': f}, file_obj=f)
        ... )
        >>> f.closed
        True

    """
    func_args = [] if func_args is None else func_args
    func_kwargs = {} if func_kwargs is None else func_kwargs
    try:
        if chunk_size is None:
            for item in iterable:
                func(item, *func_args, **func_kwargs)
                yield item
        else:
            for chunk in chunked(iterable, chunk_size):
                func(chunk, *func_args, **func_kwargs)
                for item in chunk:
                    yield item
    finally:
        if file_obj is not None:
            file_obj.close()
```

If that works, I'll make a PR for that and we can close this?

Your comment about the tests for `with_iter` are correct; I think those need some improvement.

yardsale8:
That is perfect, thanks.  Thanks for the quick solution and feel free to close this PR whenever.

yardsale8:
BTW, the following re-factorization would eliminate the replicated loop.
      
``` python
seq = iterable if chunk_size is None else chunked(iterable, chunk_size)
for item in seq:
    func(item, *func_args, **func_kwargs)
    yield item
```

Thanks again for the quick turn around!

PR #113, opened 2017-02-16
Add optional arguments for side_effect()

bbayles:
Re: PR #112, this adds some new optional keyword arguments to `side_effect()`.

~~Previously the function passed in was only allowed to accept a single argument. It now must accept an item from the iterable as the first argument, and a fixed set of other arguments can be passed in. Additionally, keyword arguments can be specified.~~

Also added is `file_obj`, which is a hint that the function will operate on a file object that should be closed after iterating. See the tests, docstring, and the older PR for motivating use cases.

---

~~I could be persuaded to get rid of the argument / keyword argument and just give examples with lambdas or partials. Anybody feel strongly?~~

Many thanks to @yardsale8 for the suggestion.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -743,12 +743,24 @@ def collapse(iterable, base_type=None, levels=None):
         yield x
 
 
-def side_effect(func, iterable, chunk_size=None):
+def side_effect(
+    func,
+    iterable,
+    chunk_size=None,
+    func_args=None,
+    func_kwargs=None,
+    file_obj=None,
+):
     """Invoke *func* on each item in *iterable* (or on each *chunk_size* group
     of items) before yielding the item.
 
-    `func` must be a function that takes a single argument.  Its return value
-    will be discarded.
+    `func` must be a function that takes an item from the iterable as its first
+    argument. Specify subsequent arguments by passing an
+    *func_args* (an iterable) and *func_kwargs* (a mapping).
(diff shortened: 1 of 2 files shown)


bbayles:
> I could be persuaded to get rid of the argument / keyword argument and just give examples with lambdas or partials. Anybody feel strongly?

I persuaded myself; this PR now just adds the `file_obj` hint.

more_itertools/more.py
```diff
@@ -770,16 +774,31 @@ def side_effect(func, iterable, chunk_size=None):
...
         [1, 5, 9]
 
+    Writing to a file-like object:
+
+        >>> from io import StringIO
+        >>> from more_itertools import consume
+        >>> f = StringIO()
+        >>> func = lambda x: print(x, file=f)
+        >>> it = [u'a', u'b', u'c']
+        >>> consume(side_effect(func, it, file_obj=f))
```
erikrose: I think we ought to change `file_obj` to `file` for brevity. That was going to be my only reason, but look at the `print` call just above: `file=`. So, consistency, too. :-) What do you think?

bbayles: `file` is the name of a built-in in Python 2, otherwise I would have done that.

erikrose: But as a kwarg, it doesn't collide with names: it's more akin to a dictionary key.

Issue #114, opened 2017-02-23
Consider making side_effect(file_obj=) arg more generic


erikrose:
Closing a file after yielding its last line is all well and good, but I wonder: can we be more general and add power without losing much brevity?

```python
side_effect(log, some_file, last=lambda: some_file.close())
```

```python
side_effect(ingest, listdir(some_dir), last=lambda: rmtree(some_dir))
```

@bbayles? @yardsale8?

bbayles:
Hmm, interesting.

(1) This makes me think that at the minimum we should change the documentation (and maybe the argument name) to be more generic - it works for anything that needs to be closed, not just file objects.

(2) it does seem like files are a justifiably special case with `side_effect`, since writing to a file is the quintessential "impure" activity.

(3) Adding a generic "do this thing after you're done `side_effecting` the iterable items" opens the door to "do this thing _before_ you start `side_effecting` the iterable items," which seems like it might be a bridge too far? The goal here is to make it easy to do something common (close a file) that's awkward to do currently (you'd have to wrap your whole pipeline in a `with` block), not to let you put more things on a line (this was my initial resistance to #112 ). 

I'll look at the language for (1) and noodle around with (3).

bbayles:
I thought the "well if you want to run a generic function after, why not before" was a reductio ad absurdum, but now that I've implemented it I kind of like it? See #115.

erikrose:
I came to the same conclusion moments after my head hit the pillow: why not do `before` also? It took another 30 seconds from there to "Why not just have it take a context manager?" From there, I landed back at `with_iter`. In fact, I still am failing to see why #112 couldn't have just been written as…

```python
consume(print(x, file=output) for output in with_iter(open("temp.tmp","w")) for x in range(3)).
```

yardsale8:
I am with Eric on the fact that we are basically reinventing context management.  Eric's suggestion was one of the first I tried, but it fails.

``` python
In [1]:  [print(x, file=out) for out in with_iter(open('temp.tmp', 'w')) for x in range(3)]
(snip)
    342     with context_manager as iterable:
--> 343         for item in iterable:
    344             yield item
    345

UnsupportedOperation: not readable
```

bbayles:
Right, `with_iter()` would be fine for reading from a file that's the source for our iterable, but in Erik's example `range()` is the source.

`contextlib.closing()` almost does what we want, but not quite.

[...]

PR #115, opened 2017-02-23
Add before and after functions to side_effect

bbayles:
Re: Issue #114, this PR takes the suggestion to remove the `file_obj` argument from `side_effect()` and replace it with something more general/powerful.

Instead of `file_obj`, we now have optional `before` and `after` arguments. These are supposed to be functions that don't take any arguments that will operate pre- and post-iteration.

I thought this was crazy at first, but now I'm kind of fond of the idea.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -746,15 +746,15 @@ def collapse(iterable, base_type=None, levels=None):
         yield x
 
 
-def side_effect(func, iterable, chunk_size=None, file_obj=None):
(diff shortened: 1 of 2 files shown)


erikrose:
This is fine with me, as far as it goes, though I'd still be interested in your thoughts on https://github.com/erikrose/more-itertools/issues/114#issuecomment-282402264.

erikrose:
Recommend leaving this unmerged for now. Please see rationale in https://github.com/erikrose/more-itertools/issues/114#issuecomment-287829005.

Issue #116, opened 2017-03-01
ilen() is slow


wbolster:
you may want to steal the `cardinality.count()` implementation from
https://github.com/wbolster/cardinality  ;)

bbayles:
Ha, thanks - I know that `consume` uses a similar trick. Switching to the length-1 deque saves ~1% on my machine. Will go in with PR #117.

wbolster:
speed increase depends on what you are iterating over. :)

and for things that have `__len__()` this brings it from O(n) to O(1) which can be a massive speedup.

bbayles:
Fixed in PR #117; many thanks.

PR #117, opened 2017-03-02
Use a length-1 deque for ilen()

bbayles:
Re: Issue #116, this PR borrows @wbolster's trick of feeding an `enumerate()`-wrapped iterable into a length-1 deque to get the length. This buys a slight speed increase.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -368,6 +368,24 @@ def ilen(iterable):
     This does, of course, consume the iterable, so handle it with care.
 
     """
+    try:
+        return len(iterable)
+    except TypeError:
+        pass
+
+    d = deque(enumerate(iterable, 1), maxlen=1)
+    return d[0][0] if d else 0
+
+
+def ilen_2(iterable):
+    """Return the number of items in ``iterable``.
+
+        >>> ilen(x for x in range(1000000) if x % 3 == 0)
+        333334
+
+    This does, of course, consume the iterable, so handle it with care.
+
+    """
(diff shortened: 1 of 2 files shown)


wbolster:
this should have a `__len__()` check as i suggested in #116

wbolster:
...or maybe not depending on the intended semantics (forcibly consume or not).

bbayles:
Trying `len()` first makes sense; I've added that.

wbolster:
as i hinted in https://github.com/erikrose/more-itertools/pull/117#issuecomment-283613962 i am not sure what the expected semantic are so this may be a breaking change...

bbayles:
Hmm, yes. I don't think one would generally think of using `ilen()` as a version of `consume()`, but now that I think about it, I could imagine using it.

@erikrose , thoughts about making that explicit?
```python
def ilen(iterable, n=None):
    """Return the number of items produced by advancing the iterator *n*
    steps ahead. 

    If *n* is ``None`` the iterable will be entirely consumed::

        >>> ilen((x for x in range(10)))
        10
        >>> ilen((x for x in range(10) if x % 3 == 0))
        4

    If the iterable has fewer than *n* items, the length of the iterable will
    be returned::

        >>> iterable = (x for x in 'abcd')
        >>> ilen(iterable, 10)
        4

    """
```

bbayles:
OK, maybe the `n` is a bridge too far. I've zapped the `__len__` stuff again and am left with the original optimization. I'll squash the commits upon merge.

PR #118, opened 2017-03-21
Add context() helper

bbayles:
This is #112 with a new name, docstring, and test. It also removes the `file_obj` argument to `side_effect()`. See #114 for the discussion.

Many thanks to @yardsale8 for the idea that sparked all this, and @erikrose for making the call.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -780,7 +780,7 @@ def side_effect(func, iterable, chunk_size=None):
         >>> from io import StringIO
         >>> from more_itertools import consume
         >>> with StringIO() as f:
-        ...     func = lambda x: print(x, end=',', file=f)
+        ...     func = lambda x: print(x, end=u',', file=f)
         ...     it = [u'a', u'b', u'c']
         ...     consume(side_effect(func, it))
         ...     print(f.getvalue())
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -747,15 +747,12 @@ def collapse(iterable, base_type=None, levels=None):
         yield x
 
 
-def side_effect(func, iterable, chunk_size=None, file_obj=None):
+def side_effect(func, iterable, chunk_size=None):
     """Invoke *func* on each item in *iterable* (or on each *chunk_size* group
     of items) before yielding the item.
 
     `func` must be a function that takes a single argument. Its return value
     will be discarded.
-
-    If *file_obj* is given, it will be closed after iterating. This can be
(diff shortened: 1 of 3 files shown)


more_itertools/more.py
```diff
@@ -1200,3 +1194,26 @@ def groupby_transform(iterable, keyfunc=None, valuefunc=None):
...
+
+
+def context(obj):
+    """Wrap *obj*, an object that supports the context manager protocol,
+    in a ``with`` statement and then yield the resultant object.
+
+    The object's ``__enter__()`` method runs before this function yields, and
+    its ``__exit__()`` method runs after control returns to this function.
+
+    This can be used to operate on objects that can close automatically when
```
erikrose: We should warn people to put the ctx mgr in the first "in" clause lest it wobble open and closed multiple times.

more_itertools/tests/test_more.py
```diff
@@ -1109,3 +1095,21 @@ def key(x):
...
+
+
+class ContextTestCase(TestCase):
+    def test_basic(self):
+        before = []
+        after = []
+
+        @contextmanager
+        def managed():
+            before.append(1)
```
erikrose: This looks correct to me, but it would be more resistant to mistakes if we chose a different value (easiest to read if it were a meaningful string) for each behavior.

more_itertools/more.py
```diff
@@ -1214,6 +1214,19 @@ def context(obj):
         >>> file_obj.closed
         True
 
+    Be sure to iterate over the returned context manager in the outermost
```
erikrose: Can we add "so that it only enters and exits once"?

more_itertools/tests/test_more.py
```diff
@@ -1105,11 +1105,11 @@ def test_basic(self):
         @contextmanager
         def managed():
             before.append(1)
-            yield 1
```
erikrose: I meant unique across all the values—1,2,3 vs. range(3)'s 0, 1, 2—but don't waste your time messing with it more.

erikrose:
Otherwise, merge merge merge!

more_itertools/tests/test_more.py
bbayles: I think I caught your meaning and fixed in ce02c10cb8483133513ccea588f2661b7d337d47

erikrose:
So you did. Maybe I was looking at the wrong thing; I was reviewing from my phone.

Issue #126, opened 2017-04-21
range for floats


BebeSparkelSparkel:
I didn't see a float range function in the library so frange could be helpful. I also used the recommendation from itertools.count to reduce float error.

> When counting with floating point numbers, better accuracy can sometimes be achieved by substituting multiplicative code such as: `(start + step * i for i in count())`.

```python
import itertools, operator

# frange(stop)
# frange(start, stop[, step])
def frange(*args):
  if len(args) == 1:
    start = 0
    stop = args[0]
    step = 1
  elif len(args) == 2:
    start, stop = args
    step = 1
  elif len(args) == 3:
    start, stop, step = args
  else:
    raise TypeError('frange expected at most 1 - 3 arguments, got {}.'.format(len(args)))

  if start < stop and 0 < step:
    compare_with = operator.lt
  elif start > stop and 0 > step:
    compare_with = operator.gt
  else:
    return

  compare_with = operator.lt if start < stop else operator.gt

  for step_count in itertools.count():
    val = start + step * step_count
    if compare_with(val, stop):
      yield val
    else:
      break


import unittest

class Testfrange(unittest.TestCase):
  def test_frange(self):
    self.assertEqual(
        tuple(frange(5)),
        tuple( range(5))
      )

    self.assertEqual(
        tuple(frange(-5)),
        tuple( range(-5))
      )

    self.assertEqual(
        tuple(frange(2, 5)),
        tuple( range(2, 5))
      )

    self.assertEqual(
        tuple(frange(2, 10, 2)),
        tuple( range(2, 10, 2))
      )

    self.assertEqual(
        tuple(frange(2, -5)),
        tuple( range(2, -5))
      )

    self.assertEqual(
        tuple(frange(2, -10, -2)),
        tuple( range(2, -10, -2))
      )

    self.assertEqual(
        tuple(frange(2, 10, -2)),
        tuple( range(2, 10, -2))
      )

    self.assertEqual(
        tuple(frange(2, -10, 2)),
        tuple( range(2, -10, 2))
      )

    self.assertEqual(
        tuple(frange(2.5, 4, 0.5)),
        (2.5, 3, 3.5)
      )

    self.assertEqual(
        tuple(frange(2.5, 4.1, 0.5)),
        (2.5, 3, 3.5, 4)
      )

    with self.assertRaises(TypeError):
      tuple(frange())

    with self.assertRaises(TypeError):
      tuple(frange(1, 2, 3, 4))


if __name__ == '__main__':
[...]

PR #127, opened 2017-04-22
Add arithmetic_sequence()

bbayles:
Re: Issue #126, this PR adds a new itertools, `arithmetic_sequence()`.

This function works for `float`s (as requested) as well as `Decimal` and `Fraction` types.

I'm happy to include this because we take care not to accumulate round-off errors, as many Stack Overflow implementations of this do.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -10,7 +10,6 @@ New Routines
 
 .. autofunction:: adjacent
 .. autofunction:: always_iterable
-.. autofunction:: arithmetic_sequence(start, stop, step)
 .. autofunction:: bucket
 .. autofunction:: chunked
 .. autofunction:: collapse
@@ -26,6 +25,7 @@ New Routines
 .. autofunction:: interleave_longest
 .. autofunction:: intersperse
 .. autofunction:: iterate
+.. autofunction:: numeric_range(start, stop, step)
 .. autofunction:: one
 .. autofunction:: padded
 .. autoclass:: peekable
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -10,6 +10,7 @@ New Routines
 
 .. autofunction:: adjacent
 .. autofunction:: always_iterable
+.. autofunction:: arithmetic_sequence(start, stop, step)
 .. autofunction:: bucket
 .. autofunction:: chunked
 .. autofunction:: collapse
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -15,7 +15,6 @@ from .recipes import flatten, take
 __all__ = [
     'adjacent',
     'always_iterable',
-    'arithmetic_sequence',
     'bucket',
     'chunked',
     'collapse',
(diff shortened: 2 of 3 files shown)


docs/api.rst
```diff
@@ -10,6 +10,7 @@ New Routines
 
 .. autofunction:: adjacent
 .. autofunction:: always_iterable
+.. autofunction:: arithmetic_sequence(start, stop, step)
```
bbayles: If someone knows how to convince Sphinx to do a second signature without duplicating the whole function, please let me know; I couldn't figure it out.

more_itertools/more.py
```diff
@@ -1214,3 +1215,65 @@ def groupby_transform(iterable, keyfunc=None, valuefunc=None):
     """
     valuefunc = (lambda x: x) if valuefunc is None else valuefunc
     return ((k, map(valuefunc, g)) for k, g in groupby(iterable, keyfunc))
+
+
+def arithmetic_sequence(*args):
+    """An extension of the built-in ``range()`` function whose arguments can
+    be any orderable numeric type.
```
bbayles: Note any *orderable* numeric type; this won't work for `complex`.

more_itertools/more.py
```diff
@@ -1214,3 +1215,65 @@ def groupby_transform(iterable, keyfunc=None, valuefunc=None):
     """
     valuefunc = (lambda x: x) if valuefunc is None else valuefunc
     return ((k, map(valuefunc, g)) for k, g in groupby(iterable, keyfunc))
+
+
+def arithmetic_sequence(*args):
```
erikrose: Why not use the word "range" in the name, both for brevity and to suggest similarity with the builtin of the same name? math_range, arithmetic_range, numeric_range

erikrose:
Otherwise, lgtm!

more_itertools/more.py
bbayles: `numeric_range()` sounds better to me.

BebeSparkelSparkel:
I think that the docs could have a better description of the input requirements like the range class has because it is unclear what input combinations are possible by looking at `more_itertools.numeric_range(start, stop, step)`.

```python
class range(stop)
class range(start, stop[, step])
```

bbayles:
If you have an idea on how to do this with `autofunction`, [please let me know](https://github.com/erikrose/more-itertools/pull/127#discussion_r112801792).

BebeSparkelSparkel:
I didn't see anything with auto function but there appears to be a way with function
[http://www.sphinx-doc.org/en/stable/domains.html#python-signatures](http://www.sphinx-doc.org/en/stable/domains.html#python-signatures)

BebeSparkelSparkel:
Also, I think it is a good idea to list the types that are acceptable in the description and not just in the examples.

bbayles:
I'll take a PR for the signature if anyone knows how to do it and still have the docstring work.

Issue #130, opened 2017-05-08
Installation and examples in README


pylang:
I suspect the README is the first (and often the only) stop from visitors to most packages.  Through the visitor's lens, I observe the following:

First, it's actually not clear how to install the package (from either the README or the docs).  I propose **adding a line on installation to the README**.

...
## Installation
    > pip install more_itertools
...

Second,  right now the first tool in the API docs is the new `more_itertools.adjacent`, which has a heavier docstring than traditional ones.  IMO, seeing this much text for the first tool is less inviting to newcomers trying to quickly figure out what this package does.  I suggest **adding a couple elegant examples to the README**, maybe a recipe and original, e.g. `flatten` and `chunked`.  

...
## Examples

```python
>>> import more_itertools as mit

# Itertools Recipe
>>> list(mit.flatten([[0, 1], [2, 3]]))
[0, 1, 2, 3]

# More-Itertools Original
>>> list(mit.chunked([1, 2, 3, 4, 5, 6, 7], 3))
[[1, 2, 3], [4, 5, 6], [7]]

```
...

The idea is to succinctly demonstrate up front that this package is a simple extension of `itertools`, and it is easy to use.

erikrose:
These sound like sane suggestions to me. In particular, it's probably time for us to divide our routines into categories to make the docs more approachable; there are rather a lot of them now!

bbayles:
Good suggestion - I've thought the same about the ordering of the docs recently. I'll take a run at this shortly.

bbayles:
- [ ] Update the README with installation instructions and examples (PR #131)
- [ ] Similarly update `docs/index.rst`
- [ ] Look at categorizing the functions

bbayles:
Here is my proposed grouping and ordering for the docs, with easy-to-grok functions listed first:

---
Grouping - routines that yield groups of items from an iterable:
* chunked
* sliced
* distribute
* divide
* split_before
* split_after
* bucket

Lookahead - routines that peek at an iterable's values:
* spy
* peekable

Windowing - routines that yield windowed data
* windowed
* stagger

Augment - routines that augment an itertable's items with additional data:
* adjacent
* count_cycle
* groupby_transform
* intersperse
* padded

Combine multiple - functions that combine multiple iterables:
* collapse
* sort_together
* interleave
* interleave_longest
* collate
* zip_offset

Summarization - functions that return data derived from an iterable:
* ilen
* first
* one
* unique_to_each

Others:
* numeric_range
[...]

PR #131, opened 2017-05-08
Update README with examples

bbayles:
Re: Issue #130, this PR updates the README here on GitHub to provide some installation instructions and motivating examples.

We can probably bikeshed this forever, but I think it's an improvement? I'll do the top-level Read The Docs page separately.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -56,4 +56,4 @@ Development
 ``more-itertools`` is maintained by `@erikrose <https://github.com/erikrose>`_
 and `@bbayles <https://github.com/bbayles>`_, with help from `many others <https://github.com/erikrose/more-itertools/graphs/contributors>`_.
 If you have a problem or suggestion, please file a bug or pull request in this
-repository - thanks for contributing!
+repository. Thanks for contributing!
--- a/README.rst
+++ b/README.rst
@@ -1,3 +1,7 @@
+==============
+More Itertools
+==============
+
 .. image:: https://coveralls.io/repos/github/erikrose/more-itertools/badge.svg?branch=master
   :target: https://coveralls.io/github/erikrose/more-itertools?branch=master
 
--- a/README.rst
+++ b/README.rst
@@ -1,23 +1,53 @@
-==============
 More Itertools
 ==============
 
 .. image:: https://coveralls.io/repos/github/erikrose/more-itertools/badge.svg?branch=master
   :target: https://coveralls.io/github/erikrose/more-itertools?branch=master
 
-I love itertools; it's one of the most beautiful, composable standard libs.
-Whenever I have an iteration problem, there's almost always an itertools
-routine that fits it perfectly. Sometimes, however, neither itertools nor the
-recipes included in its docs do quite what I need.
+Python's ``itertools`` library is a gem - you can compose elegant solutions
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -1,23 +1,53 @@
...
+    >>> list(flatten(iterable))
+    [0, 1, 2, 3]
+
+Several new recipes are available as well:
+
+.. code-block:: python
+
+    >>> from more_itertools import spy, chunked
+    >>> iterable = [0, 1, 2, 3, 4, 5, 6, 7]
+    >>> head, iterable = spy(iterable, n=2)
```
pylang: I'd show `chunked` first, then `spy`.  Somehow it is very clear what is going on with the `chunked` example without thinking.  This seems like an easier flow from `flatten`.

README.rst
```diff
@@ -1,23 +1,53 @@
-==============
 More Itertools
```
pylang: I don't think this is needed.

README.rst
```diff
@@ -1,23 +1,53 @@
...
+Getting started
+===============
+
+To get started, install the library with `pip <https://pip.pypa.io/en/stable/>`_:
+
+.. code-block:: shell
+
+    pip install more-itertools
+
+The recipes from the `itertools docs <https://docs.python.org/3/library/itertools.html>`_
```
pylang: You may instead want the permalink to the recipes section https://docs.python.org/3/library/itertools.html#itertools-recipes

README.rst
```diff
@@ -1,23 +1,53 @@
...
+Python iterables.
+
+Getting started
+===============
+
+To get started, install the library with `pip <https://pip.pypa.io/en/stable/>`_:
+
+.. code-block:: shell
+
+    pip install more-itertools
```
pylang: If I recall, either dash or underscore (`more-itertools` or `more_itertools`) should work from pip, but only the latter works for imports in python.  For consistency, would it be better to use the latter throughout?

bbayles: Eh. I think you'll need some more help than we can provide if you're attempting to import something with a `-` in the name.

README.rst
```diff
@@ -5,19 +5,55 @@ More Itertools
...
+
+For the full listing of functions, see the API documentation at `more-itertools.readthedocs.io <https://more-itertools.readthedocs.io/en/latest/api.html>`_.
+
+Development
+===========
+
+``more-itertools`` is maintained by `@erikrose <https://github.com/erikrose>`_
+and `@bbayles <https://github.com/bbayles>`_, with help from `many others <https://github.com/erikrose/more-itertools/graphs/contributors>`_.
+If you have a problem or suggestion, please file a bug or pull request in this
+repository - thanks for contributing!
```
erikrose: " - thanks" → ". Thanks"

erikrose:
Looks like an improvement to me! It'd be lovely if we could rig a way to not have to copy and paste this onto the front page of the Sphinx docs. Perhaps we can use the [Sphinx include directive](http://reinout.vanrees.org/weblog/2010/12/08/include-external-in-sphinx.html).

bbayles:
Cool - I'll work on something for the docs index also.

bbayles:
Bingo, that worked. Thanks!

PR #134, opened 2017-05-13
Use six.moves.range in more module

MSeifert04:
Several functions in `more.py` use `range` so ensuring it's `xrange` on python2 might speed them up and reduce the memory overhead.


bbayles:
Hmm, surprised I didn't do this originally. Thanks!

PR #135, opened 2017-05-13
use operators instead of lambdas when possible

MSeifert04:
The functions from the operator module are generally a bit faster than custom lambdas. 

In case of `numeric_range` this speed up the code by a (admittedly rather insignificant) 10%:

```
%timeit list(numeric_range(0., 200000, 1))  # old 
405 ms ± 8.16 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
%timeit list(numeric_range(0., 200000, 1))  # new
356 ms ± 5.19 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

and the overhead of creating a `partial` object compared to a `lambda` function is even for very short ranges almost negligable:

```
%timeit list(numeric_range(0., 2, 1))  # old
19.3 µs ± 119 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
%timeit list(numeric_range(0., 2, 1))  # new
20.6 µs ± 203 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -4,7 +4,7 @@ from collections import Counter, defaultdict, deque
 from functools import partial, wraps
 from heapq import merge
 from itertools import chain, count, groupby, islice, repeat, takewhile, tee
-from operator import itemgetter
+from operator import itemgetter, lt, gt
 from sys import version_info
 
 from six import binary_type, string_types, text_type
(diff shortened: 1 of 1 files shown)


bbayles:
Many thanks for running the performance tests.

Issue #136, opened 2017-05-13
numeric_range with one argument


MSeifert04:
Is there any use-case for `numeric_range` with only one argument? It seems the type of the objects that are returned are solely depending on the `type` of `start` and `step`. 

When only `stop` is given (even if I use `float`s or `Decimal` as `stop`) it will always return integers. And it's a lot slower than `range`.

bbayles:
The one-argument form is there to match the way `range` works.

Does PR #137 resolve this? I agree if you pass in a `float` you should probably get `floats` back.

MSeifert04:
Yes #137 would resolve the question about the use-case. It makes it a bit unpredictable though (see comment on the PR).

I actually just wondered why there's a 1-argument form at all, wouldn't `numeric_range(start, stop, step=1)` be better - it makes the function a lot easier and the result type is always defined by the `start` argument (easier documentation)?

bbayles:
Re: 1-argument version existence, again, wanted to match `range`. I'd rather not break the API and do a major version release to change it.

MSeifert04:
Oh, I didn't realize you already released `numeric_range`.

bbayles:
Oh, maybe I haven't yet... Either way, I want to match `range()`.

MSeifert04:
Yes, with #137 there's also a convincing use-case for it 👍

bbayles:
Thanks for the careful examination of this; hope to see more!

PR #137, opened 2017-05-13
Match output types for numeric_range(stop)

bbayles:
Re: Issue #136, this PR makes the one-argument version of `numeric_range()` output items that match the type of the input.


MSeifert04:
I'm not sure if that follows the "least astonishment" principle. Mainly because the returned types depend on the `stop` argument if and only if only the `stop` argument is given:

```
>>> list(numeric_range(3.59)
[0.0, 1.0, 2.0, 3.0]
>>> list(numeric_range(0, 3.59)
[0, 1, 2, 3]
```

Not a problem per-se but it should be mentioned in the docs. :)

bbayles:
Ah, yes. I put a note in about the type matching the first argument passed in.

more_itertools/more.py
```diff
@@ -1248,19 +1245,21 @@ def numeric_range(*args):
         >>> list(numeric_range(start, stop, step))
         [Fraction(1, 2), Fraction(1, 1), Fraction(3, 2), Fraction(2, 1)]
 
-    Negative steps are supported:
+    If *step* is zero,``ValueError`` is rasied. Negative steps are supported:
 
         >>> list(numeric_range(3, -1, -1.0))
         [3.0, 2.0, 1.0, 0.0]
 
+    The yielded items will have the same type as the first argument.
```
MSeifert04: except when the addition of `start` and a multiple of `step` "coerces" to another type. :)

bbayles: Good counterexample!

bbayles:
We're getting there! Thanks for the attention to detail.

more_itertools/more.py
```diff
@@ -1222,24 +1222,23 @@ def numeric_range(*args):
     """An extension of the built-in ``range()`` function whose arguments can
     be any orderable numeric type.
 
-    If the *start* argument is omitted, it defaults to ``0``. If the *step*
-    argument is omitted, it defaults to ``1``. If *step* is zero,
-    ``ValueError`` is rasied.
-
-    With only *stop* specified:
+    With only *stop* is specified, *start* defaults to ``0`` and *step*
```
MSeifert04: `If` instead of `With`

more_itertools/more.py
```diff
@@ -1248,7 +1247,7 @@ def numeric_range(*args):
         >>> list(numeric_range(start, stop, step))
         [Fraction(1, 2), Fraction(1, 1), Fraction(3, 2), Fraction(2, 1)]
 
-    Negative steps are supported:
+    If *step* is zero, ``ValueError`` is rasied. Negative steps are supported:
```
MSeifert04: ra**is**ed.

MSeifert04: And the second sentence could start with `But negative ...` to emphasis the relation to zero-steps.

more_itertools/more.py
bbayles: Eh.

more_itertools/more.py
bbayles: Good catch on `raise`d.

more_itertools/more.py
MSeifert04: Or `When`?

I'm not a native speaker but `With only stop is specified ...` sounds wrong.

bbayles: The _is_ is unnecessary. This PR is teaching me I should just listen to you the first time.

Issue #139, opened 2017-05-18
nth item


pylang:
I came across [this recipe](http://stackoverflow.com/a/8337489/4531270) by R. Hettinger:

```python
>>> x = [False,True,True,False,True,False,True,False,False,False,True,False,True]
>>> nth_item(50, True, x)
-1
>>> nth_item(0, True, x)
1
>>> nth_item(1, True, x)
2
>>> nth_item(2, True, x)
4
>>> nth_item(3, True, x)
6
```

Code

```python
>>> from itertools import compress, count, imap, islice
>>> from functools import partial
>>> from operator import eq

>>> def nth_item(n, item, iterable):
        indicies = compress(count(), imap(partial(eq, item), iterable))
        return next(islice(indicies, n, None), -1)
```

I thought this may be a useful addition.

bbayles:
I was a bit confused by the name. We have [nth](https://more-itertools.readthedocs.io/en/latest/api.html#more_itertools.nth), which does return the nth item. However, this is "what is the index of the nth occurrence of a particular item?"

So with a better name I'm open to this. `index_of_nth_occurrence` isn't very snappy?

pylang:
I agree.  `nth_index` makes more sense.

pylang:
`locate(n, item, iterable)` , as in locate the `n`th `item` in this `iterable`?  Or `where`?

Numpy has a [`where`][2] method I believe.  They also have [`argwhere`][3] which more similar to this application.  Not sure if it is general enough to call `index`... I'm not attached to any of these names.  

   [2]: https://docs.scipy.org/doc/numpy-1.12.0/reference/generated/numpy.where.html
   [3]: https://docs.scipy.org/doc/numpy/reference/generated/numpy.argwhere.html

bbayles:
I've asked Raymond Hettinger on Twitter if he would bless the inclusion of this.

The implementation is dynamite, but I think we'll want to get the name and docstring right. Making "nth occurrence" be interpreted as 0-based is confusing in a way that "nth item" is not. A co-worker of mine suggested "find the index of the item, skipping the first n matches."

I might also suggest changing this to: `locate(iterable, pred, n=0)` to be more flexible. That is, "find the first index where `pred` is `True`, skipping the first `n` matches. 

Thinking more generally, I wouldn't mind writing some more itertools that deal with locating indexes. Consider:
[...]

PR #140, opened 2017-05-20
Add locate()

bbayles:
Re: issue #139, this PR adds ~three~ one new tool~s~:
* `locate(iterable, pred, n)` - Returns the ~first index~ indexes where `pred` returns `True`, ~skipping the first `n` matches.~
* ~`item_index(iterable, item, start, end)` - Like `list.index()`, but for iterables. Returns the first index of `item` in `iterable`, restricting the search between `start` and `end`.~
* ~`sub_index(iterable, sub, start, end)` - List `str.index()`, but for iterables. Returns the first index where sub-sequence `sub` starts, restricting the search between `start` and `end`.~

`locate()` is a generalization of what was requested in the issue; the docstring has an example of how to look for a particular item.

~`item_index()` could be rolled into `sub_index()`, but it would be slower.~

I like these because they fit the of "easy to specify, seem like itertools should be applicable, tricky to get right" criteria we aim for.

CC: @pylang

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -113,7 +113,7 @@ These tools combine multiple iterables.
 Summarizing
 ===========
 
-These tools return summarized or aggregate data from an iterable.
+These tools return summarized or aggregated data from an iterable.
 
 ----
 
@@ -123,6 +123,9 @@ These tools return summarized or aggregate data from an iterable.
 .. autofunction:: first(iterable[, default])
 .. autofunction:: one
 .. autofunction:: unique_to_each
+.. autofunction:: item_index
+.. autofunction:: sub_index
+.. autofunction:: locate
 
 ----
 
@@ -200,7 +203,6 @@ Others
 .. autofunction:: numeric_range(start, stop, step)
 .. autofunction:: side_effect
 .. autofunction:: iterate
-.. autofunction:: sub_index
 
 ----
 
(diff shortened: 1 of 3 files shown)


MSeifert04:
I did some timings and a naive approach:

```
def item_index2(iterable, item, start=0, end=None):
    iterable = islice(iterable, start, end)
    for idx, ititem in enumerate(iterable, start):
        if ititem == item:
            return idx
    else:
        raise ValueError('item "{!r}" not found.'.format(item))
```

seems to be faster.

```
iterable = ['a']*10000 + ['b']
item = 'b'

%timeit item_index2(iterable, item)
# 2.6 ms ± 112 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
%timeit item_index(iterable, item)
# 6.96 ms ± 162 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

iterable = ['a']*5 + ['b']
item = 'b'

%timeit item_index2(iterable, item)
# 4.12 µs ± 361 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
%timeit item_index(iterable, item)
# 10.4 µs ± 358 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

This is probably also the case for the `subindex` approach.

bbayles:
Thanks for running the test. I'll leave the `item_index()` and `sub_index()` out of this PR for now.

I've changed `locate()` to return an iterator over matching indexes, which seems like the right place to end up.

PR #144, opened 2017-05-24
Add a quick-and-dirty regex to remove Sphinx calls that the sdist builder doesn't understand. Fix #143.

### LINKED ISSUE #143: ReST on PyPI page not rendering
https://pypi.python.org/pypi/more-itertools/ is showing the raw restructured text source code, not rendered HTML. Maybe a syntax error somewhere?

erikrose:
This unbreaks the readme's RST for PyPI.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,6 +4,7 @@ try:
     import multiprocessing
 except ImportError:
     pass
+from re import sub
 
 from setuptools import setup, find_packages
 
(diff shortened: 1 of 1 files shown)


bbayles:
Excellent - thanks for the solution.

bbayles:
Excellent - thanks for the solution.

erikrose:
Np. I've had to do that a time or two elsewhere. There's probably a fancy way to actually fire off Sphinx, but the simple hacky solution has never failed me yet. :-)

PR #145, opened 2017-05-28
lstrip, rstrip, and strip functions

bbayles:
I needed to do something like `str.rstrip` to an iterable the other day, and thought the implementation might be worthwhile to include here.

I'm not married to this - Watchers, what do you think? I'll write tests if we think it's cromulent.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -14,7 +14,7 @@ from itertools import (
     takewhile,
     tee
 )
-from operator import contains, itemgetter, lt, gt
+from operator import itemgetter, lt, gt
 from sys import version_info
 
 from six import binary_type, string_types, text_type
@@ -55,6 +55,7 @@ __all__ = [
     'split_before',
     'spy',
     'stagger',
+    'strip',
     'unique_to_each',
     'windowed',
     'with_iter',
@@ -1339,7 +1340,12 @@ def lstrip(iterable, skip_items):
     contains will be stripped.
 
     """
-    return dropwhile(partial(contains, set(skip_items)), iterable)
+    try:
+        skip_items = set(skip_items)
+    except TypeError:
+        skip_items = list(skip_items)
+
+    return dropwhile(skip_items.__contains__, iterable)
 
 
 def rstrip(iterable, skip_items):
@@ -1355,13 +1361,34 @@ def rstrip(iterable, skip_items):
     contains will be stripped.
 
     """
-    skip_items = set(skip_items)
+    try:
+        skip_items = set(skip_items)
+    except TypeError:
+        skip_items = list(skip_items)
+
     cache = []
+    cache_append = cache.append
     for x in iterable:
         if x in skip_items:
-            cache.append(x)
+            cache_append(x)
         else:
             for y in cache:
                 yield y
-            cache = []
+            del cache[:]
             yield x
+
+
+def strip(iterable, skip_items):
+    """Yield the items from *iterable*, but strip any that match *skip_items*
+    from the beginning and end.
+
+        >>> iterable = (None, False, None, 1, 2, None, 3, False, None)
+        >>> list(strip(iterable, (None, False, '')))
+        [1, 2, None, 3]
+
+    This function is analagous to :func:`str.strip`.
+    Note that the *skip_items* argument is not a prefix or suffix; any items
+    that it contains will be stripped.
+
+    """
+    return rstrip(lstrip(iterable, skip_items), skip_items)
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -1336,49 +1336,41 @@ def locate(iterable, pred=bool):
     return compress(count(), map(pred, iterable))
 
 
-def lstrip(iterable, skip_items):
-    """Yield the items from *iterable*, but strip any that match *skip_items*
-    from the beginning.
+def lstrip(iterable, pred):
+    """Yield the items from *iterable*, but strip any from the beginning
+    for which *pred* returns ``True``.
+
+    For example, to remove a set of items from the start of an iterable:
 
         >>> iterable = (None, False, None, 1, 2, None, 3, False, None)
-        >>> list(lstrip(iterable, (None, False, '')))
+        >>> pred = lambda x: x in {None, False, ''}
+        >>> list(lstrip(iterable, pred))
         [1, 2, None, 3, False, None]
 
-    This function is analagous to :func:`str.lstrip`.
-    Note that the *skip_items* argument is not a prefix; any items that it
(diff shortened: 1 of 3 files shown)


bbayles:
Oops, didn't mean to hit the review request button.

more_itertools/more.py
```diff
@@ -1321,3 +1324,44 @@ def locate(iterable, pred=bool):
...
+        >>> iterable = (None, False, None, 1, 2, None, 3, False, None)
+        >>> list(lstrip(iterable, (None, False, '')))
+        [1, 2, None, 3, False, None]
+
+    This function is analagous to :func:`str.lstrip`.
+    Note that the *skip_items* argument is not a prefix; any items that it
+    contains will be stripped.
+
+    """
+    return dropwhile(partial(contains, set(skip_items)), iterable)
```
MSeifert04: You could also use `dropwhile(set(skip_items).__contains__, iterable)`. That could be a bit faster but not much, most of the time will be spent inside the `set.__contains__` function.

more_itertools/more.py
```diff
@@ -1321,3 +1324,44 @@ def locate(iterable, pred=bool):
...
+    This function is analagous to :func:`str.rstrip`.
+    Note that the *skip_items* argument is not a suffix; any items that it
+    contains will be stripped.
+
+    """
+    skip_items = set(skip_items)
+    cache = []
+    for x in iterable:
+        if x in skip_items:
+            cache.append(x)
```
MSeifert04: Especially if there are lots of "skip_items" in between it could make sense to cache the `cache.append` method:

    skip_items = set(skip_items)
    cache = []
    cache_append = cache.append
    for x in iterable:
        if x in skip_items:
            cache_append(x)
        else:
            for y in cache:
                yield y
            del cache[:]
            yield x

MSeifert04:
What about `strip` on both sides? 

Personally I haven't had any use case for any of the strip, lstrip and rstrip for non-strings but for strings `strip` is used much more often than just `rstrip` or `lstrip` (no hard evidence - just my personal experience).

more_itertools/more.py
bbayles: Yes, this is a good note.

more_itertools/more.py
bbayles: Hm, yes. I was thinking of `try/catch`-ing the `set()`-ification to fall back to a `list` if the items aren't hashable.

bbayles:
Thanks for being a sounding board. I added your optimizations, the ability to use non-hashable skips, double-sided `strip()`, and some tests.

I haven't seen many requests for this on StackOverflow (and if there were many, I'm sure you'd have seen them) - [this one](https://stackoverflow.com/questions/42942230/any-efficient-way-to-strip-a-list-of-items) seems like it, though.

Another thing I was messing with: "find sub-sequence within an iterable, a.k.a. `str.find(sub, start, end)`, but for iterables" - any thoughts on adding that? I've got an implementation.

bbayles:
Re: find-subsequence, I may just add that as an example to `locate()`.

pylang:
My initial reaction, I think this is great.  I particularly like the idea of `rstrip` since I've recently needed some ways to "keep all but the last n items".  To do so for the first n items, I tend to think of  `itertools.dropwhile` or the `more_itertools.consume` recipe.   I will play with this a bit more and follow up.  

P.S. I first thought these were string-only as I associate "`strip`" with strings.  Perhaps a name like "`drop`" would suffice, similar to pandas' `drop` to remove a column.

pylang:
A lot of take/drop tools require some knowledge of the size (`n`) of the iterable ,e.g. `tail`, `head`, etc.  Few deal with actual items, e.g. `filterfalse`, `dropwhile` and `takewhile` via predicates.  The proposed "`strip`" tools seek to pursue an item comparison approach.  Ways to drop or "strip" items:

|Remaining Side|Tool|Approch|
|:-:|:-:|:-:|
|left  |`head`      |`n`|
|left  |`take`      |`n`|
|left  |`consume`   |`n`|
|left  |`takewhile` |predicate|
|right |`tail`      |`n`|
|right |`dropwhile` |predicate|
|either|`filerfalse`|predicate|
||||
|left  |`rstrip`    |comparison|
|right |`lstrip`    |comparison|
|either|`strip`     |comparison|

I think more tools that access items is a good idea just as iterating items is better than iterating indices.  Additionally:
- Size-based tools are harder to use with generators, which exhaust when trying to get the `len`.
- Accessing the nth most tail item is tricky with present tools.

Traditionally, item access in `itertools` this has been handled with predicates.  Whether comparison in `strip` (vs. general predicates) is best in the long run, that is hard to say.  I'm interested to try.   

Some possible applications:

**`lstrip`**

- [Skip first element](https://stackoverflow.com/questions/10079216/skip-first-entry-in-for-loop-in-python)
- [Skip n values](https://stackoverflow.com/questions/5509302/whats-the-best-way-of-skip-n-values-of-the-iteration-variable-in-python)


**`rstrip`**

- [Return all but last elements in a generator](https://stackoverflow.com/questions/2425096/how-to-write-a-generator-that-returns-all-but-last-items-in-the-iterable-in-pyth)
- [Detecting the last element](https://stackoverflow.com/questions/1630320/what-is-the-pythonic-way-to-detect-the-last-element-in-a-python-for-loop)

Possible renames:

- `strip, lstrip, rstrip` --> `drop, ldrop, rdrop` or a general `drop`
- `skip_items` --> `skip` or `ignore`

bbayles:
@pylang, thanks for the research - excellent.

Re: predicate vs. comparison, I think you make a good case for using a predicate function instead of item comparison here. I'll make that change.

Re: naming, I still like the `strip` names better than `drop`, but am open to other suggestions.

Re: other item access functions, I've thought before about adding support for "negative `n`" to `head()` and `tail()`.

For `head()` it would mean "select all but the last `n` lines". This matches what [GNU head](https://www.gnu.org/software/coreutils/manual/html_node/head-invocation.html) does.

For `tail()` it would be "select all but the first `n` lines".  Unfortunately [GNU tail](https://www.gnu.org/software/coreutils/manual/html_node/tail-invocation.html#tail-invocation) uses a `+` sign instead.

erikrose:
> "keep all but the last n items"

Did you need to do something beyond what `itertools.islice` would give you?

bbayles:
`islice()` can go from the left but not the right, yeah? Negative slicing isn't supported.

Maybe that's a better `more_itertools` addition than "negative n for `tail`"; I'll fiddle with it.

bbayles:
Merging this - thanks all for the feedback.

erikrose:
(You're right about islice().)

PR #146, opened 2017-06-03
islice_extended function

bbayles:
Re: [this discussion](https://github.com/erikrose/more-itertools/pull/145#issuecomment-305239617), this PR adds `islice_extended()`, which is like `islice()`, but without the restrictions on the arguments.

The goal here is to do slices with negative arguments without needing to store the entire iterable. This is generally possible, but a bit tricky since we don't know the length of the iterable.

Some cases are easier than others. I wrote them all out and then consolidated ones with similar logic. Below I'll sketch out one, but I think the comments in the code should describe the others pretty well.

---

To work through one example (positive start, negative stop, positive step), suppose we have `islice_extended(iter('ABCDEFGH'), 2, -2, 1)` we'll want to wind up with this (blue means yielded):
image

We can start by advancing to the `start` position; we don't need anything before that:
image

We don't know where `-2` is, so we'll cache two items in `deque` (orange means cached):
image

As we iterate through we'll pop the leftmost item in our cache and add the current item to the right side of the cache:
image
image
image
image

---

As the CPython docs [say](https://github.com/python/cpython/blob/2ddf5a19c3a06978edff2c8ba0aaf5df3528204a/Objects/sliceobject.c#L198), slice indexing is hard to get right! I *think* I've covered the bases here, but watchers, tell me if you see something amiss.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -1419,7 +1419,7 @@ def islice_extended(iterable, *args):
     argc = len(args)
     if argc == 1:
         stop, = args
-        start = type(stop)(0)
+        start = 0
         step = 1
     elif argc == 2:
         start, stop = args
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -1416,25 +1416,12 @@ def islice_extended(iterable, *args):
         [110, 108, 106, 104, 102, 100]
 
     """
-    argc = len(args)
-    if argc == 1:
-        stop, = args
-        start = 0
-        step = 1
-    elif argc == 2:
-        start, stop = args
-        step = 1
-    elif argc == 3:
-        start, stop, step = args
-    else:
-        err_msg = 'islice_extended takes at most 4 arguments, got {}'
-        raise TypeError(err_msg.format(argc))
-
-    if step == 0:
-        raise ValueError('step argument must be nonzero')
-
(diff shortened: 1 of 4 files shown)


erikrose:
This is a clever and very itertoolsy approach. :-)

bbayles:
I found a couple problem cases to fix by using generated tests instead of manual cases; I'll fix those.

pylang:
The concept is great.   I'm intrigued by how you handled negative steps. +1

bbayles:
I've added more systematic testing and fixed an over-reliance on caching when `step` is negative and `stop` is `None`.

I'll merge unless there are objections - many thanks; this was fun to puzzle through.

I've found a bug in slicing for `peekable` in doing this; I'll make a separate PR for that.

more_itertools/more.py
```diff
@@ -1394,3 +1395,115 @@ def strip(iterable, pred):
 
     """
     return rstrip(lstrip(iterable, pred), pred)
+
+
+def islice_extended(iterable, *args):
+    """An extension of :func:`itertools.islice` that supports negative values
+    for *stop*, *start*, and *step*.
+
+        >>> from more_itertools import ncycles
```
pylang: `ncycles` is imported and not used

PR #148, opened 2017-06-11
Fix some cases where peekable slicing gives the wrong answer

bbayles:
This PR fixes an issue with slicing some `peekable`-wrapped iterables - certain combinations of starts, stops, and steps could give the wrong answer.

The tests are now exhaustive (I think?), so hopefully this won't need to be repeated.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -15,7 +15,7 @@ from itertools import (
     tee
 )
 from operator import itemgetter, lt, gt
-from sys import version_info
+from sys import maxsize, version_info
 
 from six import binary_type, string_types, text_type
 from six.moves import filter, map, range, zip, zip_longest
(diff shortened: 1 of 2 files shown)


more_itertools/more.py
```diff
@@ -253,24 +253,28 @@ def next(self):
...
-        if (
-            ((start is not None) and (start < 0)) or
-            ((stop is not None) and (stop < 0))
-        ):
-            stop = None
-        elif (
-            (start is not None) and (stop is not None) and (start > stop)
-        ):
-            stop = start + 1
+        # Normalize the slice's arguments
```
bbayles: This procedure is the same as what [CPython does](https://github.com/python/cpython/blob/2ddf5a19c3a06978edff2c8ba0aaf5df3528204a/Objects/sliceobject.c#L221-L233) for normal sequences.

Issue #151, opened 2017-06-22
Intersperse every n items


pylang:
From [this SO post][1], given 

```python
number = 123456789012345678901234567890
expected = "12345 67890 12345 67890 12345 67890"
```

This looks like an opportunity for `interspearse`.  However, the present implementation "injects" (actually zips) one unique element between every element of the iterable.  I propose modifying `interspearse`  to inject an element between every n elements, e.g. a space every 5 characters in the `expected` string.

Here is a quick modification to  [the `intersperse` code][3] adding an `n` keyword argument:

```python
import itertools

import more_itertools as mit


def intersperse(e, iterable, n=1):
    it = iter(mit.chunked(iterable, n))                    # dependency 
    filler = itertools.repeat(e)         
    zipped = mit.collapse(zip(filler, it))                 # dependency
    next(zipped)
    return zipped
```

Results

```python
print(list(intersperse('x', 'ABCD')))
print(list(intersperse('x', 'ABCD', 2)))
# ['A', 'x', 'B', 'x', 'C', 'x', 'D']
# ['A', 'B', 'x', 'C', 'D']


print(list(intersperse(None, [1,2,3])))
print(list(intersperse(None, [1,2,3], 2)))
# [1, None, 2, None, 3]
# [1, 2, None, 3]

print("".join(intersperse(" ", str(number), 5)))
# 12345 67890 12345 67890 12345 67890

```
These are minor changes, i.e. adding `chunked` and substituting `flatten` with `collapse`.  The downside is that this modified implementation depends on other tools.  I imagine the hope is to keep new recipes independent.   Before proceeding, are there any thoughts on adding a keyword, suggestions for a different implementation or desire to keep as is?

Related posts
- [Insert element in Python list after every nth element][5]
- [Pythonic way to insert every 2 elements in a string][6]
- [How do I insert a space after a certain amount of characters ...][8]
- [Python - How to add space on each 3 characters?][7]
- [Merge lists in Python by placing every nth item ...][2]

[1]: https://stackoverflow.com/questions/21196761/python-how-to-print-space-per-nth-char/21197057#21197057
[2]: https://stackoverflow.com/questions/34692738/merge-lists-in-python-by-placing-every-nth-item-from-one-list-and-others-from-an
[3]: https://github.com/erikrose/more-itertools/blob/master/more_itertools/more.py#L498
[5]: https://stackoverflow.com/questions/31040525/insert-element-in-python-list-after-every-nth-element
[...]

PR #152, opened 2017-06-25
Intersperse n

bbayles:
Re: issue #151, this PR incorporates @pylang's method for interspersing every `n` items.

I confirmed that the `n=1` case is slowed down several times by the generalization, so I made that a special case here. I couldn't find any public code on Github that uses `intersperse`, but if there's anyone out there maybe they'll appreciate the lack of performance regression.


pylang:
Thanks.  Much appreciated for making the PR.

more_itertools/tests/test_more.py
```diff
@@ -407,18 +407,39 @@ class IntersperseTest(TestCase):
...
+            (5, '_' , ['0', '1', '2', '3', '4', '_', '5']),
+            (6, '_' , ['0', '1', '2', '3', '4', '5']),
+            (7, '_' , ['0', '1', '2', '3', '4', '5']),
+            (3, ['a', 'b'], ['0', '1', '2', ['a', 'b'], '3', '4', '5']),
+        ]:
+            iterable = (x for x in '012345')
+            actual = list(intersperse(element, iterable, n=n))
+            eq_(actual, expected)
+
+        assert_raises(ValueError, lambda: list(intersperse('x', '012', n=0)))
```
pylang: I would make  `assert_raises(ValueError, lambda: list(intersperse('x', '012', n=0)))` a separate test.  I tried testing new implementation and it took time to realize this was the source of the failure.

PR #153, opened 2017-06-26
Add transition()

hristog:


bbayles:
Many thanks for the PR. This is interesting - is this motivated by some real-world use case?

I'm inclined to merge a version of this. Here are my notes:
* I'd like it if the intermediate results were yielded, rather than just the final result.
    * Given that, I think the first item from the iterable should be yielded unchanged, followed by each state
    * You could put an example in the docstring that shows using `tail(1, transition(*args))` to get the last result.
* The empty iterable case should be handled
* `api.rst` should be updated to include the new tool

Below is an implementation that yields intermediate results, as well as sample input and ouput.

Thanks again!

---

```python
def transition(*iterables):
    it = iter(iterables)

    # Yield the first state unchanged
    curr = {}
    for curr in it:
        yield curr.copy()
        break

    # Compute the next new state and yield it
    for next_ in it:
        curr = {k: next_[v] for k, v in iteritems(curr) if v in next_}
        yield curr.copy()
```

Example:

```python
# Input
{0: 5, 1: 5, 2: 4, 3: 3, 4: 5}
{0: 5, 1: 3, 2: 0, 3: 4, 4: 3}
{0: 4, 1: 3, 2: 1, 3: 3, 4: 5}
{0: 4, 1: 4, 2: 4, 3: 4, 4: 5}
{0: 5, 1: 5, 2: 5, 3: 4, 4: 5}

# Output
{0: 5, 1: 5, 2: 4, 3: 3, 4: 5}
{2: 3, 3: 4}
{2: 3, 3: 5}
{2: 4}
{2: 5}
```

hristog:
Hi @bbayles, yes, this was inspired by a situation where I had two mappings, `A->B` and `B->C`, and I had to have them around unmodified, but also use a combined form of them, that I began referring to as a transition map, since it was essentially `A->C`. And I thought others might have also into similar cases.

Thanks for your notes - I like the idea of `yield`ing intermediate results, too. My updated solution will involve `take(1, it)` instead of entering a loop for just one iteration. I'll create a new commit that addresses them at some point later today.

A couple of questions, though:
- Is `curr.copy()` really needed in the latter case (in your example), given that we've constructed a new dictionary via comprehension by that point?
- Given that `.copy` on `dict`s creates a shallow copy, does it really help in the former case either? With or without `.copy()`, modifying the result of `transition()` (in the case of a single iterable passed) will affect the original iterable as well. Or am I missing something?

Thanks!

bbayles:
Re: `copy`, should add to the docstring that modifying the results of `transition` during iteration will produce unexpected behavior.

hristog:
I wasn't sure what you meant regarding the `copy` bulletpoints - I'll update the implementation with a further commit, if you've actually meant that I should include `.copy()` corresponding to all `yield` statements.

All of your other notes have been addressed, I believe. Please, do let me know if I've missed anything.

bbayles:
Thanks for the updates.

As I play with this, I'm on the fence about including the function. I looked through some of my code and couldn't find a place where I could have used it - the closest were some math problems from Project Euler. If you can find something on StackOverflow or ActiveState where this would be applicable I'd certainly merge?

I haven't articulated a consistent policy for what's in scope for this library, apologies.  I'm very interested in encouraging people to propose additions and improvements, so don't take this as a slight!

hristog:
Thanks for your message.

Yes, apart from coming up with alternative solutions to programming puzzles, I don't think the `transition` functionality (just as most tools from the various convenience packages) - applied over more than two iterables - will come up a lot in real world projects (of any size).

As I mentioned above - the two-maps transition case did come up in one of my projects, and I wanted to share it with the community, naturally. The problem generalizes well, and is a very nice recipe - that's basically the justification of my PR.

I do understand your position, so - no hard feelings. However, whilst I haven't (and don't have time to) performed an ActiveState or StackOverflow research directed at finding users who were in need of a similar tool, I'm afraid I haven't come across with anything similar online so far. Thus, my generalized formulation of the function and corresponding PR.

PR #160, opened 2017-08-27
Use Counter to count the items in "distinct_permutations".

MSeifert04:
In `distinct_permutations` there was a loop that could be replaced by a simple `collections.Counter`. Given #36 it seemed like it was for python 2.6 compatibility. Is that still required (it's not included in the test matrix)?

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -487,9 +487,7 @@ def distinct_permutations(iterable):
                     yield x
                 item_counts[item] += 1
 
-    item_counts = {}
-    for item in iterable:
-        item_counts[item] = item_counts.get(item, 0) + 1
+    item_counts = Counter(iterable)
 
     return perm_unique_helper(item_counts, [None] * len(iterable),
                               len(iterable) - 1)


bbayles:
Is there any performance difference from this? As you noticed, this was in the [original implementation](https://github.com/erikrose/more-itertools/pull/36/commits/960609504cb386bf5469752e0741a2f62e5d0b9c#diff-a7f69885eda5d871c519f326bf410e62R259) and removed for the now-not-relevant Python 2.6. I'd be interested in a re-run of the methods [Erik posted](https://github.com/erikrose/more-itertools/pull/36#issuecomment-37902314) back then?

MSeifert04:
That's complicated. It's very likely slower on Python 2 but it will be faster for moderately sized `iterables` in Python 3 because `Counter` uses a C function to count in Python 3. However in Python 3 it will be slower for short iterables because `Counter.update` has a quite high constant factor.

That's all about tradeoffs.

However I don't think that's really important because it would be just micro-optimizing when most of the time is spent in the `perm_unique_helper`:

```
from more_itertools import distinct_permutations
%timeit distinct_permutations([1, 2, 3, 4, 5])
# 12.7 µs ± 209 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

%timeit list(distinct_permutations([1, 2, 3, 4, 5]))
# 1.41 ms ± 24 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

erikrose:
I tend to optimize for large datasets in most cases, since small ones will be fast in any case.

bbayles:
In this case if your dataset gets particularly large you're not going to get an answer at all...

Issue #161, opened 2017-08-28
`lstrip` vs. `dropwhlie`


pylang:
I notice `more_itertools.lstrip` is similar to `itertools.dropwhile`:

```python
import itertools as it

import more_itertools as mit


iterable = [0, None, 1, 2, 0, 3, None, 0]
pred = lambda x: x in {None, 0}

list(mit.lstrip(iterable, pred))
# [1, 2, 0, 3, None, 0]

list(it.dropwhile(pred, iterable))
# [1, 2, 0, 3, None, 0]
```

I recall `lstrip` is a derivative of `strip`, but It may be worth noting in the docstring the similarity between `lstrip` and `dropwhile` (see #122).

bbayles:
This is intentional; `lstrip` is only there as a companion to `rstrip`. The former is just another name for `dropwhile`, but the latter is more complicated.

pylang:
Understood, as evident in the [source][1].  I'm suggesting a minor note in the docs for clarity.  Example: 

> Note that this a wrapper of ``dropwhile()`` from the `itertools` module.

  [0]: https://github.com/erikrose/more-itertools/blob/master/more_itertools/more.py#L1364

PR #162, opened 2017-08-30
Switch away from nose

bbayles:
Issue #51 attempted to remove `nose`. I [said](https://github.com/erikrose/more-itertools/pull/51#issuecomment-258636373) at the time that I would "get to a place where testing is possible with built-in tools." This PR gets us most of the way to that place.

A few points:
* The test modules have had several authors, some using `nose`'s methods and some using `unittest`'s, some using classes and some using functions, etc. Now everything using `unittest` methods and classes.
* `nose` was being used to run the doctests. I've switched to the [`unittest` + `doctest` method](https://docs.python.org/2/library/doctest.html#unittest-api)
* I'm inclined to remove `tox`, but [I said I wouldn't once before](https://github.com/erikrose/more-itertools/pull/61#issuecomment-258627046). It's confined to manual tests; as of PR #111 it's not used by Travis.

Eventually I will clean up tests a bit more.

__BONUS__: 0732f27 resolves issue #161.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ python:
 
 install:
     - "pip install ."
-    - "pip install -U coveralls nose"
+    - "pip install -U coveralls"
 
 script:
     - "coverage run --include='more_itertools/*.py' --omit='more_itertools/tests/*' -m unittest discover -v"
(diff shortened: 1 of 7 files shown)


pylang:
Nice work.  +1 for unification.

PR #163, opened 2017-09-17
Run flake8 style checker in Travis

bbayles:
This PR adds the [flake8](http://flake8.pycqa.org/en/latest/) linter and style checker to the automated tests. This helps ensure PEP 8 compliance in the code (and in my experience, helps me find stupid bugs before pushing them publicly and being embarrassed).

This entails a few changes, which I know some don't like. Feel free to propose an alternative, but I'm pretty set on this!
* No more `import *` in the tests (the `__all__` attributes are tested via the `__init__` module)
* No more long lines (I don't like 79 either, but YCAGWYW)
* Some linter refuse (`  # noqa`), but thankfully just a few instances

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,7 @@
 # Hack to prevent stupid error on exit of `python setup.py test`. (See
 # http://www.eby-sarna.com/pipermail/peak/2010-May/003357.html.)
 try:
-    import multiprocessing
+    import multiprocessing  # noqa
 except ImportError:
     pass
 from re import sub
--- a/setup.py
+++ b/setup.py
@@ -9,14 +9,28 @@ from re import sub
 from setuptools import setup, find_packages
 
 
+def get_long_description():
+    # Fix display issues on PyPI caused by RST markup
+    readme = open('README.rst').read()
+
+    version_lines = []
+    with open('docs/versions.rst') as infile:
+        next(infile)
+        for line in infile:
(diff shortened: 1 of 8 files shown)


setup.py
```diff
@@ -9,14 +9,28 @@
 from setuptools import setup, find_packages
 
 
+def get_long_description():
```
bbayles: OK, yes, I turned a one-liner into a fifteen-liner.

erikrose:
A net win, noqas aside. And I don't see how to get rid of them without turning off unread symbol detection altogether. So on we go!

erikrose:
For some reason I stopped reading early in the diff. Comments on the rest:

It's fine with me if you want to rewrite my big one-liner in setup.py for clarity (though I'm less enthusiastic if it's just to satisfy an arbitrary line-length constraint). I guess the thing I'm least enthused about is the shallow imports (`mi.whatever`) in the tests, since it adds a lot of noise while reading them and I can't see the gain. But you're the one reading them more than I am, so follow your own aesthetic if you have a reason.

Thanks, as always, for all your care and work!

bbayles:
Re: one-liner, I started messing with it for line-length, realized I didn't understand it the way I thought it did, and then broke it up.

Re: `mi.` - I might switch to `from more_itertools import each, little, thing`, but not here.

Issue #164, opened 2017-10-06
always_iterable treats dict as iterable


jaraco:
I'm finally getting around to reconciling `always_iterable` in [jaraco.itertools](/jaraco/jaraco.itertools), as contributed to this project in #37 and #108.

In September last year, I discovered that one would be unlikely to want to iterate over a dictionary (or other Mapping) when using `always_iterable`, so in [jaraco.itertools 2.0](http://jaracoitertools.readthedocs.io/en/latest/history.html#id2), Mappings were treated as non-iterable. The reasoning, as [found in the docs](http://jaracoitertools.readthedocs.io/en/latest/#jaraco.itertools.always_iterable), is that a dictionary is likely to be intended as a single object rather than a sequence of keys, but also that one can readily pass iter(dict) or dict.keys() if one does want the value to be treated as iterable.

In order for more_itertools.more.always_iterable to supplant jaraco.itertools.always_iterable, I'd like for more_itertools to adopt this behavior as well.

What do you think?

bbayles:
My first intuition was that `always_iterable()` should let you loop over dict keys like normal, but I can see your point.

I [sort of wanted](https://github.com/erikrose/more-itertools/pull/108#discussion_r96780782) to take out the special handling of strings to make the default behavior super simple, but decided against it then.

I think PR #165 adds some flexibility such that we don't have to dictate which types are "not considered iterable." This is similar to what `collapse()` does, and can cover the `dict` case you mention.

I considered adding something for "don't override anything, just check `__iter__` if `base_type` is `None`" - any opinions about that? As-is you could do `base_type=type('Dummy', (), {})`.

bbayles:
Sort of resolved in PR #165.

PR #165, opened 2017-10-07
Add base_type argument to always_iterable()

bbayles:
Re: Issue #164, this PR adds a `base_type` argument to `always_iterable`.

This allows for the "by default, strings are not considered iterable" behavior to be overridden, and allows other types (e.g. `dict`) to be considered non-iterable.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -1119,7 +1119,7 @@ def divide(n, iterable):
     return ret
 
 
-def always_iterable(obj):
+def always_iterable(obj, base_type=None):
     """
     Given an object, always return an iterable.
 
@@ -1139,10 +1139,15 @@ def always_iterable(obj):
         >>> always_iterable([1, 2, 3])
         [1, 2, 3]
 
-    Strings (binary or unicode) are not considered to be iterable::
+    By default, strings (binary or unicode) are not considered to be iterable::
 
         >>> always_iterable('foo')
         ('foo',)
+
+    To consider other types as non-iterable, set *base_type*::
+
+        >>> always_iterable({'a': 1})
+        {'a': 1}
 
     This function is useful in applications where a passed parameter may be
     either a single item or a collection of items::
@@ -1162,8 +1167,9 @@ def always_iterable(obj):
     if obj is None:
         return ()
 
-    string_like_types = (text_type, binary_type)
-    if isinstance(obj, string_like_types) or not hasattr(obj, '__iter__'):
+    base_type = (text_type, binary_type) if base_type is None else base_type
+
+    if isinstance(obj, base_type) or not hasattr(obj, '__iter__'):
         return obj,
 
     return obj
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -1146,8 +1146,6 @@ def always_iterable(obj, base_type=(text_type, binary_type)):
         >>> obj = 'foo'
         >>> list(always_iterable(obj))  # Default behavior
         ['foo']
-        >>> list(always_iterable(obj, base_type=None))  # No special handling
-        ['f', 'o', 'o']
 
         >>> obj = {'a': 1}
         >>> list(always_iterable(obj))  # Default behavior
@@ -1171,10 +1169,7 @@ def always_iterable(obj, base_type=(text_type, binary_type)):
     if obj is None:
         return ()
 
-    is_base_type = False if base_type is None else isinstance(obj, base_type)
-    is_iterable = hasattr(obj, '__iter__')
-
-    if is_base_type or not is_iterable:
+    if isinstance(obj, base_type) or not hasattr(obj, '__iter__'):
         return obj,
 
     return obj
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -1139,35 +1139,42 @@ def always_iterable(obj, base_type=(text_type, binary_type)):
         >>> always_iterable([1, 2, 3])
         [1, 2, 3]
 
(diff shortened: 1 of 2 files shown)


bbayles:
Oops, looks like string types don't have `__iter__` on Python 2.

more_itertools/more.py
```diff
@@ -1119,7 +1119,7 @@ def divide(n, iterable):
     return ret
 
 
-def always_iterable(obj):
+def always_iterable(obj, base_type=(text_type, binary_type)):
```
erikrose: I like this factoring: flexibility but not in your face. Can I propose a few other spellings that I think read better?

`always_iterable(foo, but_not=(some_type,))`
`always_iterable(foo, except_for=(some_type,))`

bbayles: Hmm, yes. I don't think either `but_not` or `except_for` really make my parameter naming dreams come true.

`ignore_types`?
`override_types`?
`no_iterate`?
`leave_alone`?
`unitary_types`?
`consider_non_iterable`?
`consider_unitary`?

 I don't love `base_type`, but we already have it for `collapse()`.

jaraco:
I'm warming up to this idea. I like the generalization.

Thinking about the use-case, where dictionary objects are basic, it's also used for mixed types, so in my usage, it will  be something like:

```
item = always_iterable(input, base_type=(dict,) + six.string_types)
```

Which is starting to have enough complexity to justify a bespoke function with documentation.

On the other hand, I'm liking that a user can readily override the types to enable strings or bytes to be treated as iterable as well.

My preference would be for `base_type` to by default exclude objects that one would not normally expect to be treated as iterables, and my experience (and the jaraco.itertools implementation) tells me that dictionaries should not be treated as iterable by default.

I would be much more enthusiastic about this change if the default `base_type` included `dict`.

more_itertools/more.py
jaraco: In my mind, consistency wins, so perhaps `base_type` is best. I looked at the docs for `isinstance`, and its parameter is `class_or_tuple`, which isn't very inspiring. I do like `unitary_types`.

bbayles: I think I'm going to make the signature `always_iterable(obj, *unitary_types)` - that way it's more obvious you can pass in one argument or many.

jaraco: hmm. I don't like `*unitary_types` - that deviates from standard practice as found in `collapse()`, `isinstance()`, and in the `except` clause. Plus, it confines the signature (satisfying all positional arguments and disallowing passing by keyword):

```
>>> def always_iterable(item, *unitary_types): pass
... 
>>> always_iterable(None, unitary_types=(str,))
TypeError: always_iterable() got an unexpected keyword argument 'unitary_types'
```

Maybe the signature will never need an additional parameter, but my instinct is a single parameter is preferable.

bbayles: Good points.

jaraco:
I had one other thought. If we wished for this API to be even more general and more oriented to the functional programming style, this additional parameter could be the discriminating function itself:

```
def always_iterable(items, is_object=lambda ob: isinstance(ob, (str, bytes))):
	if is_object(items) or not hasattr(ob, '__iter__'):
		return items,
	...
```

I'm reluctant to actually recommend this approach because

- it's again inconsistent with other functions in the same module
- the extra layer of abstraction may be non-intuitive and even confusing to novice users... the whole function is applying too much meta-programming for its simple purpose.
- a lambda is required because `isinstance` doesn't take keyword arguments (otherwise, one could construct the default or an alternative with `functools.partial(isinstance, class_or_tuple=...)`

Still, I wanted to share it as something I was considering.

more_itertools/more.py
jaraco: I'm still +1 for `base_type` and I would suggest only using another name for this parameter if at the same time, the library deprecates `base_type` in `collapse` in favor of the name elected for `always_iterable`.

bbayles:
Handing off to a user-supplied function is not a bad idea, but I think I agree with you in ultimately not endorsing it.

One thing that's sort of bugged me is that string types would be treated differently in Python 2 and 3 if `base_type` is set to something non-default.

I fix that in the latest commit by using `iter()` to check for iterability, not the presence of `__iter__`. Now we won't miss things that support the sequence protocol and don't define `__iter__`.

image

I'm pretty set on mostly leaving it up to Python to determine what's iterable (so you get `dict` keys by default), with a carve-out for strings because they're hard to get right across versions. I looked through some of my code and found that dealing with the AWS API I would like to have `dict`s be "basic," but in other places would like to have keys.

I'm not set on returning `iter()`-wrapped things, though - I can go back to "tuples or the object itself" if that seems better.

more_itertools/more.py
erikrose: I didn't realize we used `base_type` elsewhere. Consistency über alles. Consider my kvetching about naming withdrawn.

bbayles:
Last call for this one. I'll update release notes with a note about the change in returned type, of course.

jaraco:
Sorry. I got distracted by other things.

> I'm not set on returning iter()-wrapped things, though - I can go back to "tuples or the object itself" if that seems better.

In my mind, the result is "iterable" and not "an iterator", so wrapping each result in `iter` feels like an extra, unnecessary layer of abstraction. But since iterators are iterable, either approach is acceptable to me. I do like that it essentially enforces that the result is not a re-usable iterator, so a developer wouldn't be tempted to assume the result is a tuple or list or other re-usable iterator.

So I guess I could go either way. Glad to get this feature.

PR #167, opened 2017-11-20
Add validator parameter to bucket()

bbayles:
This PR adds a new keyword argument to `bucket()`, which helps protect against accidentally exhausting memory when checking for membership.

I was initially hung up on this issue when doing [PR 66](https://github.com/erikrose/more-itertools/pull/66#issuecomment-259952519), thinking that you should have to specify your keys to `bucket()` up-front. However, I ultimately conceded that this was too limiting.

[This discussion](https://groups.google.com/forum/#!searchin/python-ideas/itertools|sort:date/python-ideas/L6qwK69xkIA/78rlTjK5BAAJ) in `python-ideas` sparked an idea for a solution - an optional validation function.

If you know what keys will be in the iterable, you can convey that and then not worry about checking for membership.

Before:
```python
>>> iterable = count(1)  # All numbers starting with 1
>>> key = lambda x: int(str(x)[0])  # First digit of each number
>>> D = bucket(iterable, key)
>>> 0 in D  # Run out of memory and die
>>> list(D[0])  # Run out of memory and die
```

After:
```python
>>> iterable = count(1)  # All numbers starting with 1
>>> key = lambda x: int(str(x)[0])  # First digit of each number
>>> validator = lambda x: 0 < x < 10  # No leading zeros
>>> D = bucket(iterable, key, validator=validator)
>>> 0 in D
False
>>> list(D[0])
[]
```

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -607,7 +607,7 @@ class bucket(object):
     child iterables based on a *key* function.
 
         >>> iterable = ['a1', 'b1', 'c1', 'a2', 'b2', 'c2', 'b3']
-        >>> s = bucket(iterable, key=lambda s: s[0])
+        >>> s = bucket(iterable, key=lambda x: x[0])
         >>> a_iterable = s['a']
         >>> next(a_iterable)
         'a1'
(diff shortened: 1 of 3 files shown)


erikrose:
Seems reasonable to me. The pedant in me is a bit bothered that bucket() could actually internally store some bucket keys that don't satisfy the validator, but it might hurt performance to check them.

bbayles:
Good point - it's weird to store those.

Fiddling with the performance impact, it depends on how selective you want to be. For "very selective" the hit of the additional check is outweighed by avoiding adding to the cache.

erikrose:
Yes, mallocs can get quite expensive. I like it this new way; it feels more correct or consistent, and we can always dink around with performance optimizations later if need be. Happy Thanksgiving!

PR #168, opened 2017-11-20
Add peekable.index method

bbayles:
This PR adds an `index()` method to `peekable()`-wrapped iterables. Like `list.index(x)`, this returns the index of first occurrence of `x`. It caches values as it searches, so you don't advance the iterator by searching.

I can be talked out of this, but I'm tempted to also add `__contains__` with similar caching. (Update: I did)

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -168,7 +168,7 @@ class peekable(object):
     storage.
 
     To locate the position of the first instance of some value in the
-    iterable, use :meth:`index`. This won't advance the source iterator.
+    iterable, use :meth:`index`. This won't advance the iterator.
 
         >>> p = peekable(iter('abcdabcd'))
         >>> p.index('b')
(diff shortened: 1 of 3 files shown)


more_itertools/more.py
```diff
@@ -289,6 +319,32 @@ def __getitem__(self, index):
 
         return self._cache[index]
 
+    def index(self, value, start=0, stop=None):
```
erikrose: Did you consider delegating to `__getitem__` for the implementation? For example, call getitem with successively greater indices until we find a matching item. It would get us down to a single `for` loop and relieve us of cache maintenance duties here.

bbayles: Oh, good call - I had it in my head that this wouldn't work, but it does!

bbayles:
I think I've talked myself out of this addition; the `locate()` function is more flexible.

erikrose:
The difference is that locate() always consumes the whole iterable. But I have no strong opinion since I haven't run into this need myself.

bbayles:
You're right; `locate()` plus PR #170's `seekable()` is really what I had in mind.

My recent string of PRs were all based on trying to do [`consecutive_groups`](https://github.com/erikrose/more-itertools/pull/169). I thought I'd need `peekable.index` for that, but wound up not using it. That got me thinking about how to do progressive caching, which led to [`seekable`](https://github.com/erikrose/more-itertools/pull/170).

PR #169, opened 2017-11-22
Add consecutive_groups()

bbayles:
This PR adds `consecutive_groups()`, a tool for finding consecutive blocks of items in an iterable. This is an extension of the example in the [old itertools docs](https://docs.python.org/release/2.4.4/lib/itertools-example.html).

--- 

I've always admired the trick offered in the example, which is excerpted here:

```python
>>> data = [ 1,  4,5,6, 10, 15,16,17,18, 22, 25,26,27,28]
>>> for k, g in groupby(enumerate(data), lambda (i,x):i-x):
...     print map(operator.itemgetter(1), g)
... 
[1]
[4, 5, 6]
[10]
[15, 16, 17, 18]
[22]
[25, 26, 27, 28]

```

This works because `enumerate` counts up by 1 for each item in the iterable (`0, 1, 2, 3...`), so if we have a run of numbers (`20, 21, 22, 23`), we'll wind up with a constant difference between the index and each item in the run (`20 - 0 = 20`, `21 - 1 = 20`, `22 - 2 = 20`, ...). This we use for grouping.

---

The example shows how to work with iterables of numbers, but the same method can be applied to any sequence with a clear ordering. I've used this for finding consecutive IP addresses, for example.

```python
>>> from socket import inet_aton
... from struct import unpack
... 
... from more_itertools import consecutive_groups
... 
... iterable = ['192.0.2.0', '192.0.2.2', '192.0.2.3', '192.0.2.5']
... ordering = lambda x: unpack(b'!I', inet_aton(x))[0]
... 
... for group in consecutive_groups(iterable, ordering):
...     print(list(group))
['192.0.2.0']
['192.0.2.2', '192.0.2.3']
['192.0.2.5']
```

The [tests](https://github.com/erikrose/more-itertools/compare/consecutive-groups-rebased?expand=1#diff-228d30171f84b762c55cb716df56ec37R1457) show another example where there's not an obvious numerical representation, but all we need is a ranking method.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -124,6 +124,7 @@ These tools return summarized or aggregated data from an iterable.
 .. autofunction:: one
 .. autofunction:: unique_to_each
 .. autofunction:: locate
+.. autofunction:: consecutive_groups
 
 ----
 
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -29,6 +29,7 @@ __all__ = [
     'chunked',
     'collapse',
     'collate',
+    'consecutive_groups',
     'consumer',
     'count_cycle',
     'distinct_permutations',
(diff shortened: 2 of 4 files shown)


pylang:
Absolutely.   I'm glad you found a general name for it.  Here's another [example from SO](https://stackoverflow.com/questions/2154249/identify-groups-of-continuous-numbers-in-a-list).

PR #170, opened 2017-11-24
Add seekable()

bbayles:
This PR adds a new wrapper class, `seekable()`. This does progressive caching of an iterable, allowing you to return to the values that have been seen so far. In addition, it allows for jumping forward and backward in the stream:

```python
>>> from more_itertools import seekable, take
>>> it = seekable((str(n) for n in count()))
>>> take(5, it)
['0', '1', '2', '3', '4']
>>> it.seek(0)
>>> take(6, it)
['0', '1', '2', '3', '4', '5']
>>> it.seek(10)
>>> take(5, it)
['10', '11', '12', '13', '14']
```

---

The question of "how do you reset an iterator?" is all over Stack Overflow. The standard answers are:
1. You can't, sorry
2. Wrap it with `list()`
3. Use `tee()` to copy it

I think this class improves on (2) - you can use it with infinite iterators (for a while, anyway).

I'm certain this class offers a usability improvement compared to (3).

I was originally going to call this `rewindable()` or `resettable()` or something and just have it seek to 0, but then I realized that it's pretty easy to pick an arbitrary index. The name "seekable" just happens to pair nicely with "peekable."

I fiddled with adding similar functionality to `peekable()`, but I couldn't make it seem natural.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -29,8 +29,8 @@ These tools yield groups of items from a source iterable.
 .. autofunction:: partition
 
 
-Lookahead
-=========
+Lookahead and lookback
+======================
 
 These tools peek at an iterable's values without advancing it.
 
@@ -41,6 +41,7 @@ These tools peek at an iterable's values without advancing it.
 
 .. autofunction:: spy
 .. autoclass:: peekable
+.. autoclass:: seekable
 
 
 Windowing
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -1620,10 +1620,9 @@ class seekable(object):
             else:
                 return item
 
-        if self._cache_iter is None:
-            item = next(self._source)
-            self._cache.append(item)
-            return item
+        item = next(self._source)
+        self._cache.append(item)
+        return item
 
     next = __next__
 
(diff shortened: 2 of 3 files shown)


pylang:
Looks cool.  Forgive me, I only took a peek at your `seekable` (no pun).  

I see it works on infinite iterators.  I'm interested in this `rewindable` idea.  Does it in fact rewind as well?  In other words, after `it.seek(10)`, `take(5, it)` can you `it.seek(0)`, `take(5, it)` --> `['0', '1', '2', '3', '4', '5']` or is the prior result exhausted?


Can you add this to the example?

bbayles:
There's no cache clearing, so you can seek back and forth without losing prior results:

```python
... from more_itertools import seekable, take
... it = seekable((str(n) for n in count()))
>>> it.seek(10)
>>> take(5, it)
['10', '11', '12', '13', '14']
>>> it.seek(0)
>>> take(5, it)
['0', '1', '2', '3', '4']
```

I have a [test](https://github.com/erikrose/more-itertools/pull/170/files#diff-228d30171f84b762c55cb716df56ec37R1509) with this case in it, but I'll add it to the second example.

PR #171, opened 2017-11-26
Add exactly_n()

michael-celani:
This pull request adds the function `exactly_n(iterable, n)`. This function is designed to be used like the built-in `any()` and `all()` functions. It will return `True` if and only if the input `iterable` has `n` `True` elements.

```python
>>> exactly_n([True, True, False], 2)
True
>>> exactly_n([True, True, False], 1)
False
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,4 @@ _build
 
 # IDE files
 .idea
+.vscode
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -1590,8 +1590,13 @@ def consecutive_groups(iterable, ordering=lambda x: x):
         yield map(itemgetter(1), g)
 
 
-def exactly_n(iterable, n):
-    """Return ``True`` if exactly ``n`` items in the iterable are ``True``.
+def exactly_n(iterable, n, predicate=bool):
+    """Return ``True`` if exactly ``n`` items in the iterable are ``True``
+    according to the predicate, a function of one argument which returns
+    a boolean.
+    
+    This function will short circuit as soon as it finds n + 1 ``True``
+    elements.
 
         >>> exactly_n([True, True, False], 2)
         True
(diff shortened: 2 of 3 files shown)


bbayles:
Many thanks for the PR - always happy to have a new contributor.

For this I'm inclined to recommend using `quantify` instead of a new itertool - I think it can do the task, and it has some more flexibility on what counts as truthy:

```python
>>> from more_itertools import quantify
>>> quantify([True, True, False]) == 2
True
>>> quantify([True, True, False]) == 1
False
>>> quantify([]) == 0
True
>>> quantify(range(100), lambda x: x < 10)
10

```

michael-celani:
While it is true that `quantify` is a better general-purpose tool, the benefit of `exactly_n` is that it will short-circuit if the number of `True` elements ever grows larger than `n` while `quantify` will always use the entire iterator. That can have performance implications if yielding an element is particularly expensive. I wouldn't be opposed to adding a `predicate` parameter to match the functionality of `quantify`. Do you think that concern is enough to justify a new itertool? If not, I'll close the request.

bbayles:
re: short circuit, good point - `quantify()` doesn't return `False` until the whole iterable is exhausted, even though it could.

My alternative would be:

```python
len(take(n + 1, filter(pred, iterable))) == n
```

That's not obvious, so I'll accept a PR with that version. Could you also:
* Add the new function to the API docs
* Add a note to the docstring about short-circuiting
* Add a test with an infinite iterator that short circuits (`exactly_n(count(), 100)` or something)

bbayles:
I caused the merge conflicts; apologies. Would you mind if I implemented my comments above? I'm gearing up for a release.

michael-celani:
I've implemented your comments in my branch, aside from the API docs - I'm wrapping my head around that still, so if you think it'd be faster for you to implement go ahead.

bbayles:
Sure - I'll do that. Thanks! I'll add your contribution to the release notes also.

PR #172, opened 2017-11-26
Add run_length encoding and decoding

bbayles:
This PR adds `run_length.encode()` and `run_length.decode()`, which compress and decompress iterables, respectively.

[This python-ideas thread](https://groups.google.com/d/msg/python-ideas/hT7rFXxoLi8/vk74mR4dCAAJ) discusses such an addition, and we [once discussed](https://github.com/erikrose/more-itertools/pull/44#issuecomment-50954715) something related here.

There are a ton of (mostly homework?) questions about this on StackOverflow as well. I didn't see any that use the trick [we use](https://github.com/erikrose/more-itertools/pull/117/commits/5c7c4e6586967a0983a803016b8e82f304838c6b) for `ilen`.


pylang:
I'm used to the odd name "compression", which admittedly can be confused with data compression (zip, tar, etc.).  How did you come about the name `run_length`?

bbayles:
[Wikipedia](https://en.wikipedia.org/wiki/Run-length_encoding) has a good overview of this topic. "Run length" is the standard name for this procedure. One presumes it comes from the output of the encoder - you get which element had a "run" of repeated elements, and the "length" of the repetition.

As you'll see in the Wikipedia article, run-length encoding is itself a basic form of data compression, and is a component of some modern compression methods in use today.

bbayles:
Any objections to this? I don't have a concrete idea of what's out of scope for this library, but this might be on the edge. I'm leaning toward yes because of all the questionable implementations on StackOverflow.

There's a clear use case for these functions, but they're both one-liners and apparently not that to come up with - [this blog post](http://wordaligned.org/articles/runlength-encoding-in-python) had the same ideas for both encode and decode years ago.

bbayles:
OK, let's do this.

PR #173, opened 2017-11-29
Add difference()

bbayles:
This PR adds `difference()`, which is the inverse of `accumulate()`.

```python
>>> from more_itertools import accumulate, difference
>>> [1, 2, 3, 4, 5]
[1, 2, 3, 4, 5]
>>> list(accumulate(_))
[1, 3, 6, 10, 15]
>>> list(difference(_))
[1, 2, 3, 4, 5]
```

Like `accumulate()`, this also allows for a custom function to be specified (i.e. one that's not `operator.sub()`):

```python
>>> list(difference('itertools', lambda y, x: '{}->{}'.format(x, y)))
['i', 'i->t', 't->e', 'e->r', 'r->t', 't->o', 'o->o', 'o->l', 'l->s']
```

Numpy calls this operation [`diff()`](https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.diff.html). There was a Python [issue](https://bugs.python.org/issue13252) that suggested `decumulate()`. In signal processing it would be the "first difference," so `difference()` sounded fine to me.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+more_itertools/more.py merge=union
+more_itertools/tests/test_more.py merge=union
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -211,12 +211,13 @@ Others
 .. autofunction:: numeric_range(start, stop, step)
 .. autofunction:: side_effect
 .. autofunction:: iterate
+.. autofunction:: difference(iterable, func=operator.sub)
 
 ----
 
 **Itertools recipes**
 
 .. autofunction:: consume
-.. autofunction:: accumulate
+.. autofunction:: accumulate(iterable, func=operator.add)
 .. autofunction:: tabulate
 .. autofunction:: repeatfunc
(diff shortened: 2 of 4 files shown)


pylang:
*For Reference*

In searching for use cases, I found one for `difference` called a ["difference machine"](https://en.wikipedia.org/wiki/Difference_engine) based on the differentiation method used in calculus: 

For the polynomial `2x^2 - 3x + 2`:

```python
p = lambda x: 2*x**2 - 3*x + 2
[p(x) for x in range(5)]
# [2, 1, 4, 11, 22]
```

Difference Engine

```python
list(difference(_))[1:]                            # 1st difference
# [-1, 3, 7, 11]

list(difference(_))[1:]                            # 2nd difference, constant
# [4, 4, 4]
```


This example supports the tool's mathematical relevance (just as ["prefix sum"](https://en.wikipedia.org/wiki/Prefix_sum) does for `itertools.accumulate`) and justifies the name choice.

PR #176, opened 2017-12-01
Allow one() to throw user-specified exceptions.

kalekundert:
The use-case for me is that I often want to present a different error message to the user depending on whether there were too many or too few items in the iterable that was supposed to just have one thing in it.  So I added ``too_short`` and ``too_long`` arguments to ``one()``.  If the iterable is either too short or too long, ``one()`` will raise the exception provided by the appropriate argument.  The default behaviour is to do the exact same thing as before (raise a ValueError in either case).

Tests and documentation are both updated.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -420,10 +420,10 @@ def with_iter(context_manager):
             yield item
 
 
-def one(iterable):
+def one(iterable, too_short=None, too_long=None):
     """Return the only element from the iterable.
 
-    Raise ValueError if the iterable is empty or longer than 1 element. For
+    Raise an exception if the iterable is empty or longer than 1 element. For
     example, assert that a DB query returns a single, unique result.
 
         >>> one(['val'])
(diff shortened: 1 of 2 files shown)


bbayles:
Thanks! I'll placate `flake8` and get this in for [4.0.0](https://github.com/erikrose/more-itertools/pull/175).

pylang:
Would it be better to generalize, allowing the user to add a custom message?

kalekundert:
Thanks for fixing up the documentation and the flake8 stuff for me.  :-)

I think it'd be better to leave the `too_short` and `too_long` arguments as actual exception objects, rather than exception classes that get filled in with a canned message.  The reason is that you may want to raise exceptions that contain additional pieces of information.  This is actually a key part of the application that made me submit this PR: I was writing some code to call a plugin, and it's an error if there are either 0 or >1 plugins with the expected name.  I wanted to write something like this:

```python
def load_plugin(name):
    all_plugins = [...] # a list of all the plugins installed on the system...
    return more_itertools.one(
            (x for x in all_plugins if x.name == name),
            too_short=NoSuchPlugin(name, all_plugins),
            too_long=DuplicatePlugin(name, all_plugins),
    )
```

With the ultimate goal of formatting a nice message for the user explaining what went wrong and being able to ask "did you mean this other plugin that has a similar name" and things like that.  

Anyways, hopefully that makes sense and rationalizes why I think it's better to have the arguments be actual exception instances rather than exception classes.

bbayles:
Makes sense! I've changed that in the other PR.

PR #178, opened 2017-12-02
Add split()

michael-celani:
We have ``split_before()``, a splitting function where lists start with the delimiting item, and ``split_after()``, a splitting function where lists end with the delimiting item, but no ``split()`` function that's analogous to ``str.split()``, which omits the delimiting item entirely. This pull request adds such a function.

Note that this itertool is not implemented using `itertools.takewhile()` because that implementation cannot tell if we exhausted the iterator because it ran out of elements or if because the last element's predicate returned `True`. This is problematic because that determines whether or not we should yield an additional empty list. I don't know how to work around that, so I thought a new itertool would be appropriate:

```python
>>> list(split('abcdc', lambda x: x == 'b'))
[['a'], ['c', 'd', 'c']]
>>> list(split('abcdcb', lambda x: x == 'b'))  # takewhile() would consume the last 'b'
[['a'], ['c', 'd', 'c'], []]
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,4 @@ _build
 # IDE files
 .idea
 .vscode
+.DS_Store
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -17,7 +17,7 @@ These tools yield groups of items from a source iterable.
 .. autofunction:: sliced
 .. autofunction:: distribute
 .. autofunction:: divide
-.. autofunction:: split
+.. autofunction:: split_at
 .. autofunction:: split_before
 .. autofunction:: split_after
 .. autofunction:: bucket
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -17,6 +17,7 @@ These tools yield groups of items from a source iterable.
 .. autofunction:: sliced
 .. autofunction:: distribute
 .. autofunction:: divide
+.. autofunction:: split
 .. autofunction:: split_before
 .. autofunction:: split_after
 .. autofunction:: bucket
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -125,6 +125,7 @@ These tools return summarized or aggregated data from an iterable.
 .. autofunction:: unique_to_each
 .. autofunction:: locate
 .. autofunction:: consecutive_groups
+.. autofunction:: exactly_n
 
 ----
 
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -55,6 +55,7 @@ __all__ = [
     'side_effect',
     'sliced',
     'sort_together',
+    'split',
     'split_after',
     'split_before',
     'spy',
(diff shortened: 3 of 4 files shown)


bbayles:
Many thanks! Let me think about this. I'd played with a version of "`str.split()`, but for iterables" when doing `split_before()` and `split_after()`, but I kept getting hung up on the analogy.

My issue was `str.split(sep, maxsplit=-1)` has two different modes of operation - grouping separators when `sep` is `None` and not grouping otherwise.

The former behavior is a straightforward application of `groupby`, and I think you have the latter here? Perhaps a different name for this would be appropriate?

michael-celani:
I don't think a different name for this would be appropriate, because people already understand the specific functionality afforded by ``str.split()`` and if they wanted to search for that in ``more_itertools`` they would search for ``split`` first.

pylang:
I've considered submitting a tool like this for `more_itertools`.  Thanks @michael-celani for the PR. There are many cases for _splitting at_ a certain occurrence.  How about the name `split_at`?  It seems less ambiguous, and fits in with `split_before` and `split_after`.

pylang:
Secondly,  I wonder whether an option should be made for retaining/discarding the delimiter.
Perhaps the default option can emulate `str,split`, with a keyword for keeping the delimited item.

```python
list(split_at('abcdc', lambda x: x == 'b'))
# [['a'], ['c', 'd', 'c']]

list(split_at('abcdc', lambda x: x == 'b'), discard=False)
# [['a'], ['b'], ['c', 'd', 'c']]
```

michael-celani:
I wouldn't be opposed to calling it ``split_at()``, as that still keeps the ``split`` part of the name.

One thing that would be worth discussing is if we should also add ``maxsplits`` parameter to the other ``split`` functions. The proposed ``discard`` parameter only reallly makes sense on ``split_at()``.

michael-celani:
One thing I want to bring up with the current implementation of the ``discard`` parameter is whether or not we should return empty lists in addition to the parameter. ``str.split()`` will return empty strings if two delimiters are in a row:

```python
>>> 'abaa'.split('a')
['', 'b', '', '']
```

Similarly, ``split_at()`` does the same thing:

```python
>>> list(map(''.join, split_at('abaa', lambda c: c  == 'a')))
['', 'b', '', '']
```

Do we still want the ``discard`` parameter to still return these "empty" iterables? @pylang, taking your examples, here's the current output:

```python
>>> s = "The rain in Spain stays mainly in the plain"
>>> words = s.split()
>>> articles = {"the", "in", "a"}
>>> pred = lambda x: x.lower() in articles

>>> list(split_at(words, pred))
[[], ['rain'], ['Spain', 'stays', 'mainly'], [], ['plain']]
>>> list(split_at(words, pred, discard=False))
[[], ['The'], ['rain'], ['in'], ['Spain', 'stays', 'mainly'], ['in'], [], ['the'], ['plain']]
```

bbayles:
I'm still thinking about this, but I think the `split_at` name is good.

Ideally I'd like the tests to be: give a string to this function and compare the result to what `str.split` would produce, given a separator. So whatever it does re: blanks.

pylang:
@michael-celani Although the empty objects seem odd, after some thought, I agree with @bbayles.  Closely emulating `str.strip`s behavior would be ideal.

bbayles:
OK, I've thought about this now. Good discussion above, and many thanks to @pylang for the research, as always.

Here's what I'd like to merge:
* The `split_at()` name is good - it fits with the other two functions, and fills the gap they leave.
* Let's drop the `discard` parameter, and always exclude the delimiter items - see below.
* Let's drop the `maxsplit` parameter. The other two functions don't have them, and stopping before the end is sort of built into iterator objects.

---

The `discard=False` results can be achieved with `groupby()`:

```python
>>> from itertools import groupby
... [list(g) for k, g in groupby(range(6), lambda x: x == 3 or x == 5)]
[[0, 1, 2], [3], [4], [5]]
```

```python
>>> from itertools import groupby
>>> s = "The rain in Spain stays mainly in the plain"
>>> articles = {'the', 'in', 'a'}
>>> pred = lambda x: x.lower() in articles
>>>  [list(g) for k, g in groupby(s.split(), pred)]
[['The'],  ['rain'],  ['in'],  ['Spain', 'stays', 'mainly'],  ['in', 'the'],  ['plain']]
```

michael-celani:
I'm fine with that - I've merged the changes into the branch.

bbayles:
Many thanks!

Issue #181, opened 2017-12-04
seekable decorator


pylang:
Before `seekable` further diffuses into the wild, I wanted to ask if it is feasible to give `seekable`/`peekable`-like tools the option to decorate generator functions?  Example:

```python
from itertools import count


@seekable
def numbers():
    """Return an infinite iterator of numeric strings."""
    return (str(n) for n in count())

it = numbers()
next(it), next(it), next(it)
# ('0', '1', '2')
```

I am particularly interested in making normal generator functions "reusable" ([see last `Counter` example in the "Generators" section](https://pymbook.readthedocs.io/en/latest/igd.html#generators)).  The `seekable` class seems it may able to do this simply.  This is just a rough idea, so I am not partial to the example mentioned here.

bbayles:
Ah, this is interesting. The normal decorator pattern does work with generator functions:

<details><summary>Function decorator</summary>

```python
>>> from functools import wraps
... from more_itertools import seekable
... 
... def seekable_generator(f):
...     @wraps(f)
...     def wrapper(*args, **kwargs):
...         return seekable(f(*args, **kwargs))
... 
...     return wrapper
... 
... @seekable_generator
... def yielder(n):
...     for i in range(n):
...         yield i
... 
... g = yielder(10)
... print(list(g))
... g.seek(0)
... print(list(g))
[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
```

</details>

---

This is really neat, but I'm not sure if that really buys much - why not just do the usual thing (wrap the return value of the generator function), as below?


<details><summary>Normal usage</summary>

```python
... g = seekable(yielder(10))
... print(list(g))
... g.seek(0)
... print(list(g))
[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
```

</details>

---

Assuming `peekable()` or `seekable()` are useful as decorators, does it make sense to make separate functions for them as decorators? That would involve making new names like `seekable_generator()` above.

An alternative would be to overload the functionality of `seekable()` to detect when it's being used as a decorator:

<details><summary>Class-based decorator</summary>

```python
... class seekable(object):
...     def __init__(self, x):
...         if callable(x):
...             self._func = x
...         else:
...             self._source = iter(x)
...         self._cache = []
...         self._index = None
...     
[...]

PR #182, opened 2017-12-06
Make seekable() a decorator

bbayles:
Re: Issue #181 , this PR takes @pylang's suggestion seriously and augments `seekable()` such that it can behave as a decorator.

My concern with my [hacky original implementation](https://github.com/erikrose/more-itertools/issues/181#issuecomment-349185921) (which checked the argument to `seekable()` to see whether it was `callable()`) was in wrapping classes that were both iterable and callable.

I think I've satisfied myself that it's possible to avoid any issues with that and preserve the original behavior of `seekable()`. So in this implementation we check whether the argument is iterable. It it is, we continue as always. If not, we check whether it's a function.

Anybody see any downsides? As you can see, I took a two-belts-and-four-suspenders approach in the tests to convince myself.

I'll look at doing `peekable()` also.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -42,7 +42,7 @@ These tools peek at an iterable's values without advancing it.
 
 .. autofunction:: spy
 .. autoclass:: peekable
-.. autoclass:: seekable
+.. autofunction:: seekable
 
 
 Windowing
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -1728,15 +1728,47 @@ class seekable(object):
         >>> next(it), next(it), next(it)
         ('0', '1', '2')
 
+    :func:`seekable` can be used as a decorator for generator functions (and
+    other callables that return iterable objects):
+
+        >>> @seekable
+        ... def yielder(n):
+        ...     for i in range(n):
+        ...         yield str(i)
+        ...
+        >>> it = yielder(5)
+        >>> list(it)
+        ['0', '1', '2', '3', '4']
+        >>> it.seek(0)
+        >>> list(it)
+        ['0', '1', '2', '3', '4']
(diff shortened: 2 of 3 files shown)


pylang:
I had to revise my notes on class decorators.  Your code makes sense to me.  I haven't come across any cons to this approach so far.

bbayles:
I remembered a thing that complicates this, which is decorating class methods. The current approach doesn't quite work for that; if it's too complex I may give up on this.

bbayles:
OK, that wasn't so bad - we can use the [`__get__` method](https://docs.python.org/3/reference/datamodel.html#object.__get__) to save the calling instance, so we can later use it to pass as `self` to the method being decorated.

[This post](https://christiankaula.com/python-decorate-method-gets-class-instance.html) shows another example.

pylang:
Good to know.  For reference I'll go ahead and add [this post](https://krzysztofzuraw.com/blog/2016/python-class-decorators.html), particularly example 2 on django-style properties.

pylang:
In 87f39a3a5bce0a851b8b4451d8c212701f039364, I want to raise [this issue with `functools.update_wrapper`][0] if it valuable for you.   It seems it does not update the `help()` documentation.

  [0]: https://stackoverflow.com/questions/25973376/functools-update-wrapper-doesnt-work-properly/25973438#25973438

bbayles:
I was aware of that, thanks. `functools.update_wrapper()` also sets `__doc__` for standalone functions, but not methods. I'm not too concerned about that or the `help()` issue - I don't think I have ever used `help()` in an interactive session.

bbayles:
Started looking at this for `seekable` too, and decided I'd rather go the wrapper function route. I'll open a new PR for that.

PR #183, opened 2017-12-08
Make seekable() and peekable() decorators

bbayles:
Re: Issue #181 , this PR takes another shot at @pylang's idea about making `peekable()` and `seekable()` act as decorators.

PR #182 had some discussion about one implementation, using decorator classes. That implementation has only [minor issues](https://github.com/erikrose/more-itertools/pull/182#issuecomment-349863951), but I found that this decorator function implementation seems cleaner.

In a nutshell, this PR:
* Adds `more._decorator_factory`, which handles turning iterable-wrapping classes into function-wrapping decorators.
* Adds thorough tests for the new function
* Uses the new function to make `peekable` and `seekable` into decorators
* Cleans up the `peekable` docstring a bit.

The new function isn't documented as part of the library and is considered internal.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -207,6 +207,23 @@ class _Peekable(object):
     negative indexes and slices are supported. Indexing will cache only the
     necessary items, but be aware that this may require significant storage.
 
+    :func:`peekable` can be used as a decorator for generator functions (and
+    other callables that return iterable objects):
+
+        >>> @peekable
+        ... def yielder(n):
+        ...     for i in range(n):
+        ...         yield str(i)
+        ...
+        >>> it = yielder(5)
+        >>> it.peek()
+        '0'
+        >>> list(it)
+        ['0', '1', '2', '3', '4']
+        >>> it.peek('default')
+        'default'
+
+
     """
     def __init__(self, iterable):
         self._it = iter(iterable)
@@ -1748,7 +1765,6 @@ class _Seekable(object):
         >>> it.seek(0)
         >>> list(it)
         ['0', '1', '2', '3', '4']
-        >>> it.seek(0)
 
     The cache grows as the source iterable progresses, so beware of wrapping
     very large or infinite iterables.
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -1,7 +1,7 @@
 from __future__ import print_function
 
 from collections import Counter, defaultdict, deque
-from functools import partial, wraps
+from functools import partial, update_wrapper, wraps
 from heapq import merge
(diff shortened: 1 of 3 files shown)


pylang:
The decorator function is easier to grok, though I was uncertain about the flexibility for wrapping classes and class methods.  It seems you nailed it well.

You've tested functions, generators and instance methods.  Is it worth including tests for class methods and static methods too?  I suppose it is possible to decorator a class that includes such methods.  If so, the signatures for each are bit different and may require special handling.

bbayles:
Good idea, yes. That also tests stacking decorators.

more_itertools/more.py
```diff
@@ -74,6 +74,32 @@
...
+        else:
+            return cls(iterable_or_callable)
+
+        @wraps(iterable_or_callable)
+        def callable_wrapper(*args, **kwargs):
+            return cls(iterable_or_callable(*args, **kwargs))
+
+        return callable_wrapper
+
+    decorator.__doc__ = cls.__doc__
```
erikrose: May be able to use @wraps for this as well

bbayles: Good call; `update_wrapper` seems to do the trick.

erikrose:
I like this patch. After all, why not make it easier and more readable to make something peekable or seekable on the production end rather than on just on the consumption one? The introspective approach in the decorator factory also relieves the user of having to remember anything new, at the cost of a few more instructions on construct. I don't think there are a lot of super-expensive `__iter__` methods out there, are there? (ORMs?)

more_itertools/more.py
```diff
@@ -74,6 +74,32 @@
 _marker = object()
 
 
+def _decorator_factory(cls):
```
erikrose: Trying to think of a name for this that allows for it returning a non-decorator also (as it does). `_either_decorator_or_wrapper`? `_optionally_a_decorator`?

bbayles: I went with `_wrapper_factory`, since that covers both - it's either wrapping the iterable or the function.

bbayles:
I think I had changed the `_Peekable` and `_Seekable` classes in the last version to avoid calling `iter()` twice, but I put them back in this attempt. I couldn't think of any places where it would be a super expensive thing, but I guess I can imagine it in the ["reiterable"](https://github.com/erikrose/more-itertools/blob/282fe757f1fbe7712c759ff612f5457e1aff7b20/more_itertools/tests/test_more.py#L1698-L1705) case.

bbayles:
Closing in favor of #187, I think.

PR #184, opened 2017-12-09
implement cyclic_permutations

hiqua:
I didn't see an implementation for https://en.wikipedia.org/wiki/Cyclic_permutation, so here is one.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -21,6 +21,7 @@ These tools yield groups of items from a source iterable.
 .. autofunction:: split_before
 .. autofunction:: split_after
 .. autofunction:: bucket
+.. autofunction:: cyclic_permutations
 
 ----
 
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -21,7 +21,6 @@ These tools yield groups of items from a source iterable.
 .. autofunction:: split_before
 .. autofunction:: split_after
 .. autofunction:: bucket
-.. autofunction:: cyclic_permutations
 
 ----
 
@@ -175,6 +174,7 @@ These tools yield combinatorial arrangements of items from iterables.
 **New itertools**
 
 .. autofunction:: distinct_permutations
+.. autofunction:: cyclic_permutations
 
 ----
 
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -7,6 +7,7 @@ from itertools import (
     chain,
     compress,
     count,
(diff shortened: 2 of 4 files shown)


bbayles:
Many thanks - this will be in the next release.

hiqua:
Many thanks to you for developing and maintaining more-itertools, I use it a lot!

pylang:
Nice contribution.   I see, this implementation casts an iterable to a list.  It also calls `len()` twice.

```python
def cyclic_permutations(iterable):
     """Return a list of cyclic permutations of *iterable*.
 
         >>> cyclic_permutations(range(4))
         [(0, 1, 2, 3), (1, 2, 3, 0), (2, 3, 0, 1), (3, 0, 1, 2)]
     """
     lst = list(iterable)
     return take(len(lst), windowed(cycle(lst), len(lst)))
```

To reduce calculations and memory consumption for long iterables, perhaps substitute with iter-like tools:

```python
from itertools import tee, cycle

from more_itertools import take, windowed, ilen


def cyclic_permutations(iterable):
    """Return a list of cyclic permutations of *iterable*.
 
        >>> cyclic_permutations(range(4))
        [(0, 1, 2, 3), (1, 2, 3, 0), (2, 3, 0, 1), (3, 0, 1, 2)]
    """
    it1, it2 = tee(iterable)
    n = ilen(it1)
    return take(n, windowed(cycle(it2), n))
```

bbayles:
Hmm, I doubt there would be a benefit from that. By their nature the combinatoric generators need to know the full sequence. For example, the first thing [itertools.permutations](https://github.com/python/cpython/blob/1707e4020fa8dca8e6a3ac4f9da105b54d597b66/Modules/itertoolsmodule.c#L3039) does is pull in the whole iterable. Same with [more_itertools.distinct_permutations](https://github.com/erikrose/more-itertools/blob/eda0e65f2c2f5a52531c69fb1d519824d4d63247/more_itertools/more.py#L527).

Also, the `list.len` isn't slow (even for long iterables) - it's [`Py_SIZE`](https://github.com/python/cpython/blob/1a5856bf9295fa73995898d576e0bedf016aee1f/Include/listobject.h#L74), which is in turn just [accessing](https://github.com/python/cpython/blob/25420fe290b98171e6d30edf9350292c21ef700e/Include/object.h#L119) the `ob_size` attribute of the object.

PR #185, opened 2017-12-12
Add seekable.elements method

bbayles:
This PR adds ~`seekable.items()`~`seekable.elements()`, which just returns a copy of the cache of things that have been iterated over so far.

~I'm having it return a copy instead of a reference to `self._cache`, because mutating the result would then affect the seeking behavior.~

~An alternative name might be `elements` - [Counter](https://docs.python.org/3/library/collections.html#collections.Counter.elements) has this, but it's a little different.~

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -217,6 +217,7 @@ Others
 .. autofunction:: side_effect
 .. autofunction:: iterate
 .. autofunction:: difference(iterable, func=operator.sub)
+.. autoclass:: SequenceView
 
 ----
 
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -56,6 +56,7 @@ __all__ = [
     'rstrip',
     'run_length',
     'seekable',
+    'SequenceView',
     'side_effect',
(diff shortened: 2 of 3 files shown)


bbayles:
Changed my mind and decided I'd rather not duplicate the cache with `.items()`. I made `SequenceView`, which is intended to do what `dict.viewitems()` does on Python 2 and what `dict.items()` does on Python 3.

I'm not sure about documenting and exposing that class, but I like how nicely the ABCs allow for it to be constructed.

pylang:
@bbayles Curious, this is a git-related question.  When the commit says:

> Merge branch 'master' into seekable-items

Did you perform a `git pull` or `git rebase`?

bbayles:
I think a [git merge origin/master](https://stackoverflow.com/a/16957483/353839).

erikrose:
I'd lean away from the name `items`: here it returns simple elements, but in the more common Python use on dicts, it returns 2-tuples.

bbayles:
`elements` it is!

PR #187, opened 2017-12-12
Add make_decorator function

bbayles:
So... the hyper-meta thing we discussed in [Issue #181](https://github.com/erikrose/more-itertools/issues/181#issuecomment-350423213) does just as good a job on `peekable` and `seekable` as the more convention PR #182, _and_ allows for fairly useful things like creating decorator versions of `chunked`.

So, like, we could use it. I'm not taking the step of making decorator versions of things in the library, but the examples show how you could do it.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -217,6 +217,7 @@ Others
 .. autofunction:: side_effect
 .. autofunction:: iterate
 .. autofunction:: difference(iterable, func=operator.sub)
+.. autofunction:: make_decorator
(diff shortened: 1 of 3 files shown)


pylang:
This looks like the right call.  Even better than #185 as this leaves the option to the user to decorator whatever they wish.

bbayles:
OK, here we go. I'll probably write up a separate documentation page showing how to do this before the next release.

PR #188, opened 2017-12-26
Add classify()

bbayles:
This PR adds a variation on `bucket()` that caches items in different categories as you work through an iterable. I'm open to suggestions on the name.

I implemented the `__contains__` and  `__getitem__` methods as shortcuts into the underlying `defaultdict`, but I could be persuaded to either (a) remove them, or (b) add others like `get`, `keys`, `items`, etc.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -28,6 +28,7 @@ __all__ = [
     'always_iterable',
     'bucket',
     'chunked',
+    'classify',
     'collapse',
     'collate',
     'consecutive_groups',
@@ -669,6 +670,8 @@ class bucket(object):
 
     The original iterable will be advanced and its items will be cached until
     they are used by the child iterables. This may require significant storage.
+    For a similar wrapper that doesn't automatically advance the original
+    iterable, see :class:`classify`.
 
     By default, attempting to select a bucket to which no items belong  will
     exhaust the iterable and cache all values.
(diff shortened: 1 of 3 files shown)


erikrose:
Hi, Bo. I (at least) could benefit from a real-world example in the docs.

bbayles:
Hmm, my motivation was roughly the `defauldict(list)` example in the docstring. Almost exactly like [here](https://github.com/obsrvbl/flowlogs-reader/blob/master/flowlogs_reader/aggregation.py#L48), where you've got a (big) API response that you want to categorize and summarize.

 My thought was that I find myself doing the "loop through and append to a `defaultdict(list)`" thing a lot ([example](https://github.com/obsrvbl/ona/blob/70135811615288bd186e762b6b656fe2357f4fbd/src/scripts/ona_service/check_point_pusher.py), [example](https://github.com/obsrvbl/ona/blob/70135811615288bd186e762b6b656fe2357f4fbd/src/scripts/ona_service/pusher.py#L190)) and that I wanted a shortcut for it.

I suppose this isn't much shorter than the loop method, since you still have to consume at least part of the iterable. An alternative is to use `side_effect`, but it's not particularly elegant.

<details><summary>Using `side_effect`</summary>

```python
iterable = imap(str, count())
key = lambda x: x[0]

D = defaultdict(list)
it = side_effect(lambda x: D[key(x)].append(x), iterable)
```
</details>

---

Maybe I should just have a function that exhausts the iterable and returns the `defaultdict(list)`, and skip the "wrap an iterator" thing?

more_itertools/more.py
```diff
@@ -1898,3 +1901,68 @@ def inner_wrapper(*args, **kwargs):
         return outer_wrapper
 
     return decorator
+
+
+class classify(object):
+    """Wrap *iterable* and return an object that progressively saves its items
```
erikrose: We could say "return a mapping" to be a bit more specific, or "a mapping-like object" if we don't fulfill the whole contract.

more_itertools/more.py
```diff
@@ -1898,3 +1901,68 @@ def inner_wrapper(*args, **kwargs):
...
+        >>> from collections import defaultdict
+        >>> mapping = defaultdict(list)
+        >>> for item in iterable:
+        ...     category = key(item)
+        ...     mapping[category].append(item)
+
+    Access the underlying :obj:`defaultdict` through the :attr:`mapping`
+    attribute:
+
+        >>> 'a' in it.mapping
```
erikrose: I suppose we oughtn't give this as an example, as it looks like you meant it to be more idiomatic to say `'a' in it` itself, given `__contains__` below.

erikrose:
Sorry it took my awhile to understand what you were getting at. I thought I could get it as a clear 1-liner, but no:

```python
>>> key = lambda x: 'odd' if x % 2 else 'even'
>>> defaultdict(list, ((key(x), x) for x in [1, 2, 3, 4, 5, 6]))
defaultdict(<class 'list'>, {'odd': 5, 'even': 6})  # 1, 2, 3, and 4 are missing.
```

Lgtm! I thought about suggesting a name akin to "bucket"—eager_bucket or something—but again, no: it's got too many differences from `bucket` to be worth drawing a comparison.

bbayles:
I turned lukewarm on this. Maybe I'll take another shot eventually.

Issue #122, opened 2017-04-02
`roundrobin` vs. `interleave_longest`


pylang:
Recently while looking at the recipes, I noticed  [`robinrobin`][1] gives similar results to [`interleave_longest`][2]:

```python
import more_itertools as mit

iterables = ['ABC', 'D', 'EF']

list(mit.roundrobin(*iterables))
# ['A', 'D', 'E', 'B', 'F', 'C']

list(mit.interleave_longest(*iterables))
# ['A', 'D', 'E', 'B', 'F', 'C']
```

I realize `interleave_longest` was discussed among other items in #22, but it seems it's similarity to an existing recipe may have been overlooked.  Is there a rationale for keeping both tools?  

  [1]: https://github.com/erikrose/more-itertools/blob/master/more_itertools/recipes.py#L256-L275
  [2]: https://github.com/erikrose/more-itertools/blob/master/more_itertools/more.py#L701-L711

bbayles:
Yes, I think they produce equivalent output - this was an oversight; if I'd realized it I would have made `interleave()` say "For a version that doesn't terminate on the shortest iterable, see `recipes.roundrobin()`" or something.

I might add a note to `interleave_longest()`'s docstring to say it's an alternate implementation of `roundrobin()`... I think it's a much better one.

pylang:
I agree `interleave_longest` is a more pythonic implementation than the `roundrobin` recipe.  A quick profile suggests it seems to be about 2x faster too.  A more thorough performance comparison may further justify keeping it.

bbayles:
I'll get 3a88f59523e0992320644ea8e26c1cbf11aae197 in the next release, which clarifies the situation.

bbayles:
Situation clarified in the docs; thanks @pylang .

PR #189, opened 2018-01-01
Improve asymptotic running time of interleave_longest

michael-celani:
`interleave_longest()` currently works by using `zip_longest()` with a `fillvalue` of `_marker`, and then filtering out all the `_marker` references in the returned iterable.

This has a running time of `O(len(iterables) * max(map(ilen,iterables))`. This is particularly inefficient if one iterable is much larger than any of others.

Using the proposed `OrderedDict` based implementation, the running time can be cut down to `O(sum(map(ilen, iterables)) + len(iterables))`.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -23,8 +23,10 @@ from six.moves import filter, map, range, zip, zip_longest
 
 from .recipes import consume, flatten, take
 
-if version_info > (3, 1, 0):
+try:
     from collections import OrderedDict
+except ImportError:
+    OrderedDict = None
 
 __all__ = [
     'adjacent',
@@ -832,7 +834,7 @@ def interleave_longest(*iterables):
 
 # If available, the OrderedDict implementation has better worst-case
 # runtime guarantees than the chain implementation
-if version_info > (3, 1, 0):
+if OrderedDict is not None:
     _interleave_longest_docstring = interleave_longest.__doc__
     interleave_longest = _interleave_longest_ordered
     interleave_longest.__doc__ = _interleave_longest_docstring
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -23,6 +23,9 @@ from six.moves import filter, map, range, zip, zip_longest
 
 from .recipes import consume, flatten, take
 
+if version_info > (3, 1, 0):
+    from collections import OrderedDict
+
 __all__ = [
     'adjacent',
     'always_iterable',
@@ -793,6 +796,28 @@ def interleave(*iterables):
     return chain.from_iterable(zip(*iterables))
 
 
+def _interleave_longest_ordered(*iterables):
+    """Helper for ``interleave_longest()``, called when the user is
+    using a version of Python with OrderedDict available.
+
+    This implementation of interleave_longest removes iterators from
+    consideration after they are exhausted instead of generating and
+    discarding instances of _marker.
+    """
+    iterables = OrderedDict(enumerate(map(iter, iterables)))
+    while iterables:
+        keys_to_clear = []
+
+        for key, iterable in iterables.items():
+            try:
+                yield next(iterable)
+            except StopIteration:
+                keys_to_clear.append(key)
+
+        for key in keys_to_clear:
+            del iterables[key]
+
+
 def interleave_longest(*iterables):
     """Return a new iterable yielding from each iterable in turn,
     skipping any that are exhausted.
@@ -803,6 +828,14 @@ def interleave_longest(*iterables):
     """
     i = chain.from_iterable(zip_longest(*iterables, fillvalue=_marker))
     return filter(lambda x: x is not _marker, i)
+
+
+# If available, the OrderedDict implementation has better worst-case
+# runtime guarantees than the chain implementation
+if version_info > (3, 1, 0):
+    _interleave_longest_docstring = interleave_longest.__doc__
+    interleave_longest = _interleave_longest_ordered
+    interleave_longest.__doc__ = _interleave_longest_docstring
 
 
 def collapse(iterable, base_type=None, levels=None):


pylang:
Interesting.  Did you happen to profile your implementation?  It would be informative to compare it against the present `more_itertools.interleave_longest` and `more_itertools.roundrobin` recipes.

michael-celani:
I didn't know about ``roundrobin``.

I ran a test with my first implementation as well as your implementation and roundrobin. It specifically targets the "difference in number of items case":
```python
import cProfile
from playground import *

for num_lists in range(5):
    print(f'==== # Iterables {num_lists} ====')
    for variance in (10, 20, 50, 100, 1000, 10000, 100000):
        print(f'==== Variance {variance} ====')
        iterables = [[] * num_lists, list(range(variance))]
        cProfile.run('for i in range(100): [*interleave_longest(*iterables)]')
        cProfile.run('for i in range(100): [*interleave_longest_ordered(*iterables)]')
        cProfile.run('for i in range(100): [*roundrobin(*iterables)]')
```

The output of that test is here:
https://pastebin.com/RwxHNg9r

It looks like ``next`` was a bottleneck, so in my implementation I mapped the keys to the ``__next__()`` function, but I still couldn't beat ``roundrobin``:
https://pastebin.com/6z7qqbd8

``roundrobin`` is simply a more efficient version of what I'm proposing with the same basic idea, so I'll drop my implementation for ``roundrobin``'s from now on. It's fairly clear that ``roundrobin`` is better in cases where one iterator is much bigger than the other, so let's look at more general cases.

I would assume in the real world that we would see a lot of iterators that are relatively similar in size, so I went for a general test that generates a lot of random list of lists with a normal distribution. It should be rare that lists are of wildly different sizes, but it is possible with this test.

Here's the new test:
```python
import cProfile
from playground import interleave_longest, roundrobin
from random import normalvariate, randint


for std_dev in (10, 25, 50, 100, 500, 1000, 2000, 5000):
    print(f"σ: {std_dev}")
    test_cases = list(list(range(int(normalvariate(10000, std_dev))) for _ in range(randint(2, 5))) for _ in range(1000))

    pr = cProfile.Profile()
    pr.enable()
    for test_case in test_cases:
        A = list(interleave_longest(*test_case))
    pr.disable()
    pr.print_stats()

    pr = cProfile.Profile()
    pr.enable()
    for test_case in test_cases:
        A = list(roundrobin(*test_case))
    pr.disable()
    pr.print_stats()
```

The results confirm what we'd expect:
https://pastebin.com/UwdJwuJ0

``interleave_longest`` is more susceptible to variance in the lengths of the input iterables, whereas ``roundrobin`` is more susceptible to variance in the amount of input iterables. You can see the 'waste' in ``interleave_longest`` by looking at the difference in calls between `<lambda>` and `roundrobin`.

It looks like the overhead of the calls is still too great for ``roundrobin`` to be better in all but very big cases. I think it would be worth updating the docstring to mention this.

michael-celani:
I was about to close this pull request but then I tried it in PyPy, where ``roundrobin`` is better in literally every way:
https://pastebin.com/09SNdHe8

I've changed my pull request to use ``roundrobin`` in PyPy.

pylang:
What a great analysis.  This would be a cool blog post if you were ever so motivated.  I appreciate the statistical approach to profiling.  It's the first time I've seen it done this way.  

It seems to counter an earlier observation (#122), in which `interleave_longest` appeared to be 2x faster.   However, that test was likely done with a less robust test using jupyter's `%timeit`.  Thanks for this insight.

more_itertools/more.py
```diff
@@ -805,6 +806,13 @@ def interleave_longest(*iterables):
     return filter(lambda x: x is not _marker, i)
 
 
+# The performance of roundrobin is far superior to interleave_longest in PyPy
+if python_implementation() == 'PyPy':
+    _interleave_longest_docstring = interleave_longest.__doc__
+    interleave_longest = roundrobin
+    interleave_longest.__doc__ = _interleave_longest_docstring
```
MSeifert04: Wouldn't that change `roundrobin`s `__doc__` because `interleave_longest = roundrobin` just creates an alias.

michael-celani:
Restarted my computer. I ran a slightly different test that doesn't discard the results of the lists and compares them for equality afterwards. Here's the test:

```python
import cProfile
from playground import interleave_longest, roundrobin
from random import normalvariate, randint


for std_dev in (10, 25, 50, 100, 500, 1000, 2000, 5000):
    print("σ:", std_dev)
    test_cases = list(list(range(int(normalvariate(10000, std_dev))) for _ in range(randint(2, 5))) for _ in range(1000))

    pr = cProfile.Profile()
    pr.enable()
    listA = [list(interleave_longest(*test_case)) for test_case in test_cases]
    pr.disable()
    pr.print_stats()

    pr = cProfile.Profile()
    pr.enable()
    listB = [list(roundrobin(*test_case)) for test_case in test_cases]
    pr.disable()
    pr.print_stats()

    print(listA == listB)
```

Results:
Cython 3.6.1 - https://pastebin.com/L3RAkNdC
PyPy3 3.5.3 - https://pastebin.com/LR9abz7A

Interestingly, in _this_ test, ``roundrobin`` trounces ``interleave_longest`` even in Cython.
Can I get some reproducability on this? I would like more data from other computers.

bbayles:
Hmm, I'm inclined to not merge this, given the discussion in issue #122. If `roundrobin` is better on CPython, PyPy, etc. on all sizes then we can drop the competing implementation. If `interleave_longest` has an edge in some cases, then let's leave it with a note that `roundrobin` exists.

bbayles:
(Also, excellent analysis, and I second the recommendation for writing this up in a blog post.)

bbayles:
Ah, I didn't realize that this topic was well-explored on [python-ideas](https://groups.google.com/forum/#!searchin/python-ideas/roundrobin|sort:date/python-ideas/gPdoHO86YC8/z2ULsMouAwAJ) and [bugs.python.org](https://bugs.python.org/issue32099).

I think we could improve the existing recipe's performance a bit by `return`-ing `filter(partial(is_not, _marker), i)`. However, my guess is that `roundrobin` is still going to win by virtue of dropping empty iterators.

Here's what I'd like to do:
* Make the code change below, aliasing `roundrobin` and fixing up the docstring.
* Separate the tests for `interleave` and `interleave_longest`
* Add a few more tests for both, including mixed types - strings, iterators, an infinite iterator, etc.

If this PR can evolve into that, I'll merge it. I'll celebrate the victory in the release notes before the next version.

---

```python
def interleave_longest(*iterables):
    """Return a new iterable yielding from each iterable in turn,
    skipping any that are exhausted.

        >>> list(interleave_longest([1, 2, 3], [4, 5], [6, 7, 8]))
        [1, 4, 6, 2, 5, 7, 3, 8]

    This function is an alias for :func:`roundrobin`.

    """
    # There was previously an alternate implementation here, but testing
    # showed that the existing roundrobin recipe is superior.


_interleave_longest_docstring = interleave_longest.__doc__
interleave_longest = partial(roundrobin)
interleave_longest.__doc__ = _interleave_longest_docstring
```
</details>

pylang:
+1.

I want to note that there is still hope for `interleave_longest`.   For instance, one may look at `roundrobin` and think, "why do we need `interleave_longest`  It's less performant ..."?   Another way to look at it might be "We have the legacy `roundrobin` recipe.  Let's leave that alone.  And even though `interleave_longest` is similar _now_, we can still extend it in the future to become something else."

For example, adding an optional `fillvalue` for shorter iterables:

```python
>>> list(interleave_longest([1, 2, 3], [4, 5], [6, 7, 8], fillvalue=0))
[1, 4, 6, 2, 5, 7, 3, 0, 8]
```

This is demonstrating we can gradually add features to this tool while leaving the classic recipe untouched.

michael-celani:
I think the reason `interleave_longest` performs so poorly in PyPy is because the [bottleneck is `filter()`](https://bitbucket.org/pypy/pypy/issues/2709/filter-way-slower-than-inlining-the-check). The equivalent code below performs much better in PyPy, but worse in Cython:

```python
def interleave_longest(*iterables):
    i = chain.from_iterable(zip_longest(*iterables, fillvalue=_marker))
    for val in (x for x in i if x is not _marker):
        yield val
```

Though reading the blog posts one of them mentioned `roundrobin` is quadratic in terms of the _number of iterators_. I took another look, and there it is - I didn't notice! I was more impressed by the fact that it was faster than my `OrderedDict` implementation and it appeared to be the same run time, but that's because it's hidden in the implementation of `cycle` and `islice`!  `roundrobin performs terribly in instances of a lot of small iterators:

For 1000 tests of 2000 single element lists:
https://pastebin.com/tJ1XretN

We clearly can't have something that crashes on something as small as 2000 iterators. My `OrderedDict` implementation to the rescue - it has `O(1)` remove, doesn't have the constant time overhead of re-adding the element like the `deque`-based implementations, and doesn't recurse on itself!

For 1000 tests of 2000 single element lists (of the implementations that aren't roundrobin):
https://pastebin.com/4m9N7AYX

 But it still performs worse than the current function in the average case...so it's a matter of tradeoffs. 

Here's what I found out about all the implementations there are:
- `interleave_longest()` is certainly the simplest to understand, but `filter()` is really damaging on PyPy and it suffers in cases where there's one iterator that's much larger than the others.
- `roundrobin()` is by far the best in cases where there's a small amount of iterators, because there's much less overhead in the iterator function approach compared to instantiating objects. However, it can't handle a large amount of iterators very well.
- Finally, `interleave_longest_ordered()`, my `OrderedDict` (really `OrderedSet`) based implementation scales well in both variables, but in smaller and average cases is dominated by the cost of the data structure. (Incidentally, it's comparable to `interleave_longest_deque()`, a `deque` based implementation that pops the element from the head and appends it to the tail if it's not exhausted. The `OrderedDict` implementation is better in Cython and worse in PyPy. It probably has to do with the internal implementation of both data structures.)

At runtime, the _only_ thing we know is the amount of iterables. My suggestion is as follows: based on the amount of iterables, dynamically choose between the current implementation and `roundrobin`. We should also remove the `filter()` call to not hamstring PyPy.

Also of note:
- Grabbing references to `iter(it).__next__` is apparently a performance enhancement over calling `next()`. Every single case where I made the optimization dropped the running time of my tests substantially.

bbayles:
More good analysis. I'll withdraw my earlier [request](https://github.com/erikrose/more-itertools/pull/189#issuecomment-354900787) to make `interleave_longest` an alias for `roundrobin`, since the former doesn't have the quadratic-in-number-of-iterables behavior.

New request:
* Change the `return` value of `interleave_longest` to `filter(partial(is_not, _marker), i)`, where `is_not` is imported from the `operator` module
* Change the note in the docstring for `roundrobin` to mention that it may be faster for small numbers of iterables, but `interleave_longest` should be preferred for larger numbers.
* Add a note in the docstring for `interleave_longest` to mention that `roundrobin` does the same thing, and may be faster for small numbers of iterables.
* Separate the tests for `interleave` and `interleave_longest` and add tests with mixed types etc.

---

Below is a collection of alternate implementations, if someone wants to test them:

<details><summary>filter with partial</summary>

```python
def interleave_longest_filter_partial(*iterables):
    func = partial(is_not, _marker)
    it = chain.from_iterable(zip_longest(*iterables, fillvalue=_marker))
    return filter(func, it)
```

</details>


<details><summary>queue with append</summary>

```python
def interleave_longest_queue_append(*iterables):
    if PY2:
        next_methods = (iter(it).next for it in iterables)
    else:
        next_methods = (iter(it).__next__ for it in iterables)

    q = deque(next_methods)
    popleft = q.popleft
    append = q.append

    while q:
        next_method = popleft()
        try:
            yield next_method()
        except StopIteration:
            pass
        else:
            append(next_method)
```

</details>


<details><summary>queue with rotate</summary>

```python
def interleave_longest_queue_rotate(*iterables):
    if PY2:
        next_methods = (iter(it).next for it in iterables)
    else:
        next_methods = (iter(it).__next__ for it in iterables)

    q = deque(next_methods)
    rotate = q.rotate
    popleft = q.popleft

    while q:
        try:
            while True:
                yield q[0]()
                rotate(-1)
        except StopIteration:
            popleft()

```

</details>

<details><summary>yield loop</summary>

```python
def interleave_longest_loop(*iterables):
    if PY2:
        next_methods = [iter(it).next for it in iterables]
    else:
        next_methods = [iter(it).__next__ for it in iterables]

    n = len(iterables)
    indexes = list(range(n))
    finished = [False] * n
    while n:
        for i in indexes:
            if finished[i]:
                continue
            try:
                yield next_methods[i]()
            except StopIteration:
                n -= 1
                finished[i] = True

```

</details>

bbayles:
Slight update - a comprehension seems faster on PyPy; I amended my request above. See also PyPy issue [2709](https://bitbucket.org/pypy/pypy/issues/2709/filter-way-slower-than-inlining-the-check), for why `filter` is so slow.

bbayles:
Had a few moments to put together PR #192, which adds some guidance on performance and works around PyPy's difficulty with `filter()`.

I'll close this PR after I merge that one, but if someone comes up with a solidly better implementation of `interleave_longest` I'm open to it. I'd love to link to a writeup of the analysis on this too (in the release notes).

michael-celani:
Sorry about not getting to this - I've been dealing with a lot of big changes I'm making in my life.

Feel free to link to the PR for a writeup.

bbayles:
No problem - good luck!

PR #190, opened 2018-01-01
Add reversed_extended()

michael-celani:
The builtin ``reversed()`` function returns a reverse iterator on iterables that implement ``__reversed__()`` or both ``__len__()`` and ``__getitem__()``, but won't work on other iterables, such as the results of ``map()``, file iterators, or arbitrary generator functions:

```python
>>> print(*map(lambda x: x * 2, reversed([1,2,3])))
6 4 2

# Changing the location of the reversed causes errors
>>> print(*reversed(map(lambda x: x * 2, [1,2,3]))) 
TypeError: 'map' object is not reversible
```

```python
>>> with open('/Users/mcelani/Desktop/my_file.txt', 'r') as file_iter:
...   print(*reversed(file_iter), sep='')
... 
  File "<stdin>", line 2, in <module>
TypeError: '_io.TextIOWrapper' object is not reversible

# Need to add extra 'list' cruft to get this to work
>>> with open('/Users/mcelani/Desktop/my_file.txt', 'r') as file_iter: 
...   print(*reversed(list(file_iter)), sep='')
... 
World
Hello
```

This pull request implements ``reversed_extended()``, which is a simple `try` / `catch` wrapper around ``reversed()`` that automatically handles non-reversible iterators by caching all the values and returning them in reverse order. Reversible iterables still leverage the return of ``reversed()``. 

You can import it as its own function or as a drop-in replacement for ``reversed()``:
```python
>>> from more_itertools import reversed_extended as reversed
>>> with open('/Users/mcelani/Desktop/my_file.txt', 'r') as file_iter:
...   print(*reversed(file_iter), sep='')
... 
World
Hello
```

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -214,7 +214,7 @@ Others
 **New itertools**
 
 .. autofunction:: numeric_range(start, stop, step)
-.. autofunction:: reversed_extended
+.. autofunction:: always_reversible
 .. autofunction:: side_effect
 .. autofunction:: iterate
 .. autofunction:: difference(iterable, func=operator.sub)
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -214,6 +214,7 @@ Others
 **New itertools**
 
 .. autofunction:: numeric_range(start, stop, step)
+.. autofunction:: reversed_extended
 .. autofunction:: side_effect
 .. autofunction:: iterate
 .. autofunction:: difference(iterable, func=operator.sub)
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -54,6 +54,7 @@ __all__ = [
     'one',
     'padded',
     'peekable',
+    'reversed_extended',
     'rstrip',
     'run_length',
     'seekable',
@@ -1633,6 +1634,24 @@ def islice_extended(iterable, *args):
 
             for item in cache[i::step]:
                 yield item
+
+
+def reversed_extended(iterable):
+    """An extension of :func:`reversed` that supports all iterables, not
+    just those which implement the ``Reversible`` or ``Sequence`` protocols.
+
+        >>> print(*reversed_extended(x for x in range(3)))
+        2 1 0
+
+    If the iterable is already reversible, this function returns the
+    result of :func:`reversed()`. If the iterable is not reversible,
+    this function will cache the remaining items in the iterable and
+    yield them in reverse order, which may require significant storage.
+    """
+    try:
+        return reversed(iterable)
+    except TypeError:
+        return reversed(deque(iterable))
 
 
 def consecutive_groups(iterable, ordering=lambda x: x):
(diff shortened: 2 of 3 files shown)


bbayles:
Oh, I like the idea. I'll review / merge when I get a chance - new baby at home!

michael-celani:
Congratulations!

more_itertools/more.py
```diff
@@ -1635,6 +1636,24 @@ def islice_extended(iterable, *args):
                 yield item
 
 
+def reversed_extended(iterable):
```
bbayles: What do you think about `always_reversible` instead of `reversed_extended`?

more_itertools/more.py
```diff
@@ -1635,6 +1636,24 @@ def islice_extended(iterable, *args):
...
+
+    If the iterable is already reversible, this function returns the
+    result of :func:`reversed()`. If the iterable is not reversible,
+    this function will cache the remaining items in the iterable and
+    yield them in reverse order, which may require significant storage.
+    """
+    try:
+        return reversed(iterable)
+    except TypeError:
+        return reversed(deque(iterable))
```
bbayles: I think I'd rather see `list` here instead of `deque`. [Both](https://github.com/python/cpython/blob/a5552f023e1d8cbafee1e51d316cc581deb2295f/Modules/_collectionsmodule.c#L1846) [have](https://github.com/python/cpython/blob/1fb72d2ad243c965d4432b4e93884064001a2607/Objects/listobject.c#L2923) special reverse iterator implementations; neither falls back to the "`__len__` and then `__getitem__` in reverse order" default method.

The `deque` would be faster if we were destructively popping through it, of course, but I think profiling should show `list` to be just as fast (and maybe slightly faster due to less accounting) for sequential access.

more_itertools/more.py
michael-celani: Probably better.

more_itertools/more.py
michael-celani: My tests show that there is no difference between the two in Cython, but a clear win for `list()` in PyPy. I'll change it to `list()`.

docs/api.rst
```diff
@@ -214,6 +214,7 @@ Others
 **New itertools**
 
 .. autofunction:: numeric_range(start, stop, step)
+.. autofunction:: reversed_extended
```
bbayles: Could you change this instance as well?

erikrose:
@bbayles Congrats on the little one! I can't believe you're still firing off reviews! Don't lose sleep on account of more-itertools; I just went through the new-baby time, and, for me anyway, sleep is the key to being a good father (and anything else, for that matter). Cheers!

bbayles:
Ah, open source is how I unwind! Little one is actually sleeping pretty well now, but the first week was as tough as everyone says.

PR #191, opened 2018-01-05
Add bitmap.encode and bitmap.decode

michael-celani:
``bool`` is rather expensive in Python. It's a subclass of ``int``, so they both end up taking at least 24 bytes:

```python
>>> sys.getsizeof(10)
24
>>> sys.getsizeof(True)
24
```

But we can store a lot more truth values in a single integer by leveraging a bitmap. This pull request implements a set of functions that compress booleans into an integer / length pair and recreates that stream of booleans from it.

```python
>>> uncompressed = [1, 0, 1, 1, 0]
>>> bitmap.encode(uncompressed)
(13, 5)

>>> compressed = (13, 5)
>>> list(bitmap.decode(compressed))
[True, False, True, True, False]
```

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -129,6 +129,7 @@ These tools return summarized or aggregated data from an iterable.
 .. autofunction:: consecutive_groups(iterable, ordering=lambda x: x)
 .. autofunction:: exactly_n(iterable, n, predicate=bool)
 .. autoclass:: run_length
+.. autoclass:: bitmap
 
 ----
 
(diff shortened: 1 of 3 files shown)


bbayles:
I think I'm going to reject this one - [bitarray](https://github.com/ilanschnell/bitarray/) does all this and more.

I have always found it a bit odd that out of everything in `binascii`, `struct`, and the `int` / `bytes` constructors you can't automate ints-from-bits. I suspect it's to force new programmers to learn bitshifting.

Keep the ideas coming, though!

PR #192, opened 2018-01-10
Additional tests and notes on performance for interleave_longest and roundrobin

bbayles:
Re: the discussion in PR #189, this PR implements [these](https://github.com/erikrose/more-itertools/pull/189#issuecomment-355298498) points to help clarify things w.r.t. `interleave_longest` and `roundrobin`.

* Change the `return` value of `interleave_longest` to `(x for x in it if x is not _marker)` - this helps with PyPy
* Change the note in the docstring for `roundrobin` to mention that it may be faster for small numbers of iterables, but `interleave_longest` should be preferred for larger numbers
* Add a note in the docstring for `interleave_longest` to mention that `roundrobin` does the same thing, and may be faster for small numbers of iterables.
* Separate the tests for `interleave` and `interleave_longest` and add tests with mixed types etc.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -803,7 +803,7 @@ def interleave_longest(*iterables):
 
     """
     i = chain.from_iterable(zip_longest(*iterables, fillvalue=_marker))
-    return filter(lambda x: x is not _marker, i)
+    return (x for x in i if x is not _marker)
 
 
 def collapse(iterable, base_type=None, levels=None):
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -801,6 +801,10 @@ def interleave_longest(*iterables):
         >>> list(interleave_longest([1, 2, 3], [4, 5], [6, 7, 8]))
         [1, 4, 6, 2, 5, 7, 3, 8]
 
+    This function produces the same output as :func:`roundrobin`, but may
+    perform better for some inputs (in particular when the number of iterables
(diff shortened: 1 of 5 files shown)


more_itertools/more.py
```diff
@@ -801,9 +801,13 @@ def interleave_longest(*iterables):
...
         [1, 4, 6, 2, 5, 7, 3, 8]
 
+    This function produces the same output as :func:`roundrobin`, but may
+    perform better for some inputs (in particular when the number of iterables
+    is large).
+
     """
     i = chain.from_iterable(zip_longest(*iterables, fillvalue=_marker))
-    return filter(lambda x: x is not _marker, i)
+    return (x for x in i if x is not _marker)
```
erikrose: Of course, this changes the result from a list to an iterable, but that’s how it’s specced, so it should be fine.

bbayles: FYI, the original didn't return a list due to [this import](https://github.com/erikrose/more-itertools/blob/821450b4d31979ae1f3bd88992e9401154260185/more_itertools/more.py#L22).

pylang: Thank you for clarifying.

erikrose: Aha. Tricky!

erikrose: That’s what I get for reviewing on a handheld. ;-)

PR #193, opened 2018-01-16
Add nth_combination from recipes

bbayles:
The itertools docs [now](https://github.com/python/cpython/pull/5161/files) have a new recipe; `nth_combination`. This PR adds it here, plus some tests to show that it works.


pylang:
Seeing a new itertools recipe added to python is like witnessing the birth of a star.

bbayles:
image

Issue #195, opened 2018-02-01
map/reduce recipe


pylang:
Would this [`map_reduce` recipe][0] by R. Hettinger be of interest?  Here is an example from the docstring:

    >>> def even_odd(elem):                     # sample mapper
    ...     if 10 <= elem <= 20:                # skip elems outside the range
    ...         key = elem % 2                  # group into evens and odds
    ...         return key, elem

    >>> map_reduce(range(30), even_odd)         # show group members
    {0: [10, 12, 14, 16, 18, 20], 1: [11, 13, 15, 17, 19]}

    >>> map_reduce(range(30), even_odd, sum)    # sum each group
    {0: 90, 1: 75}

  [0]: https://code.activestate.com/recipes/577676-dirt-simple-mapreduce/?in=user-178123

bbayles:
Hmm, yes. I like it - it's actually very similar to the `defaultdict(list)-maker` I played with in PR #188. The "reduce" part is an optional part added onto the end.

A simpler version would use `defaultdict(list`) and leave out the "skip where the `mapper` returns `None`" (that can be implemented in the reducer). This feels like a better inclusion than what I was groping toward before, so if you want to message Raymond H. and ask for his blessing, I'm good with it? (he previously sanctioned [this one](https://twitter.com/bbayles/status/865353484383776768)).

I'd be interested also if someone wants to explore something like this for `bucket` - at each stage it could yield the "reduction." Something like `bucket` + `accumulate`.

---

```python
def map_reduce(iterable, map_func, reduce_func=None):
    ret = defaultdict(list)
    for item in iterable:
        ret[mapper(item)].append(item)

    if reduce_func is not None:
        for k, v in ret.items():
            ret[k] = reduce_func(v)
    
    return ret

```

rhettinger:
Yes, go ahead.

pylang:
Many thanks @rhettinger.

@bbayles I like using `defaultdict`.  It's  clean, but I believe removing `if r is not None:`  will cause  `map_func`  to behave more like a `key_func`.   Any transformations from the mapping function would be added as a key rather than split between a key, value pair.

```python
def even_odd_transform(elem):
    key = elem % 2
    return key, 10*elem


def map_reduce_proposed(iterable, map_func, reduce_func=None):
    ret = ct.defaultdict(list)
    for item in iterable:
        ret[map_func(item)].append(item)

    if reduce_func is not None:
        for k, v in ret.items():
[...]

PR #196, opened 2018-02-10
Add map_reduce function

bbayles:
Re: issue #195 , this PR adds a variant of the "dirt simple map/reduce" [recipe](https://code.activestate.com/recipes/577676-dirt-simple-mapreduce/?in=user-178123). 

Per the discussion in the issue, I've modified things a bit:
* With the default arguments, this now scratches the itch I noted in PR #188: "easily make a `defaultdict(list)`"
* The original asks that you filter in the `mapper` function; here I encourage users to filter using standard itertools idioms (generator expressions or list comprehensions). This keeps the functions passed in simple, such that they just have to do one thing.

I think these modifications make the docstring easy to comprehend, which I usually see as a strong virtue:

image

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -129,6 +129,7 @@ These tools return summarized or aggregated data from an iterable.
 .. autofunction:: consecutive_groups(iterable, ordering=lambda x: x)
 .. autofunction:: exactly_n(iterable, n, predicate=bool)
 .. autoclass:: run_length
+.. autofunction:: map_reduce
 
 ----
 
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -2032,6 +2032,10 @@ def map_reduce(iterable, keyfunc, valuefunc=None, reducefunc=None):
     Note that all items in the iterable are gathered into a list before the
     summarization step, which may require significant storage.
 
+    The returned object is a :obj:`collections.defaultdict` with the
+    ``default_factory`` set to ``None``, such that if behaves as a normal
+    dictionary.
+
     """
     keyfunc = (lambda x: x) if (keyfunc is None) else keyfunc
     valuefunc = (lambda x: x) if (valuefunc is None) else valuefunc
@@ -2046,4 +2050,5 @@ def map_reduce(iterable, keyfunc, valuefunc=None, reducefunc=None):
         for key, value_list in ret.items():
             ret[key] = reducefunc(value_list)
 
+    ret.default_factory = None
     return ret
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -2037,7 +2037,6 @@ def map_reduce(iterable, keyfunc, valuefunc=None, reducefunc=None):
     dictionary.
 
     """
-    keyfunc = (lambda x: x) if (keyfunc is None) else keyfunc
     valuefunc = (lambda x: x) if (valuefunc is None) else valuefunc
 
     ret = defaultdict(list)
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -2033,7 +2033,7 @@ def map_reduce(iterable, keyfunc, valuefunc=None, reducefunc=None):
     summarization step, which may require significant storage.
 
     The returned object is a :obj:`collections.defaultdict` with the
-    ``default_factory`` set to ``None``, such that if behaves as a normal
+    ``default_factory`` set to ``None``, such that it behaves like a normal
     dictionary.
 
     """
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -51,6 +51,7 @@ __all__ = [
     'locate',
     'lstrip',
     'make_decorator',
+    'map_reduce',
     'numeric_range',
     'one',
     'padded',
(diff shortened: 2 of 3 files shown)


pylang:
Looks clean.  Here's an application on sample data extended from this [SO post][0]:

<details>
 <summary>Setup code</summary>

```
# data.txt
gender,rating,age_range
M,PG,young
M,R,adult
F,NR,adult
M,R,adult
F,PG13,young
F,PG,young
```

```python
import csv
import collections as ct


def read_file(fname):
    with open(fname, "r") as f:
        reader = csv.DictReader(f)
        for line in reader:
            yield line
```
</details>


```python
# Helper reduction function
def agg_counters(lst):
    """Return a dict of (column, Counter) key-value pairs."""
    agg = ct.defaultdict(ct.Counter)
    for d in lst:
        for k, v in d.items():
            agg[k][v] += 1
    return dict(agg)


def groupby(iterable, query):
    """Return a dict where the keys are items of the queried column."""
    kfunc = lambda d: d[query]
    vfunc = lambda d: {k: v for k, v in d.items() if k}
    rfunc = lambda lst: agg_counters(lst)
    return dict(map_reduce(iterable, kfunc, valuefunc=vfunc, reducefunc=rfunc))


filename = "data.txt"
iterable = [line for line in read_file(filename)]
groupby(iterable, "gender")
```

Ouput

```
{'F': {'age_range': Counter({'adult': 1, 'young': 2}),
  'gender': Counter({'F': 3}),
  'rating': Counter({'NR': 1, 'PG': 1, 'PG13': 1})},
 'M': {'age_range': Counter({'adult': 2, 'young': 1}),
  'gender': Counter({'M': 3}),
  'rating': Counter({'PG': 1, 'R': 2})}}
```


<details>
 <summary>Demo</summary>

```python
# Find the top age range among males
groupby(iterable, "gender")["M"]["age_range"].most_common(1)[0]
# ('adult', 2)

# Find total attendants per rating
{k: sum(v["gender"].values()) for k, v in groupby(iterable, "rating").items()}
# {'NR': 1, 'PG': 2, 'PG13': 1, 'R': 2}
```
</details>


  [0]: https://stackoverflow.com/questions/48680608/function-to-return-the-highest-count-value-using-a-rule/48683294?noredirect=1#comment84438514_48683294

thebigmunch:
Just a quick note/question on the return or, at least, the documentation of the return. It states it returns a dictionary, but it actually returns a defaultdict. This is an important distinction. In my use cases, I want a normal dict. @pylang's use above also wants a normal dict. Should the return be made a dict? If not, the docstring should be changed to note that it returns a defaultdict.

pylang:
>  @pylang's use above also wants a normal dict.

@thebigmunch just to clarify, I am interested in the functionality of a `defaultdict`, and so I do think a `defaultdict` should be returned to handle general cases.  I am using the `dict(dd)` trick here only to clean up the repr for this particular example.  Otherwise the output is a nesting of defaultdicts with unnecessary indentation.

thebigmunch:
@pylang Alright, then I think that means ``map_reduce`` should return a plain dict, and your code should explicitly convert that to defaultdict.

Semantically, this function should be returning a dict, not a defaultdict. Using a defaultdict is an implementation detail of this particular map reduce algorithm. The intent of a map reduce is provide an end product which, in this case, I believe is a plain dict. I believe the docstring as written backs that up.

I can live with converting to a plain dict every time I use this function, but then the docstring really needs to be changed. I was definitely not expecting to get a defaultdict when I first read the documentation.

bbayles:
It is a bit odd to get back a `defaultdict(list)` if you passed in a `reducefunc` argument, so I've set to `default_factory` to `None` such that the returned object behaves the same as a normal dict, and updated the docstring.

pylang:
+1 Yes I agree.  I was concerned about preserving the results of the `defaultdict`.  Converting the result to a dictionary via `dd.default_factory = None` is the proper way to do it, not `dict(dd)`.

pylang:
Yes I agree. I was concerned about preserving the results of the `defaultdict`. Converting the result to a `dictionary via dd.default_factory = None` is the proper way to do it. +1

erikrose:
Just a minor, drive-by spelling nit: `reducefunc` (etc.) would be more PEP-8-ish (and consistent with the rest of the lib) as `reduce_func`.

erikrose:
Btw, I grok this infinitely more easily than #188. :-)

bbayles:
Re: method names, I had them with underscores originally, but I think you'll find the other functions here don't (see groupby_transform, or for that matter the docs for itertools.groupy).

PEP 8 wants "underscores as necessary to improve readability" for method names; I think these are readable enough that consistency trumps.

pylang:
For reference, since this release there has been active discussion in Python-ideas about adding a grouping tool to the standard library.

Threads

- [Allow a group by operation for dict comprehension](https://mail.python.org/pipermail/python-ideas/2018-June/051719.html)
- [grouping / dict of lists](https://mail.python.org/pipermail/python-ideas/2018-July/051786.html)
- [Where should grouping() live (was: grouping / dict of lists)](https://mail.python.org/pipermail/python-ideas/2018-July/051813.html)
- [[Python-ideas] Where should grouping() live](https://mail.python.org/pipermail/python-ideas/2018-July/051826.html)

A [suggestion](https://mail.python.org/pipermail/python-ideas/2018-June/051749.html) was submitted for `more_itertools`.

PR #199, opened 2018-03-17
Add edit_distance

michael-celani:
Adds the `edit_distance()` function, which computes the Levenshtein distance of two sequences. The Levenshtein distance of two sequences is the minimum amount of insertions, deletions, and substitutions it would require to transform one iterable into another:

```python
>>> edit_distance([1, 2, 3], [1, 2, 4, 5, 3])
2
```

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -122,6 +122,7 @@ These tools return summarized or aggregated data from an iterable.
 **New itertools**
 
 .. autofunction:: ilen
+.. autofunction:: edit_distance
 .. autofunction:: first(iterable[, default])
 .. autofunction:: one
 .. autofunction:: unique_to_each
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -39,6 +39,7 @@ __all__ = [
     'distinct_permutations',
     'distribute',
     'divide',
+    'edit_distance',
     'exactly_n',
     'first',
     'groupby_transform',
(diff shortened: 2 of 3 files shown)


bbayles:
Many thanks for the PR. Lovely algorithm, not sure yet if it belongs here; I'll meditate on it.

pylang:
I'm curious @michael-celani why the name "edit_distance"?

Also there are a few other [iterable metric algorithms] (https://en.wikipedia.org/wiki/Levenshtein_distance#Relationship_with_other_edit_distance_metrics).  Should we settle for one, or include all I wonder?

bbayles:
I'm leaning toward declining this one, mainly since I'm not sure you'd reach for this library when there are [a bunch of others](https://pypi.python.org/pypi?%3Aaction=search&term=Levenshtein&submit=search) with more algorithm choices.

I'd reconsider if we had a nice itertools spin on it, such that it's a good demonstration of how to use `itertools.compress`, or something?

Here's a cute memoized version that uses `lru_cache` that I might include in a `more_functools` library:

<details><summary>Recursive / memoized version</summary>

```python
from functools import lru_cache


def lev_dist(s, t):
    @lru_cache(maxsize=None)
    def _lev_helper(len_s, len_t):
        if len_s == 0:
            return len_t

        if len_t == 0:
            return len_s

        cost = 0 if (s[len_s - 1] == t[len_t - 1]) else 1
        
        return min(
            _lev_helper(len_s - 1, len_t) + 1,
            _lev_helper(len_s, len_t - 1) + 1,
            _lev_helper(len_s - 1, len_t - 1) + cost,
        )
    
    s = tuple(s)
    t = tuple(t)
    ret = _lev_helper(len(s), len(t))
    return ret
```

</details>

bbayles:
Officially declining on this one, but will probably go for something similar that makes clever / illustrative use of one of the `itertools` primitives.

PR #200, opened 2018-03-26
make iLen concise

hfq-luojinwei:


MSeifert04:
I think this was changed to the longer variant for performance reasons.

MSeifert04:
See also https://github.com/erikrose/more-itertools/commit/5a2c25c16959fce1749a6aa1d9555e15c7ad9dd1

hfq-luojinwei:
@MSeifert04 I'm setting back the wheel of history >.<

MSeifert04:
Maybe it warrants an inline comment that deque is used on purpose for performance reasons. You probably won't be the last one that thinks it's unnecessarily "complicated". @bbayles ?

bbayles:
I'd be open to someone writing a nice explanatory comment. ~50% of the purpose of this library is to do things with the itertools / emulate itertools-y things and get people to look at how it works.

PR #206, opened 2018-05-07
more.py: document ilen

bmintz:
I found the implementation of ilen to be confusing, so I documented it.
It's still unclear to me why this is better than
```py
length = 0
for _ in iterable:
    length += 1
return length
```
Some basic testing shows that the `deque` method is a bit faster, so that could be why.


bbayles:
You're right - this was the reason. Many thanks!

PR #207, opened 2018-05-09
Distinct permutations issue

jferard:
Currently, `distinct_permutations` needs an iterable with a `len`. But that's not necessary since we have a `Counter` that contains (implicitely) the `len` of the iterable.


bbayles:
Thanks! My thought when making this originally is that you could just pass in `list(iterable)`, since there's no way you're going to be using this function with iterables that are very long. But this is a fine improvement.

PR #208, opened 2018-05-09
Change distinct_permutations implementation

jferard:
This PR is totally different from https://github.com/erikrose/more-itertools/pull/207: in https://github.com/erikrose/more-itertools/pull/207, I propose to fix an issue in `distinct_permutations`. Here, I propose another implementation of `distinct_permutations`. (I found the issue when making tests of the new implementation.)

The main and only reason is performance. I found an [interesting implementation of `distinct_permutations` on stackoverflow](https://stackoverflow.com/questions/6284396/permutations-with-unique-values/39014159#39014159) and after some refactoring, I made it faster, using generators.

Here's the full code of my benchmark:

```
import string
import random

def test_func(func, values):
    def test_func_closure():
        return list(func(values))
    return test_func_closure

times = {}
for _ in range(100):
    values = [random.choice(string.ascii_letters[:5]) for _ in range(10)]
    result = set(itertools.permutations(values))
    for func in [distinct_permutations_jferard, distinct_permutations_mi]:
        t = test_func(func, values)
        assert sorted(t()) == sorted(result), "expected: {}, actual: {}".format(result, t())
        times.setdefault(func, []).append(timeit.timeit(t, number=20))

for k,v in times.items():
    print (k, sum(v))
```

And here's the result:

```
<function distinct_permutations_jferard at 0x7f2eec375510> 46.91712202400777
<function distinct_permutations_mi at 0x7f2eec375c80> 248.31336366801224
```

Roughly five times faster.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -509,48 +509,41 @@ def one(iterable, too_short=None, too_long=None):
 
 def distinct_permutations(iterable):
     """Yield successive distinct permutations of the elements in *iterable*.
-
         >>> sorted(distinct_permutations([1, 0, 1]))
         [(0, 1, 1), (1, 0, 1), (1, 1, 0)]
-
     Equivalent to ``set(permutations(iterable))``, except duplicates are not
     generated and thrown away. For larger input sequences this is much more
     efficient.
-
     Duplicate permutations arise when there are duplicated elements in the
     input iterable. The number of items returned is
     `n! / (x_1! * x_2! * ... * x_n!)`, where `n` is the total number of
     items input, and each `x_i` is the count of a distinct item in the input
     sequence.
-
     """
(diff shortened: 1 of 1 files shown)


bbayles:
I merged your other PR - thanks for that! This one I'll need to study a bit.

Off the bat, I'm not sure about the re-use of the `permutations_generator` variable name so much. Could you clean that up such that it's a bit clearer? Also, the docstring change is angering the tests on Travis.

Also, would you mind reaching out to the author of the Stack Overflow post, so he can give his blessing and be credited?

jferard:
I made some formatting (comments) and renamed the helper function and a variable. Should be better.
I also asked LittleRoys on stackoverflow (see my  comment [here](https://stackoverflow.com/a/39014159/6914441)) He seems still active and I hope I'll get an asnwer.

Issue #205, opened 2018-04-23
"rindex"/ "rfind" tool


pylang:
From [rhettinger's challenge][3], should we consider adding a tool like  [`rindex`][1] or [`rfind`][0] to this library?  In particular, these tools find index for the last occurrence of an item.

**Sample Implementations**

[rhettinger's solution][5]
```python
def rindex(seq, x):
    return next(compress(count(len(s)-1, -1), (map(x.__eq__, reversed(s)))), -1)
```

more-itertools solution
```python
import more_itertools as mit


def rindex(iterable, item):
    """Return the highest index for the given item."""
    return next(mit.islice_extended(mit.locate(iterable, lambda x: x == item), -1, None))
```

**Uses**

- [How to find the last occurrence of an item in a Python list][7]
- [Equivelant to rindex for lists in Python [duplicate]][8]
- [Find index of last occurrence of a substring in a string][9]
- See N. Coglan's response to [this tweet][6]

  [0]: https://docs.python.org/2/library/string.html#string.rfind
  [1]: https://docs.python.org/2/library/string.html#string.rindex
  [3]: https://twitter.com/raymondh/status/987735142016864256
  [5]: https://twitter.com/raymondh/status/987835455046479872
  [6]: https://twitter.com/Persaik/status/987771879866892288
  [7]: https://stackoverflow.com/questions/6890170/how-to-find-the-last-occurrence-of-an-item-in-a-python-list
  [8]: https://stackoverflow.com/questions/9836425/equivelant-to-rindex-for-lists-in-python
  [9]: https://stackoverflow.com/questions/9572490/find-index-of-last-occurrence-of-a-substring-in-a-string?rq=1

bbayles:
I saw that Twitter message - nice solution!

I think the nice thing about `locate()` is that it returns all the indexes rather than just one (that was actually [your idea](https://github.com/erikrose/more-itertools/issues/139#issuecomment-302894121)!). So how about an `rlocate()`, which is basically a reversed `locate()`? I'd accept that PR.

This is close to what you have above, of course.

pylang:
Many thanks.  Hmm.  It's true `locate()` is a bit more general by returning all indicies.  It may even obviate any of these tools `r...` tools.  It presented the idea mostly for pondering.  

[...]

PR #212, opened 2018-05-24
Add rlocate function

bbayles:
Re: issue #205, this PR adds `rlocate()`, for finding indexes from the right-hand side.

The implementation is something like what I sketched [here](https://github.com/erikrose/more-itertools/issues/205#issuecomment-384143997) and what @jferard wrote [here](https://github.com/erikrose/more-itertools/issues/205#issuecomment-388319648).

My version doesn't have the "tell us if you have an expensive predicate" option; it seemed discromulent to me.

### DIFF
--- a/more_itertools/tests/test_more.py
+++ b/more_itertools/tests/test_more.py
@@ -1846,3 +1846,34 @@ class MapReduceTests(TestCase):
         d = mi.map_reduce([1, 0, 2, 0, 1, 0], bool)
         self.assertEqual(d, {False: [0, 0, 0], True: [1, 2, 1]})
         self.assertRaises(KeyError, lambda: d[None].append(1))
+
+
+class RlocateTests(TestCase):
+    def test_default_pred(self):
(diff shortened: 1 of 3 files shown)


more_itertools/tests/test_more.py
```diff
@@ -1846,3 +1846,34 @@ def test_ret(self):
...
+
+    def test_custom_pred(self):
+        iterable = ['0', 1, 1, '0', 1, '0', '0']
+        pred = lambda x: x == '0'
+        for it in (iterable[:], iter(iterable)):
+            actual = list(mi.rlocate(it, pred))
+            expected = [6, 5, 3, 0]
+            self.assertEqual(actual, expected)
+
+    def test_efficient_reversal(self):
```
bbayles: This test shows why the [first method](https://github.com/erikrose/more-itertools/pull/212/files#diff-a7f69885eda5d871c519f326bf410e62R2097) is useful; this wouldn't return in a timely manner when searching from the left.

PR #213, opened 2018-06-04
Implement last()

tmshn:
Implement `last()` method, which gets last element of the iterable.

Works same as `first(tail(iterable, n=1)[, default])`, but

- More concise and descriptive
- No iteration for random-accessible container

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -124,6 +124,7 @@ These tools return summarized or aggregated data from an iterable.
 
 .. autofunction:: ilen
 .. autofunction:: first(iterable[, default])
+.. autofunction:: last(iterable[, default])
 .. autofunction:: one
 .. autofunction:: unique_to_each
 .. autofunction:: locate(iterable, pred=bool)
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -150,11 +150,14 @@ def last(iterable, default=_marker):
     If *default* is not provided and there are no items in the iterable,
     raise ``ValueError``.
     """
-    if not isinstance(iterable, (list, deque)):
-        iterable = deque(iterable, maxlen=1)
     try:
-        return iterable[-1]
-    except IndexError:
+        try:
+            # Try to access the last item directly
+            return iterable[-1]
+        except (TypeError, AttributeError, KeyError):
+            # If not slice-able, iterate entirely using length-1 deque
+            return deque(iterable, maxlen=1)[0]
+    except IndexError:  # If the iterable was empty
         if default is _marker:
             raise ValueError('last() was called on an empty iterable, and no '
                              'default value was provided.')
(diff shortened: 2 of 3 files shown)


pylang:
An interesting approach.  Is there a rationale for using a deque over traditional unpacking? e.g.

    _*, last = iterable

MSeifert04:
@pylang  You have a typo there, shouldn't it be `*_, last = ...`? I guess `deque` might be faster though.

tmshn:
> shouldn't it be *_, last = ...? I guess deque might be faster though.

In addition to the speed, `*_, last` doesn't work for Python 2.

tmshn:
> I'm not really sure why you don' use tail

I simply feel `first(tail(1, iterable))` (or `next(tail(1, iterable))`) is ugly, and `last(iterable)` is much elegant.

> almost none of the functions here actually special case for random-access

That's true except for `sliced` vs `chunked`. But IMO, it's better to think about performance where it's easy.

bbayles:
I think a `last()` function is a fine addition. These implementations are a bit better than [earlier attempts](https://github.com/erikrose/more-itertools/issues/), too!

I think I'd rather `try/except` returning `iterable[-1]` for the sliceable case, then fall to `tail(iterable,1)[0]`. This instead of the type checking for `list` and `deque`.

tmshn:
Ah, `try/except` instead of type check sounds good idea. I'll give it a try.

bbayles:
Here's my best stab at a body for the function.

```python
def last(iterable, default=_marker):
    # If the iterable is slice-able, grab the last item.
    try:
        return iterable[-1]
    except IndexError:
        pass
    # Otherwise, feed it thorugh a length-1 deque to keep the last item.
    except TypeError:
        try:
            return deque(iterable, maxlen=1)[-1]
        except IndexError:
            pass

    # If we encountered an IndexError above, then the iterable was empty.
    if default is _marker:
        raise ValueError(
            'last() was called on an empty iterable, and no default value '
            'was provided.'
        )
    return default
```

Your tests and docstring are perfect, so I'll merge this PR once we've got the `try/except` implementation. Excellent!

tmshn:
Fixed to use `try/except`.
I actually did not adopt the code @bbayles wrote directly; the differences are

- nested `try/except` to make the logic more clear; outer one is for empty iterable, and inner for non slice-able
- catch `AttributeError` for user-defined classes which does not inherit `object` and `KeyError` for dict-like classes
    - (a dict having `-1` as key does not behaves as intended, but IMO that's a nit picky corner case that we don't need to care about

tmshn:
I mistakenly closed the pr, now re-opened 😆

bbayles:
This looks great - one more request. Could you add some test cases that exercise your `except AttributeError` and `except KeyError` additions?

tmshn:
Added 👉 b6a6b98

bbayles:
Great work - this will be in the next release.

Issue #214, opened 2018-06-11
Reference table to docs


pylang:
It is feasible to get a reference table like the following in the docs:

![2018-06-11 16_39_32-aiostream aiostream documentation](https://user-images.githubusercontent.com/10778668/41255978-415a105c-6d96-11e8-9e18-f39bcbc4b283.png)

The tools are presented in a table by category.  Clicking each tool takes you to the api docs.   

We've had a similar discussion in the past (#130) that led to categories in the docs (https://github.com/erikrose/more-itertools/pull/133, thanks @bbayles).   I think a quick table on the landing page would improve:

- finding a specific tool: now I click on `API Reference`, then <kbd>Ctrl</kbd> + <kbd>F</kbd> and search by name to find details on a certain tool
- seeing all tools pertaining to a category at a glance: now we click on a category and scroll through the tools until we hit the next category.

FYI: the example above is from [`aiostream`](https://aiostream.readthedocs.io/en/latest/) (think `asyncio` + `itertools`).   I've also seen a table-like listing of tools in @MSeifert04's [`iteration_utilities`](https://iteration-utilities.readthedocs.io/en/latest/). Perhaps he can comment on ease of maintainability.

MSeifert04:
For the main overview tables in `iteration_utilities` I use a [helper](https://github.com/MSeifert04/iteration_utilities/blob/master/helper/create_functions_table.py) based on AstroPys Table that creates that RST table. One just has to remember updating it.

If you just want a `aiostream`-like table that should be relatively easy, because the order of the cells doesn't change it's not too complicated to maintain it manually. If you need help with the initial creation I can probably help there. Automation could also be possible but in that case you probably need some marker (private property, doc parsing or something like that) to group them by.

bbayles:
Thanks for the note. I'm positive on the idea; if someone makes a PR for it I will trigger a ReadTheDocs build, and if it looks good will merge it.

PR #215, opened 2018-06-13
Readme function overview table

MSeifert04:
xref #214 

It's probably not yet perfect but it contains all functions in the API Documentation.

Before I start on the ReadTheDocs RST documentation I wanted to wait for some feedback.

cc @pylang


MSeifert04:
You can visit the branch in case you want to know how it would look like: https://github.com/MSeifert04/more-itertools/tree/table

pylang:
@MSeifert04  That looks great!  Thank you.  It looks like it's as simple as copying the link in the table section.  I went through then links and they direct to correct tool.

MSeifert04:
@pylang Thanks for checking the links! 👍 

I'm not sure about the position in the ReadMe, maybe they should be after the "api reference" mention. I also didn't check if it works with "code-formatting" the links. I may be able to check that tomorrow.

Since I didn't created that table by hand: Would it make sense to put the doc-scraping and rst-formatting code in the repo?

bbayles:
This looks good - thanks!

Re: placement, I might play with it later, but I think that's a good place to put it to start.

Re: "code formatting," I don't think it's necessary, and might looks a little weird.

Re: generation code, perhaps you could post it here?

One nitpick, which might be a pain to do - I put the categories on the left side of [the docs](https://more-itertools.readthedocs.io/en/latest/api.html) in that order deliberately. I think I'd like to have the categories in that same order in the table?

MSeifert04:
Okay, I removed the sorting.

The script is basically this (not particularly elegant and probably easy to break):

```
from collections import defaultdict
from astropy.io.ascii import RST

path_to_more_itertools = # insert your path to the more-itertools repo here

groups = defaultdict(list)
current_group = None
last_line = None
with open(path_to_more_itertools + 'docs/api.rst') as file:
    for line in file:
        if line.strip().startswith("===") and last_line is not None:
            current_group = last_line.strip()
        if line.strip().startswith(".. auto"):
            function = line.strip().split('::')[1].split('(')[0].strip()
            link = f'`{function} <https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.{function}>`_'
            groups[current_group].append(link)
        last_line = line
        
del groups["API Reference"]  # remove unnecessary API Reference group...
    
r = []
for group_name, functions in groups.items():
    if not functions:  # remove empty groups
        del groups[group_name]
        continue
    r.append((group_name, ', '.join(sorted(functions))))
    
from astropy.table import Table
print('\n'.join(RST().write(Table(rows=r))[2:]))  # convert to Table and throw away the first two header lines...
```

bbayles:
I found [this](https://rackerlabs.github.io/docs-rackspace/tools/rtd-tables.html), which seems to fit the bill.

I also changed around the table format to make it more easily hand-editable. See [here](https://raw.githubusercontent.com/erikrose/more-itertools/bbayles-tables/README.rst). What do you think of that?

MSeifert04:
Yeah, I have known about the line-wrapping problem in RTD. I just haven't thought that you actually include the README in the docs.

I like the alternative design. 👍 Did you generate that based on a script or by hand? I know my script is fragile but at least (if it works) it can do all automatically without any manual intervention - so there might be a reason to actually do it by script - or not, if you don't expect it to be a problem.

It's still a bit annoying if one adds a function with a longer name but that's far less likely than adding another function to a group.

bbayles:
The table was semi-scripted. My intention is to only update it at release time. I may regret this format if I introduce a new category, but we'll see how this works. Thanks for the contribution!

pylang:
@bbayles I went through the links in the talbe again.  Looks like `last()` and `rlocate()` don't point to any tool.  I think it's bc the links point to the *stable* version of RTD instead of *latest* version.  That means, new tools won't have links until the next release.

PR #216, opened 2018-06-16
Remove some "else after return" constructs

rec:
This is really picayune :-) but I read over your excellent code several times and it nagged at me that you _nearly always_ didn't have an `else:` after a return but not _always_ so I sent this out.

I'd feel no remorse if you just dropped it entirely :-D 

Otherwise, I'm not quite sure how to run the tests - since pytest and nose seem to already have their own versions of more-itertools installed!

Either way, thanks for writing an excellent library!

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -594,17 +594,17 @@ def intersperse(e, iterable, n=1):
     """
     if n == 0:
         raise ValueError('n must be > 0')
-    elif n == 1:
+    if n == 1:
         # interleave(repeat(e), iterable) -> e, x_0, e, e, x_1, e, x_2...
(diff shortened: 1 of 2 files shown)


rec:
Heh, you can see how to run the tests [here](https://github.com/erikrose/more-itertools/blob/master/tox.ini#L5).

bbayles:
I think I'll decline these - [this one](https://github.com/erikrose/more-itertools/pull/216/files#diff-a7f69885eda5d871c519f326bf410e62L742) I might get rid of, but the rest were deliberate!

Nonetheless, I appreciate your desire to contribute, and hope you will again!

P.S. You can [do what Travis does](https://github.com/erikrose/more-itertools/blob/master/.travis.yml#L19) to test as well.

rec:
Thanks, have a good one!

Issue #218, opened 2018-06-20
docs/_static/ not present in PyPI tarballs


sbraz:
Hi,
building the docs from the PyPI release result in the following warning:
```
copying static files... WARNING: html_static_path entry '/var/tmp/portage/dev-python/more-itertools-4.2.0/work/more-itertools-4.2.0/docs/_static' does not exist                                                  
```
You probablby need to add this to MANIFEST.in

bbayles:
Ah, you're right - the `_static` directory is not present in the most recent release, but it it there in `master.

sbraz:
Oh so it wasn't a problem of it not being included; it simply wasn't present in v4.2.0. I guess the next release will be fine.

bbayles:
That should do it - thanks!

PR #219, opened 2018-06-21
Build docs in tests

bbayles:
This PR should make Travis build docs, failing if a change introduces an error or a warning.


bbayles:
This should fix issue #218.

Issue #220, opened 2018-07-03
locate_seq, replace, and replace_seq


bbayles:
I may get to vetting these properly myself, but in case anybody watching is interested in taking that up: I pushed [a branch](https://github.com/erikrose/more-itertools/compare/replace?expand=1) with three new functions:
* `locate_seq`: Like `locate`, but for a sequence of items.
* `replace`: Like `str.replace`, but replacing single items in iterables.
* `replace_seq`: Like `replace`, but for sequences of items.

If these were proposed by someone else I think I would merge them, since they extend built-in functionality to iterables and proved to be tricky to get right.

However, these are questions I'd like to answer:
* Is there any actual need? How many StackOverflow questions are there around each? comp.lang.python threads?
* Is `replace` needed if `replace_seq` is there?
* What edge cases should be added in tests?
* Are there better / faster implementations? Especially for `replace_seq`?

pylang:
Cool.  My knee-jerk reactions are:

1. `locate_seq` -1 (name)
1. `replace` +1 (utility)
1. `replace_seq` +0 (uniqueness)

Details

1. By the name alone, I assumed this was a `locate` tool for Python sequences, collections that can be sliced.  `seq` doesn't have to be a sequence though.  An iterator can be used too, so the name throws me a bit.  You mean a "sequence" of adjacent items, or what I'd consider a window, but I'm not sure what is a better name.  Last bit, it's not quite like `locate` since it does not accept a predicate, so I think `locate_...` might be confusing.
1. I've been itching for a tool like this for ages.  This is great.
1.  If my goal was to "replace" items in an iterable, I'd probably reach for `replace` and see what I could with it first before seeking any other tool.  I'd prefer to see this feature rolled into a single `replace` function, but  I know that would a beast of a function to implement. 

Those are quick thoughts.  I'll have to take more time to look at them closely.  Thanks for the contrib!

bbayles:
I thought of a way to merge `locate` and `locate_seq`, which I think is probably the way to go with that. It doesn't map exactly on to `rlocate`, but can probably be made to work. I've pushed that branch.

What about the name `splice` instead of `replace_seq`?

pylang:
`locate` (merger) +1
- [Python: Find a substring in a string and returning the index of the substring](https://stackoverflow.com/questions/21842885/python-find-a-substring-in-a-string-and-returning-the-index-of-the-substring)
[...]

PR #221, opened 2018-07-06
Ctrl+H

bbayles:
Re: issue #220, this PR:
* Adds `replace()`, which can cut out sequences of items and splice others in their place
* Adds a *window_size* argument to `locate()` and `rlocate()` that enable them to search for multiple items instead of single items

These additions are intended to bring the functionality of `str.index()` and `str.replace()` to generic iterables.

---

I originally wrote the issue with three (or maybe four) new tools, but I've cut that down to one by unifying `replace()` and `replace_seq()` and adding sub-sequence support to the existing `locate()` and `rlocate()`.

I'll leave some additional comments in the diff.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -2188,12 +2188,11 @@ def replace(iterable, pred, substitutes, count=None, window_size=1):
     # iterable
     it = chain(iterable, [_marker] * (window_size - 1))
     windows = windowed(it, window_size)
-    w = []
 
     n = 0
     for w in windows:
         # If the current window matches our predicate (and we haven't hit
-        # out maximum number of replacements), splice in the substitutes
+        # our maximum number of replacements), splice in the substitutes
         # and then consume the following windows that overlap with this one.
         # For example, if the iterable is (0, 1, 2, 3, 4...)
         # and the window size is 2, we have (0, 1), (1, 2), (2, 3)...
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -1488,10 +1488,6 @@ def locate(iterable, pred=bool, window_size=None):
         >>> list(locate(iterable, pred=pred, window_size=3))
         [1, 5, 9]
 
-    If the *window_size* is larger than *iterable*, *iterable* will be padded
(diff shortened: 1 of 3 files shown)


more_itertools/more.py
```diff
@@ -2121,8 +2137,76 @@ def rlocate(iterable, pred=bool):
...
+        >>> list(replace(iterable, pred, substitutes, count=2))
+        [1, 1, None, 1, 1, None, 1, 1]
+
+    Use *window_size* to control the number of items passed as arguments to
+    *pred*. This allows for locating and replacing subsequences.
+
+        >>> iterable = [0, 1, 2, 5, 0, 1, 2, 5]
+        >>> window_size = 3
+        >>> pred = lambda *args: args == (0, 1, 2)  # 3 items passed to pred
+        >>> substitutes = [3, 4] # Splice in these items
```
bbayles: In these examples I've attempted to show that you can locate however many items you want with `pred` and replace them with some arbitrary list of other items. There's no relationship between the length of the sub-sequence that gets replaced and the replacements.

more_itertools/more.py
```diff
@@ -2121,8 +2137,76 @@ def rlocate(iterable, pred=bool):
...
+        >>> list(replace(iterable, pred, substitutes))
+        [1, 1, 2, 3, 1, 1, 2, 3, 1, 1]
+
+    If *count* is given, the number of replacements will be limited:
+
+        >>> iterable = [1, 1, 0, 1, 1, 0, 1, 1]
+        >>> pred = lambda x: x == 0
+        >>> substitutes = [None]
+        >>> list(replace(iterable, pred, substitutes, count=2))
+        [1, 1, None, 1, 1, None, 1, 1]
```
pylang: In this example, although count is "limited" to 2, **all** of the zeros are reolaced.  And doesn't capture the intent of `count` to limit replacements.  Consider `count=1` -> `[1, 1, None, 1, 1, 0, 1, 1]`

bbayles: Good point.

Issue #227, opened 2018-08-04
Subsequences


pylang:
A [subsequence](https://en.wikipedia.org/wiki/Subsequence) is a subset of a sequence, ordered but not necessarily consecutive/contiguous.

They show up in mathematics, computer science, linguistics, and natural sciences.   There is interest in them:

- [Longest Increasing Subsequence](https://stackoverflow.com/questions/2631726/how-to-determine-the-longest-increasing-subsequence-using-dynamic-programming) and [here](https://stackoverflow.com/questions/3992697/longest-increasing-subsequence)
- [Longest Repeated Substring](https://stackoverflow.com/questions/11090289/find-longest-repetitive-sequence-in-a-string) and [here](https://codereview.stackexchange.com/questions/63329/finding-the-largest-repeating-substring)
- [Longest Palindromic Subsequence](https://codereview.stackexchange.com/questions/161775/length-of-longest-palindrome-subsequence)

They are part of classic algorithmic problems:

- [Longest Common Subsequence Problem](https://en.wikipedia.org/wiki/Longest_common_subsequence_problem)
- [Longest Common Substring Problem](https://en.wikipedia.org/wiki/Longest_common_substring_problem)

Q: What are thoughts on adding a tool that simply collects subsequences?

# Examples

Haskell offers a [`subsequences` tool](https://hackage.haskell.org/package/base-4.11.1.0/docs/Data-List.html#v:subsequences):

**Haskell**

```python
>>> subsequences "abc"
["","a","b","ab","c","ac","bc","abc"]
```

It turns out our [`powerset` recipe](https://more-itertools.readthedocs.io/en/stable/_modules/more_itertools/recipes.html#powerset) is akin to this tool:

**Python**

```python
 >>> list(powerset([1,2,3]))
 [(), (1,), (2,), (3,), (1, 2), (1, 3), (2, 3), (1, 2, 3)]
```

**Limitations**

These tools only pertain to generic subsequences alone with limited options. 

- While iterating, the items cannot be altered (no `pred`)
- On "longest xyz problems," the longest items are at the end of the iterator (no `reverse`)
- Some problems require a "substring" of contiguous elements, i.e. `"ac"` and `(1, 3)` are invalid

# Implementation

**Proposal**

A possible implementation to handle the latter limitations might be the following:

```python
def subsequences(iterable, pred=None, reverse=False, contiguous=False):
    """Return windows of all subsequences sized from `len(iterable)` to 1.

    Examples
    --------
[...]

PR #229, opened 2018-08-07
Add substrings()

bbayles:
Re: #227, I did like the "generate all windows" thing, so here's a thing to do that.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -55,6 +55,7 @@ These tools yield windows of items from an iterable.
 **New itertools**
 
 .. autofunction:: windowed
+.. autofunction:: subsequences
 .. autofunction:: stagger
 
 ----
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -55,7 +55,7 @@ These tools yield windows of items from an iterable.
 **New itertools**
 
 .. autofunction:: windowed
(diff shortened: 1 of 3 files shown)


pylang:
Neat implementation.  

So since these items are contiguous elements, the formal name would be a "substring".  A terribly confusing name imo, but here is the distinction:

- [substring][str]: contiguous elements `"abc" -> "" "a" "b" "c" "ab" "bc" "abc"` (similar to `windowed`)
- [subsequence][seq]: contiguous/non-contiguous `"abc" -> "" "a" "b" "c" "ab" "bc" "ac" "abc"` (similar to `powerset`)

Notice the [haskell function][hask] includes the non-contiguous element `"ac"`.   If it was called "subsequence", then it should include non-contigous elements too.  I'm not a huge fan of the name "substring" though as it can be confused with the `str` type.  So I'm not sure what to call this particular standalone tool.

Is the motivation to make more efficient windowed?


  [str]: https://en.wikipedia.org/wiki/Substring
  [seq]: https://en.wikipedia.org/wiki/Subsequence
  [hask]: https://hackage.haskell.org/package/base-4.11.1.0/docs/Data-List.html#v:subsequences

PR #236, opened 2018-11-02
clarify ilen

bmintz:
This uses the same technique, but should be easier to follow.
It is around the same speed as the old approach, in my testing.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -440,10 +440,10 @@ def ilen(iterable):
     This consumes the iterable, so handle with care.
 
     """
-    len = 0
-    for len, _ in enumerate(iterable, 1):
-        pass
-    return len
+    length = 0
+    for _ in iterable:
+        length += 1
+    return length
 
 
 def iterate(func, start):
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -440,10 +440,10 @@ def ilen(iterable):
     This consumes the iterable, so handle with care.
 
     """
-    length = 0
-    for _ in iterable:
-        length += 1
-    return length
+    len = 0
+    for len, _ in enumerate(iterable, 1):
+        pass
+    return len
 
 
 def iterate(func, start):
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -440,10 +440,10 @@ def ilen(iterable):
     This consumes the iterable, so handle with care.
 
     """
-    len = 0
-    for len, _ in enumerate(iterable, 1):
+    length = 0
+    for length, _ in enumerate(iterable, 1):
         pass
-    return len
+    return length
 
 
 def iterate(func, start):
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -440,12 +440,10 @@ def ilen(iterable):
     This consumes the iterable, so handle with care.
 
     """
-    # maxlen=1 only stores the last item in the deque
-    d = deque(enumerate(iterable, 1), maxlen=1)
-    # since we started enumerate at 1,
-    # the first item of the last pair will be the length of the iterable
-    # (assuming there were items)
-    return d[0][0] if d else 0
+    len = 0
+    for len, _ in enumerate(iterable, 1):
+        pass
+    return len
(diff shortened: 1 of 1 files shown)


MSeifert04:
> It is around the same speed as the old approach, in my testing

Could you share your timings? Because it is significantly slower  in the cases i tested.

MSeifert04:
Maybe I was a bit too fast. I made some more timings and wanted to share them:

```
%matplotlib notebook

from simple_benchmark import benchmark # from https://github.com/MSeifert04/simple_benchmark
from collections import deque

def ilen_deque(xs):
    d = deque(enumerate(xs, 1), maxlen=1)
    return d[0][0] if d else 0

def ilen_loop(xs):
    len = 0
    for len, _ in enumerate(xs, 1):
        pass
    return len

def ilen_deque_it(xs):
    return ilen_deque((i for i in xs))

def ilen_loop_it(xs):
    return ilen_loop((i for i in xs))

b = benchmark([ilen_deque, ilen_loop, ilen_deque_it, ilen_loop_it],
              {2**i: [0]*(2**i) for i in range(1, 20)},
              'length')
b.plot()
```

png image 785 x 528 pixels

```
b.plot(relative_to=ilen_loop)
```
png image 800 x 429 pixels

Based on Python 3.7 (I haven't done any timings on other Python versions - but probably should be done)

bmintz:
Am I reading the graphs correctly? Seems that ilen_loop is fastest.

bbayles:
Nice work. I'll merge this if the items `flake8` flags in the CI runs are fixed?

bbayles:
Oh, and perhaps you could change the loop variable to something other than `len`? I'm allergic to shadowing built-ins.

bmintz:
lol i don't highlight builtins intentionally because `id` is such a great name. 
but ok, i will change it to length.
Also, I have a faster approach.

bmintz:
Please make sure this is indeed faster than the enumerate approach.

bbayles:
FYI, the CI tests are still failing on this:

```
./more_itertools/more.py:906:13: W504 line break after binary operator
./more_itertools/more.py:907:13: W504 line break after binary operator
```

bmintz:
It's not my code though?

bbayles:
Oh, perhaps flake8 updated? Thanks for pointing that out.

MSeifert04:
Just FYI if I compare it to [`iteration_utilities.count_items`](https://iteration-utilities.readthedocs.io/en/latest/generated/count_items.html#iteration_utilities.count_items) (I hope you forgive me the advertising):

png image 573 x 429 pixels

MSeifert04:
But again: this should also be tested on other computers (and probably on other Python versions).

MSeifert04:
You could also try `returns sum(1 for _ in xs)`. Maybe that is faster.

bmintz:
@MSeifert04 can you try it with your benchmarking tool?
I don't think it'll be faster though, since it has to step into and out of a generator.

MSeifert04:
> @MSeifert04 can you try it with your benchmarking tool?
>I don't think it'll be faster though, since it has to step into and out of a generator.
Yes, that's what I thought too. And I checked it, it's slower (roughly equal to the `+= 1` approach).

bbayles:
Many thanks for the analysis here.

achampion:
So using the benchmark reference above using `zip()` and `itertools.count()` looks like:

``` python
def ilen_zip(iterable):
    counter = it.count()
    deque(zip(iterable, counter), maxlen=0)
    return next(counter)
```

image

MSeifert04:
What Python version?

Btw I lile the style of the plot which one is this? Ggplot?

achampion:
Python 3.7.1 but it was also faster in Py3.6 (see https://github.com/erikrose/more-itertools/issues/230).  
The style is `ggplot`:

    from matplotlib import pyplot as plt
    plt.style.use('ggplot')

MSeifert04:
Ah thank you.

I don't know what's different on my installation but probably the performance of `ilen` should be tested on a broader set of computers and python versions.

PR #240, opened 2018-11-22
implement unzip (fixes #239)

### LINKED ISSUE #239: add unzip
```py
>>> g0, g1 = unzip(zip(range(10), range(1, 11)))
>>> next(g0)
0
>>> next(g1)
1
>>> next(g1)
2
>>> list(g0)
[1, 2, 3, 4, 5, 6, 7, 8, 9]
>>> list(g1)
[3, 4, 5, 6, 7, 8, 9, 10]
```

I tried to implement this but it turns out to be a lot harder than I thought. I managed to get a version working which steps all the generators forward when any generator is advanced, but I don't know how to do one that makes them operate independently.

I guess it could be done using a `more_itertools.peek`, where each gen returned by unzip keeps track of how many times it has been advanced.

Then when any gen is advanced past the farthest point in the original zipped iterable, the original zipped iterable is advanced again.

bmintz:


bbayles:
Beautiful implementation.

If you'll add some tests with edge cases (length 0 iterable; length 1 iterable, not-properly-zipped iterable) and fix the flake8 error, I'll merge this.

bmintz:
Test cases added.
"Improperly zipped" can be interpreted two ways:
* (1,), (1, 2), (1, 2, 3)  (increasing length)
* (1, 2, 3), (1, 2), (1,) (decreasing length)

In the first case, the function is documented to only return one iterable.
In the second case, I had to modify itemgetter to raise StopIteration when the tuple index is out of range. This short circuits each unzipped iterable, so that, in `xs, ys, zs = unzip(it)`:
- `ilen(xs) == 3`
- `ilen(xs) == 2`
- `ilen(ys) == 1`

This seems like a reasonable thing to do, to me.

bmintz:
Oh, and I wasn't sure if raising StopIteration like that was OK, because I still can't grok [PEP 479](https://www.python.org/dev/peps/pep-0479/), so I tested it with `from __future__ import generator_stop` in test_more.py, and there were no errors :+1:.

bbayles:
Hmm, I think the extra handling for "improperly zipped" iterables. I think I'd rather it just raise `IndexError` rather than try to soldier on, and we can document that this is what will happen? My request for tests was to make sure we know what happens so it can be documented/.

bmintz:
> I think I'd rather it just raise `IndexError` rather than try to soldier on

I raised StopIteration for precisely that reason: we want the returned iterable to be cut short if the underlying zipped iterable is. If we let the IndexError propagate, then it will also propagate to a call to e.g. `list()`.

```py
def f(x):
    if x == 10:
        raise IndexError
    return x

>>> list(map(f, range(20)))
  File "<stdin>", line 3, in f
IndexError
```
^ The above error is confusing and counter-productive. If you remove the `try:... except IndexError: raise StopIteration` from my pull request, then you can get this same error via:

```py
>>> improperly_zipped = iter([(1, 2, 3), (1, 2), (1,)])
>>> xs, ys, zs = unzip(improperly_zipped)
>>> list(xs)
[1, 1, 1]
>>> list(ys)
IndexError
>>> list(zs)
IndexError
```

Compare this with:

```py
def f(x):
    if x == 10:
        raise StopIteration
    return x

>>> list(map(f, range(20)))
[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
```

bbayles:
Hm, yeah. My feeling is that passing an iterable where all the items aren't the same length is an exceptional condition, and should raise an exception.

If we feel `IndexError` isn't helpful as-is, what if we caught and re-raised it with a different message?

```python
try:
    ...
except IndexError as e:
    e.args = 'unzip requires all elements to be the same length'
```

bmintz:
That would be a more helpful error message, for sure. However, it would not occur when the user calls unzip, only after, when the user tries to use the returned iterables from unzip. If we wanted to raise IndexError('unzip requires all elements to be the same length'), in my opinion, that should only be done if we can raise that IndexError as soon as the user calls unzip, but this exceptional condition is hard to catch when the user calls unzip, since we only look at the first element of the underlying iterable.

My reasoning for cutting the iterables short is based on a theoretical zip that exhibits this "decreasing length tuple" behavior:

```py
def zip(*iterables):
    iterators = list(map(iter, iterables))
    while True:
        to_yield = build_zipped_tuple(iterators)
        if not to_yield:
            return
        yield to_yield

def _build_zipped_tuple(iterables):
    vals = []
    for iterable in iterables:
        try:
            vals.append(next(iterable))
        except StopIteration as exc:
            break
    return tuple(vals)

for xs in zip((1, 2, 3), (1, 2), (1,)):
    print(xs)
```

This outputs:
```py
(1, 1, 1)
(2, 2)
(1,)
```

Maybe I can't see the forest for the trees here.

bmintz:
The zip I defined above is basically `itertools.zip_longest`, but without a fill value.

bbayles:
I've decided you're right; we should do whatever `zip(*zipped))` does.

bbayles:
I'll add this to the API docs and may change the docstring a bit. Thanks for this; excellent idea and a strong implementation.

PR #241, opened 2018-11-25
gitignore: ignore more venvs

bmintz:
not everyone uses env/ as their venv folder.
Also, I used venv/ instead of /venv/ because some may want a separate venv for docs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,7 +26,10 @@ pip-log.txt
 _build
 
 # Environment
-.env
+.env/
+.venv/
+env/
+venv/
 
 # IDE files
 .idea


.gitignore
```diff
@@ -26,9 +26,12 @@ pip-log.txt
...
-.env
+.env/
+.venv/
+env/
+venv/
 
 # IDE files
 .idea
 .vscode
-.DS_Store
```
bbayles: Any idea why this is being highlighted as changed?

bmintz: Ah, I figured it out. Somehow a stray `\r` character got in there, and I removed it.

PR #248, opened 2018-12-12
Fix typos

hugovk:


bbayles:
The tales of your deeds here today will live in the changelog forever.

Thanks!

Issue #246, opened 2018-12-10
Function to split iterable into uneven sized lists


rovyko:
What are your thought on adding a function that splits an iterable into different sized chunks?

I've been using more-itertools a lot lately, and combined windowed with accumulate in a one liner to group an iterable's items based on each group's size. I expanded it into a generator for cleanliness and better argument handling, and thought it might be a useful addition to the library.

```
from itertools import islice


def segmented(iterable, sizes):
    """Break *iterable* sequentially into lists of varying sizes, with a list
    of length 'n' returned for each integer 'n' in *sizes*:

        >>> list(segmented([1,2,3,4,5,6], [1,2,3]))
        [[1], [2, 3], [4, 5, 6]]

    If the sum of *sizes* is smaller than the length of *iterable*, then the
    remaining items of *iterable* will not be returned.

        >>> list(segmented([1,2,3,4,5,6], [2,3]))
        [[1, 2], [3, 4, 5]]

    If the sum of *sizes* is larger than the length of *iterable*, fewer items
    will be returned in the iteration that overruns *iterable* and further
    lists will be empty:

        >>> list(segmented([1,2,3,4], [1,2,3,4]))
        [[1], [2, 3], [4], []]

    When a None object is encountered in *sizes*, the returned list will
    contain items up to the end of *iterable* the same way that itertools.slice
    does:

        >>> list(segmented([1,2,3,4,5,6,7,8,9,0], [2,3,None]))
        [[1, 2], [3, 4, 5], [6, 7, 8, 9, 0]]

    :func:`segmented` can be useful for grouping a series of items where the
    sizes of the groups are not uniform. An example would be where in a row
    from a table, multiple columns represent elements of the same feature
    (e.g. a point represented by x,y,z) but, the format is not the same for
    all columns.
    """
	
	current_index = 0
    for size in sizes:
        if isinstance(size, int):
            current_index += size
            yield list(islice(iterable, current_index-size, current_index))
        elif size is None:
            yield list(islice(iterable, current_index, None))
            return
        else:
            raise TypeError("each item in 'sizes' must be an integer or None")
```

bbayles:
First, thanks for this! An excellent write-up.

Could you share your `windowed`/`accumulate` one-liner for comparison?

I'd also be interested to see an example with some "real" data that shows how you might select what to put in `sizes`.

Lastly, naming is always the hardest part - can we find any similar StackOverflow posts that have language of what people might search for?

Thanks again!

[...]

PR #249, opened 2018-12-13
Add split_into

rovyko:
Re: [Issue #246](https://github.com/erikrose/more-itertools/issues/246)

An addition to the `grouping` tools that splits an iterable into uneven groups based on a list of sizes. Given `split_into(iterable, sizes)`, a list of length _n_ is yielded containing the next _n_ items in `iterable` for each integer _n_ in `sizes`. 

When a `None` item is encountered in `sizes`, then the rest of `iterable` is returned. This follows from the functionality of `islice` which will yield until the end of the iterator if a `None` is passed to `end`.

`split_into` can take an iterator argument for both `iterable` and `sizes`. `iterable` will be consumed up to the sum of the integers encountered in `sizes` or will be fully consumed if a `None` value in `sizes` is encountered. `sizes` will be consumed up to and including any `None` values. A `TypeError` is thrown if an item in `sizes` is neither an instance of `int` or `None` unless the item comes after a `None` value.

All implementations, tests and references to `split_into` were inserted after `split_after` as it was the last item in the `grouping` section.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -74,6 +74,7 @@ __all__ = [
     'split_at',
     'split_after',
     'split_before',
+    'split_into',
     'spy',
     'stagger',
     'strip',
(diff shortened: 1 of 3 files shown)


more_itertools/more.py
```diff
@@ -1070,6 +1071,53 @@ def split_after(iterable, pred):
...
+    from a table, multiple columns represent elements of the same feature
+    (e.g. a point represented by x,y,z) but, the format is not the same for
+    all columns.
+    """
+    # convert the iterable argument into an iterator so its contents can
+    # be consumed by islice in case it is a generator
+    it = iter(iterable)
+
+    for size in sizes:
+        if isinstance(size, int):
```
bbayles: Could you make this:

```python
if size is None:
    yield list(islice(it, size))
else:
    yield list(it)
```

and let `islice` raise the `TypeError`? You can catch it and change the message if you like; I just don't want the `isinstance(..., int)` check.

PR #257, opened 2019-01-03
adds iter_side_effect function

JohnFirth:
I thought I'd submit this in case you thought it would be a worthwhile addition. My use case is POSTing enumerated chunks of an unchunked stream, which is something that using `side_effect` with `chunk_size` doesn't permit (as far as I can tell). If you think it's a worthwhile addition, I'll add tests.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -990,6 +990,47 @@ def side_effect(func, iterable, chunk_size=None, before=None, after=None):
             after()
 
 
+def iter_side_effect(func, iterable):
+    """Invoke *func* on the items of *iterable* as a whole. Yield the
+    original iterable's items.
+
+    Internally, *func* is invoked on an iterator produced by
+    :func:`itertools.tee`, such that the caveats applying to the consumption of
+    iterators produced by :func:`itertools.tee` apply to *func*. It's intended,
+    but not required, that *func* not exhaust the iterator.
+
+    Writing enumerated chunks of items to a file-like object:
+
+        >>> from io import StringIO
+        >>> f = StringIO()
+        >>> func = lambda it: map(lambda x: f.write(f'{x}\n'), enumerate(chunked(it, 2)))
+        >>> list(iter_side_effect(func, iter('abcde')))
+        ['a', 'b', 'c', 'd', 'e']
+        >>> f.seek(0); f.readlines()
+        ["(0, ['a', 'b'])\n", "(1, ['c', 'd'])\n", "(2, ['e'])\n"]
+        >>> f.close()
+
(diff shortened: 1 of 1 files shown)


bbayles:
Thanks! Could you tell me why it's important for this function to yield the original items from the iterable, rather than the `enumerate`-ed ones? I'm afraid I don't get that part.

JohnFirth:
You're welcome! It's useful in similar situations to those where the existing `side_effect` function is useful, that is where you want to apply to the elements of an iterable an impure function that doesn't itself return the original elements. For both `side_effect` and `iter_side_effect`, this is not so useful  when the impure function operates on individual elements (as far as I can tell), because you could use `map` instead, and return the item from the impure function after it's been used to generate the side effect. They're more useful for operations on chunks of items, because the caller doesn't have to deal with flattening the chunks afterwards. `side_effect`, though, doesn't allow operations on the iterable as a whole, only on at most individual chunks of the iterable. For me then, with my enumerated chunks, something like `iter_side_effect` is useful. 

It's the difference between allowing me to treat the operations particular to the POSTing totally separately from the main 'flow' in which the items are processed (they're additionally logged and posted elsewhere):

`    
    def post_iter_fn(iter):
        transformed_iter = map(transform, iter)
        chunks = chunked(transformed_iter, post_chunk_size)
        json_chunks = map(make_post_json, enumerate(chunks))
        return side_effect(post_to_client_url, json_chunks)

    # main flow
    posting_iter = side_effect_iter(post_iter_fn, iter)
`

and mixing them into the main flow, along with the unpacking and flattening operations, which is noisier:

`
    def post_chunk_fn(chunk):
        transformed_chunk = map(transform, chunk)
        json_chunk = make_address_post_json(transformed_chunk)
        return post_to_client_url(json_chunk)
   
    # main flow
    chunks = chunked(iter, post_chunk_size)
    posted_enumerated_chunks = side_effect(post_chunk_fn, enumerate(chunks))
    posted_chunks = map(itemgetter(1), posted_enumerated_chunks)
    posted_iter = flatten(posted_chunks)
`

bbayles:
Thanks for the details - this is a nice writeup.

I'm afraid I'm not totally sold on the inclusion here. Perhaps if you could write out how somebody who didn't know how `map`, the `itertools` module functions, or the `more_itertools` functions would write this? That might make the comparison easier.

Or, a link to a StackOverflow post or two where this would solve somebody's problem - that's always helpful.

JohnFirth:
You're welcome. I'm not totally sold on the inclusion either -- it is useful only for a relatively uncommon scenario -- so I'll leave it there. Thanks for your time :)

Issue #262, opened 2019-01-07
Add type annotations


Tadaboody:
I'd like to implement [PEP 484](https://www.python.org/dev/peps/pep-0484/) type annotations and distribute them as detailed in [PEP 561](https://www.python.org/dev/peps/pep-0561/). Do you think this is a good idea for the project?

Tadaboody:
If it is, should we make a separate type stub like the ones in [typeshed](https://github.com/python/typeshed) or implement them inside the existing source?

pylang:
I think it's a great idea.  Type annotations could introduce new tools (e.g. https://github.com/erikrose/more-itertools/issues/250).  Adding stubs would be more favorable than annotating the library directly.  Be aware, there has been some talk about add types in the past (https://github.com/erikrose/more-itertools/issues/231), but I'm not sure about the status.

Tadaboody:
I'll open a pull request, adding them iteratively over time. If @sk- is still interested I'll give him write access to that branch

bbayles:
Thanks for looking into this. I like the PEP 561 method while Python 2 support is on offer.

The PR looks like a good start. I'd like there to be tests that run in Travis that checks the types, though?

Tadaboody:
Of course testing is in order. I think running a type checker like mypy on the existing suite will both check the annotations and help fix bugs that the tests might have

Tadaboody:
Do you support splitting up the existing files like I started doing in the PR?

bbayles:
> Do you support splitting up the existing files like I started doing in the PR?

Ah, no - not really. That grouping is for the docs, no need to reproduce it elsewhere.

sk-:
@Tadaboody Feel free to add the annotations we are using so far:

```python
from typing import overload, Callable, Dict, Hashable, Iterable, List, Optional, TypeVar

_T = TypeVar('_T')
_U = TypeVar('_U')
_V = TypeVar('_V')
_H = TypeVar('_H', bound=Hashable)


def chunked(iterable: Iterable[_T], n: int) -> Iterable[List[_T]]: ...


@overload
def map_reduce(iterable: Iterable[_T], keyfunc: Callable[[_T], _H]) -> Dict[_H, List[_T]]: ...
@overload
def map_reduce(iterable: Iterable[_T], keyfunc: Callable[[_T], _H], valuefunc: Callable[[_T], _U]) -> Dict[_H, List[_U]]: ...
@overload
[...]

Issue #253, opened 2018-12-26
Drop Python 2.7 support by 2020


bbayles:
It's almost 2019, which will be the last year with any releases that explicitly support Python 2.7. See [here](https://python3statement.org/) for some information on motivation.

What this means:
* I'll continue to accept additions that require Python 2-specific code throughout 2019
* At the end of 2019 I'll make a final Python 2.7-compatible release
* After that release I will make issues for making use of Python 3-only features, like `yield from`
* I'll remove `six` from the requirements and remove the places where it's used
* I'll accept PRs that change existing tools to use Python 3-only features
* I'll accept PRs that add new tools with Python 3-only features

If someone wants to make a fork targeting older Python versions they are of course welcome to; if requested I will provide a link in the docs.

bbayles:
If now is good enough for numpy, it's good enough for me. I'll open a PR to drop Python 2-isms - let me know if you see any I missed.

Anybody have any opinions on using `yield from` instead of `return`, for tools that give back an iterator? For example, [this function](https://github.com/erikrose/more-itertools/blob/c5088d2f2af42b5ed5429c4d92cc5a83b7ab25b2/more_itertools/more.py#L2275) returns, but could `yield from` instead.

jaraco:
I would prefer return where suitable _unless_ you do in fact want the deferral of behavior preceding the yield from.

erikrose:
I generally reach for ‘yield from’ only when it can eliminate a ’for’ loop.

hugovk:
🎆https://github.com/erikrose/more-itertools/pull/268 has been merged.

Should this issue be closed, or kept open for visibility?

bbayles:
Ah, yes. I'll close this when I merge #271 and make a release.

bbayles:
And it's out. Many thanks!

PR #264, opened 2019-01-10
Use explicit keyword argument in some functions

Tadaboody:
Part of #253
Makes writing stubs for #262 much easier for some methods
Not py2 (and probably some of py3) compatible


bbayles:
This will be superseded by #268.

PR #268, opened 2019-01-25
Drop Python 2.7 support

bbayles:
Re: issue #253, this PR removes Python 2.7 support and drops several Python 2-isms. Tell me if you see more!


jaraco:
One crucial piece missing is the python_requires declaration in setup.py. You’ll want to include that to prevent installers from getting 6.0 in python 2 environments. Also, be sure to release with late versions of setuptools and twine. Let me know if you want more detailed advice or help.

bbayles:
Good catch on `python_requires`, glad to have _the_ expert on this paying attention.

MSeifert04:
One more optional thing: classes don't need to inherit from object anymore.

bbayles:
@MSeifert04, good spot.

@hugovk, that's a nice tool. Looks like each of its categories are satisfied here.

hugovk:
The tool also drops the `u''` prefix from strings in `conf.py`.

bbayles:
Merging this. This isn't the 6.0.0 release yet, but it will be close to it.

PR #270, opened 2019-02-03
Tips on unique_everseen:

bbayles:
Re: https://github.com/erikrose/more-itertools/issues/269 - @jferard , what do you think of this?

### DIFF
--- a/more_itertools/recipes.py
+++ b/more_itertools/recipes.py
@@ -355,6 +355,27 @@ def unique_everseen(iterable, key=None):
     Sequences with a mix of hashable and unhashable items can be used.
     The function will be slower (i.e., `O(n^2)`) for unhashable items.
 
+    Remember that ``list`` objects are unhashable - you may want to convert
+    an iterable of ``list`` objects to an iterable of ``tuple`` objects
+    to avoid slowdown. S
+
+        >>> slow_iterable = [1, 2], [2, 3], [1, 2]
+        >>> list(unique_everseen(slow_iterable))
+        [[1, 2], [2, 3]]
+        >>> fast_iterable = (tuple(x) for x in slow_iterable)
+        >>> list(unique_everseen(fast_iterable))
+        [(1, 2), (2, 3)]
+
+    Similary, you may want to convert unhashable ``dict`` objects using
+    the :func:`dict.items` method.
+
+        >>> slow_iterable = {'a': 'b'}, {'c': 'd', 'e': 'f'}, {'a': 'b'}
+        >>> list(unique_everseen(slow_iterable))
+        [{'a': 'b'}, {'c': 'd', 'e': 'f'}]
+        >>> fast_iterable = (tuple(x.items()) for x in slow_iterable)
+        >>> list(unique_everseen(fast_iterable))
+        [(('a', 'b'),), (('c', 'd'), ('e', 'f'))]
+
     """
     seenset = set()
     seenset_add = seenset.add


more_itertools/recipes.py
```diff
@@ -355,6 +355,27 @@ def unique_everseen(iterable, key=None):
     Sequences with a mix of hashable and unhashable items can be used.
     The function will be slower (i.e., `O(n^2)`) for unhashable items.
 
+    Remember that ``list`` objects are unhashable - you may want to convert
+    an iterable of ``list`` objects to an iterable of ``tuple`` objects
+    to avoid slowdown. S
```
MSeifert04: What does the trailing `S` mean?

more_itertools/recipes.py
```diff
@@ -355,6 +355,27 @@ def unique_everseen(iterable, key=None):
...
+        >>> list(unique_everseen(fast_iterable))
+        [(1, 2), (2, 3)]
+
+    Similary, you may want to convert unhashable ``dict`` objects using
+    the :func:`dict.items` method.
+
+        >>> slow_iterable = {'a': 'b'}, {'c': 'd', 'e': 'f'}, {'a': 'b'}
+        >>> list(unique_everseen(slow_iterable))
+        [{'a': 'b'}, {'c': 'd', 'e': 'f'}]
+        >>> fast_iterable = (tuple(sorted(x.items())) for x in slow_iterable)
```
MSeifert04: Note that `sorted` really relies on the values to be comparable which may not be the case. In some cases it might even be the case that the objects don't define a total ordering which may lead to unexpected results.

more_itertools/recipes.py
```diff
@@ -355,6 +355,27 @@ def unique_everseen(iterable, key=None):
     Sequences with a mix of hashable and unhashable items can be used.
     The function will be slower (i.e., `O(n^2)`) for unhashable items.
 
+    Remember that ``list`` objects are unhashable - you may want to convert
+    an iterable of ``list`` objects to an iterable of ``tuple`` objects
+    to avoid slowdown. S
+
+        >>> slow_iterable = [1, 2], [2, 3], [1, 2]
```
MSeifert04: It always confuses me if a tuple is created without the parenthesis. I generally only use this when `return`ing.

Not sure if this is something that has to be changed but I thought it could be helpful to mention it in case others think the same way.

more_itertools/recipes.py
```diff
@@ -355,6 +355,27 @@ def unique_everseen(iterable, key=None):
...
+        [(1, 2), (2, 3)]
+
+    Similary, you may want to convert unhashable ``dict`` objects using
+    the :func:`dict.items` method.
+
+        >>> slow_iterable = {'a': 'b'}, {'c': 'd', 'e': 'f'}, {'a': 'b'}
+        >>> list(unique_everseen(slow_iterable))
+        [{'a': 'b'}, {'c': 'd', 'e': 'f'}]
+        >>> fast_iterable = (tuple(sorted(x.items())) for x in slow_iterable)
+        >>> list(unique_everseen(fast_iterable))
```
MSeifert04: Would probably be "nicer" if the output is converted to dictionaries again so the result looks like the result above.

MSeifert04:
Just a general comment: The unique_everseen supports the `key` argument that may be more useful here instead of "converting" the inputs:

```python
list(unique_everseen(tuple(item) for item in fast_iterable))  # Replace this
list(unique_everseen(fast_iterable, key=tuple))   # with this
```

jferard:
It's correct (you have a typo before the first colon, "S"), but it's not what I had in mind. In your example, the output values are different (because the input value are different). I wanted to stress that, with a clever use of the `key` parameter (`tuple` or `frozenset` are the more obvious examples), you can speed up the output of the *same result* :

    >>> from more_itertools import unique_everseen
    >>> slow_iterable = [1, 2], [2, 3], [1, 2]
    >>> list(unique_everseen(slow_iterable)) # base case, O(n^2)
    [[1, 2], [2, 3]]
    >>> list(unique_everseen(map(tuple, slow_iterable))) # O(n) but a different output 
    [(1, 2), (2, 3)] # tuples instead of lists
    >>> list(unique_everseen(slow_iterable, key=tuple)) # O(n) with the same output.
    [[1, 2], [2, 3]] # lists as expected

Under the hood:

* In the first case, we test if the **list** was already seen (O(n^2)) and if not, we output the **list**
* In the second case, we test if the **tuple** (hashable) was already seen (O(n)) and if not, we output the **tuple**
* In the third case, we test if the **tuple** (hashable) was already seen (O(n)) and if not, we output the **list**

Benchmark:

    >>> from timeit import timeit
    >>> timeit(lambda: list(unique_everseen(slow_iterable)))
    1.4413324829999965
    >>> timeit(lambda: list(unique_everseen(map(tuple, slow_iterable))))
    0.9294636969998464
    >>> timeit(lambda: list(unique_everseen(slow_iterable, tuple)))
    0.8547339190001821

That's not spectacular, because the list has 3 elements and `seen` 2 elements, but it will be really faster in this case:

    >>> timeit(lambda: list(unique_everseen([x] for x in range(5000))), number=25)
    4.667081057001269
    >>> timeit(lambda: list(unique_everseen(([x] for x in range(5000)), tuple)), number=25)
    0.03208817000086128

more_itertools/recipes.py
bbayles: Just a typo.

bbayles:
Great comments. I forgot about the `key` parameter, since it's not in the recipes docstring.

jferard:
It's okay now! Maybe add unhashable ``set``/hashable ``frozenset`` option, but the idea is clear enough.

MSeifert04:
I'd say `key=frozenset(d.items())` would be a better suggestion than `key=tuple(sorted(d.items()))`. It's shorter, works for more kinds of inputs and I think it should give the correct results. It doesn't work if the values contain some unhashable values but that's another story 😄 

Otherwise 👍

bbayles:
I had `sorted` in for the docstring - older Pythons have the unstable dict ordering that make doctests not work. But `frozenset(d.items())` is indeed better for dicts.

PR #276, opened 2019-03-23
Add partitions function

rominf:
Add partitions function
    
Partitioning set or interval is a known thing in mathematics. This PR adds a function to do find all partitions of a collection. For more information what is considered a partition, see docstring.

Thanks, Tigran Saluev (http://saluev.com/) for the implementation.

UPD: the code was rewritten from scratch.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -19,7 +19,7 @@ from operator import itemgetter, lt, gt, sub
 from sys import maxsize, version_info
 from collections.abc import Sequence
 
-from .recipes import consume, flatten, take
+from .recipes import consume, flatten, powerset, take
 
 __all__ = [
     'adjacent',
@@ -54,6 +54,7 @@ __all__ = [
     'numeric_range',
     'one',
     'padded',
+    'partitions',
     'peekable',
     'replace',
     'rlocate',
@@ -2306,3 +2307,27 @@ def replace(iterable, pred, substitutes, count=None, window_size=1):
         # yield the first item from the window.
         if w and (w[0] is not _marker):
             yield w[0]
+
+
+def partitions(iterable):
+    """Yield all possible partitions of *iterable*.
+    Partition x is defined so that:
+    1. list(x) = [x_0, ..., x_n]
+    2. x_0 + ... + x_n = list(iterable)
+    3. all(isinstance(x_i, list) for x_i in x) == True
+    4. 1 <= n <= ilen(iterable)
+    5. isinstance(x, list) == True
+
+        >>> iterable = 'abcd'
+        >>> list(partitions(iterable))
+        [[['a', 'b', 'c', 'd']], [['a'], ['b', 'c', 'd']], \
+[['a', 'b'], ['c', 'd']], [['a', 'b', 'c'], ['d']], \
+[['a'], ['b'], ['c', 'd']], [['a'], ['b', 'c'], ['d']], \
+[['a', 'b'], ['c'], ['d']], [['a'], ['b'], ['c'], ['d']]]
+
+    """
+    sequence = list(iterable)
+    n = len(sequence)
+    for partition_indexes in powerset(range(1, n)):
+        yield [sequence[i:j] for i, j in zip((0, ) + partition_indexes,
+                                             partition_indexes + (n, ))]
--- a/more_itertools/tests/test_more.py
+++ b/more_itertools/tests/test_more.py
@@ -2311,3 +2311,26 @@ class ReplaceTests(TestCase):
         actual = list(mi.replace(iterable, pred, substitutes))
         expected = ['_', '_', 1, '_', '_', 3, '_', '_']
         self.assertEqual(actual, expected)
+
+
+class PartitionsTest(TestCase):
+    def test_basic(self):
+        iterable = 'abcd'
+        actual = list(mi.partitions(iterable))
+        expected = [
+            [['a', 'b', 'c', 'd']],
+            [['a'], ['b', 'c', 'd']],
+            [['a', 'b'], ['c', 'd']],
+            [['a', 'b', 'c'], ['d']],
+            [['a'], ['b'], ['c', 'd']],
+            [['a'], ['b', 'c'], ['d']],
+            [['a', 'b'], ['c'], ['d']],
+            [['a'], ['b'], ['c'], ['d']]
+        ]
+        self.assertEqual(actual, expected)
+
+    def test_empty(self):
+        iterable = []
+        actual = list(mi.partitions(iterable))
+        expected = [[[]]]
+        self.assertEqual(actual, expected)


rominf:
I've tried to implement it using `split_into`:
```python
def partitions_2(iterable):
    lst = list(iterable)
    n = len(lst)
    if n == 0:
        return
    for partition_index in range(1 << (n - 1)):
        partition = []
        element = []
        partition_positions = [position
                               for position in range(n)
                               if 1 << position & partition_index or position == n - 1]
        sizes = [b - a for a, b in zip([0] + partition_positions, partition_positions)]
        yield from split_into(lst, sizes)
```
But it works much slower:
```python
In [1]: import more_itertools

In [2]: a = '1'*1000

In [3]: %timeit more_itertools.partitions_2(a)
615 ns ± 53.6 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [4]: %timeit more_itertools.partitions(a)
379 ns ± 38.6 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

rominf:
I implemented this function using `split_into`. I like how it looks more, but unfortunately it's almost as twice as slower than original implementation:
```python
In [1]: from more_itertools import partitions

In [2]: def partitions_2(iterable): 
   ...:     lst = list(iterable)
   ...:     n = len(lst)
   ...:     if n == 0:
   ...:         return
   ...:     for partition_index in range(1 << (n - 1)):
   ...:         partition_positions = [position
   ...:                                for position in range(n)
   ...:                                if 1 << position & partition_index]
   ...:         sizes = [b - a for a, b in zip([-1] + partition_positions, partition_positions + [n])]
   ...:         yield list(split_into(lst, sizes))

In [3]: a = ' '*10

In [4]: %timeit _ = list(partitions(a))
1.83 ms ± 40.9 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [5]: %timeit _ = list(partitions_2(a))
3.4 ms ± 100 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

rominf:
I think that it would be more convenient for some cases not to convert `iterable` to `list`. This is true for strings. Most probably you want normal strings but not a list of strings with length 1 on the output. It's not hard for the user to call `list` if he has the non-subscribable object.
What do you think?

bbayles:
Thanks for the PR.

A few comments:
* It looks like this relies on slicing. Could you make it such that a generic iterable could be used?
* flake8 is complaining about a long line (see [Travis output](https://travis-ci.org/erikrose/more-itertools/builds/510325341?utm_source=github_status&utm_medium=notification))
* Could we get Tigran Saluev's blessing on including his implementation?

Thanks again.

Saluev:
👍

Saluev:
In my code though the first line was `collection = list(collection)`, so the problem 1 did not exist.

rominf:
> It looks like this relies on slicing. Could you make it such that a generic iterable could be used?

It's possible, but then you will get a list of lists of strings with length 1 if the input is the string. To make them strings user will have to write a quite long and slow code. Also, making `list(iterable)` will implicitly consume generators which is a bad thing and consume memory. I think that if the user wants to pass a generator, he has to explicitly (explicit is better than implicit, as Python Zen states) call `list`, which is not hard at all:
```python
partitions(list(generator))
```

Saluev:
Just in case: I don't agree with that. I think consuming a generator passed to an iteration tool is totally fine. It always happens anyway. `itertools.product` and other alike functions HAVE to do that. And I don't think we should do strings-driven optimizations in an itertools library as well.

(Also I consider the reference to Zen to be pure demagogy.)

rominf:
OK, then I propose to rename `collection` to `iterable` and add the following code so that anyone would be happy:
```python
from collections import abc
...
# in function
collection = iterable if isinstance(iterable, abc.Collection) else list(iterable)
```
What do you think?

rominf:
OK, implemented this.

@bbayles, are any questions left?

rominf:
OK, now it works with the iterables too. @bbayles, what do you think now?

rominf:
Now the iterables are supported, but collections are left as is, without converting them to strings.
@bbayles, is everything OK?

rominf:
I rewrote the code from scratch. Now it's shorter and works 30% faster.

more_itertools/more.py
```diff
@@ -2306,3 +2307,30 @@ def replace(iterable, pred, substitutes, count=None, window_size=1):
...
+    4. 1 <= n <= ilen(iterable)
+    5. isinstance(x, list) == True
+
+        >>> iterable = 'abcd'
+        >>> list(partitions(iterable))
+        [['abcd'], ['a', 'bcd'], ['ab', 'cd'], ['abc', 'd'], \
+['a', 'b', 'cd'], ['a', 'bc', 'd'], ['ab', 'c', 'd'], ['a', 'b', 'c', 'd']]
+
+    """
+    collection = iterable if _is_collection(iterable) else list(iterable)
```
bbayles: I think I'd prefer if this was just `list(iterable)`.

 reason: none of the other functions in this module . For example `list(substrings(iterable))` gives `[('a',), ('b',), ('a', 'b')]` - not `['a', 'b', 'ab']`.

Second reason: `_is_collection` is checking for a sequence, which is [kind of messy](https://stackoverflow.com/questions/43566044/what-is-pythons-sequence-protocol). It passes for `dict()` objects, using the keys. But it fails for `set()` objects. Is there a principled reason why we should allow one but not the other, for versions below 3.7?

(In [`SequenceView`](https://github.com/erikrose/more-itertools/blob/master/more_itertools/more.py#L1918) we just check to see if the object inherits from `Sequence`.)

rominf: > First reason: For example list(substrings(iterable)) gives [('a',), ('b',), ('a', 'b')] - not ['a', 'b', 'ab'].

And that's a bad thing in my opinion. Why the function which is called `substrings` returns generator (good) of tuples of strings with length 1? I may ask another way around. Do you think that `('a', 'b')` is substring of `'abc'`. I don't think so. Read for example documentation of `str.find`: "Return the lowest index in S where substring sub is found". OK, now I will try to put substring in the sense `more-itertools`:
```python
In [1]: 'abc'.find(('a', 'b'))
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-1-fabec44c37f8> in <module>
----> 1 'abc'.find(('a', 'b'))

TypeError: must be str, not tuple
```
Oops. That doesn't work.

> Second reason: _is_collection is checking for a sequence, which is kind of messy. It passes for dict() objects, using the keys. But it fails for set() objects. Is there a principled reason why we should allow one but not the other, for versions below 3.7?

I'm checking for [Collection](https://docs.python.org/3/library/collections.abc.html#collections.abc.Collection). That's not a `Sequence`. But you are right, I missed the dict. What do you think of this code:
```python
def _is_collection(iterable):
    try:
        iterable[slice()]
    except TypeError:
        return False
    return hasattr(iterable, '__len__')
```

more_itertools/tests/test_more.py
```diff
@@ -2311,3 +2311,26 @@ def test_iterable_substitutes(self):
...
+            ['a', 'bc', 'd'],
+            ['ab', 'c', 'd'],
+            ['a', 'b', 'c', 'd']
+        ]
+        self.assertEqual(actual, expected)
+
+    def test_empty(self):
+        iterable = []
+        actual = list(mi.partitions(iterable))
+        expected = []
```
rominf: I'm not sure now that this is correct.

>The empty set ∅ has exactly one partition, namely ∅ (https://en.wikipedia.org/wiki/Partition_of_a_set#Examples)

So, now I think it should be `[[[]]]`. Also, this is easier to implement and makes things more consistent.

rominf: I've changed the test and the code.

more_itertools/more.py
rominf: I've changed the `_is_collection`.

bbayles: I can understand why you want `str`s together, but the rest of the library doesn't do that. So if we can do `list` on the iterable, then I'll be happy to add this.

pylang: > So, I propose to accept this here, and change it everywhere else to make library behave more predictable.

This library is predictable in that it extends `itertools`, which generally returns iterators.  The iterator is an abstract artifact that encompasses all iterables, including strings.  Changing the library to return strings limits it.  Consider the dependents on this API.  That change would cripple a lot of code.

> Why the function which is called substrings returns generator (good) of tuples of strings with length 1 (very strange)?

Because it handles more than strings, e.g. `list(substring([1, 2, 3]))`.  Sure, empty items is weird, but the [definition](https://en.wikipedia.org/wiki/Substring) predates the library; see also [`subsequence`](https://en.wikipedia.org/wiki/Subsequence) and Haskell's implementation of substring.  Also note, the itertools `powerset` recipe returns an iterator, not a set.

rominf: Well, obviously, I'm outvoted, so, even though I still do not agree, I'm changing function to always return generator of list of lists.

rominf:
Now function converts all iterables to the lists.

@bbayles, @pylang is everything OK?

bbayles:
Merged with some additional tests and docstring changes. This will be in the next release. Many thanks!

rominf:
@bbayles, I think that the difference between [partition](https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.partition) and this function should be noted.

`partition` divides iterable into two, but those iterable are _not contiguous_ in the original iterable (function takes some items from iterable, then skips some items, then takes, and so on). But `partitions` generates all _contiguous_ itrerables.

I've tried to describe it in a formal way. You've removed that. OK, but then, insert, please, an informal note about this.

PR #278, opened 2019-03-24
Add substrings_indexes function

### LINKED ISSUE #277: Add another substrings function
I'm proposing to add another `substrings` function. Here is the implementation:
```python
def substrings_new(string, with_indexes=False):
    return ((string[i:i + length], i, i + length) if with_indexes else string[i:i + length]
             for length in range(len(string), 0, -1)
             for i in range(len(string) - length + 1))
```
Differences with `substrings`, which is already included:
1. Downside: this function requires the argument to be subscribable.
2. Order: my version returns longest substrings first. I think this is a more suitable behavior for NLP tasks.
3. Indexes can be included. Indexes are useful for replacing substrings in the original string, which is also good for NLP tasks.
4. The function returns normal strings, not the list of strings of length 1. No need to do expensive `join`.
5. Implementation: it's smaller and IMHO more readable.
6. Speed:
```python
In [1]: a = ' '*100

In [2]: %timeit list(substrings(a))
1.9 ms ± 41.5 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [3]: %timeit [''.join(substring) for substring in substrings(a)]
3.66 ms ± 171 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [4]: %timeit list(substrings_new(a))
1.3 ms ± 46.4 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

The only problem is what to do with 2 similar, but different functions in the library. That's why this is an issue, but not PR.

PS: I'm currently using this function in https://github.com/rominf/profanity-filter, so this function is useful.

rominf:
For differences between this and `substrings` function see the docstring
and https://github.com/erikrose/more-itertools/issues/277.

Closes #277.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -72,6 +72,7 @@ __all__ = [
     'stagger',
     'strip',
     'substrings',
+    'substrings_indexes',
     'unique_to_each',
     'unzip',
     'windowed',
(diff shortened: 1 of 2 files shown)


rominf:
Can you make a release and publish it on PyPI after merging this and #276?

bbayles:
Merged, thanks. I'll try to set up a release tomorrow or Tuesday.

PR #279, opened 2019-03-25
Add time_limited()

bbayles:
This PR adds `time_limited`, a function that yields items from an iterable for a limited amount of time.


pylang:
I like it.  It's something I'd see using with `make_decorator()`.

PR #285, opened 2019-04-07
Add only

bbayles:
Re: https://github.com/erikrose/more-itertools/issues/281, this PR adds `only`, which is like `one`, but it returns a default value in the `too_short` case instead of raising an Exception.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -2406,7 +2406,7 @@ def time_limited(limit_seconds, iterable):
 def only(iterable, default=None, too_long=None):
     """If *iterable* has only one item, return it.
     If it has zero items, return *default*.
-    If it has more than item, raise the exception given by *too_short*,
+    If it has more than item, raise the exception given by *too_long*,
     which is ``ValueError`` by default.
 
     >>> only([], default='missing')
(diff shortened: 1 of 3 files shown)


more_itertools/more.py
```diff
@@ -2401,3 +2401,35 @@ def time_limited(limit_seconds, iterable):
         if monotonic() - start_time > limit_seconds:
             break
         yield item
+
+
+def only(iterable, default=None, too_long=None):
+    """If *iterable* has only one item, return it.
+    If it has zero items, return *default*.
+    If it has more than item, raise the exception given by *too_short*,
```
pylang: "... more than *one* item ..."

`too_short` looks like it should  be `too_long`.

more_itertools/more.py
```diff
@@ -2401,3 +2401,35 @@ def time_limited(limit_seconds, iterable):
...
+    which is ``ValueError`` by default.
+
+    >>> only([], default='missing')
+    'missing'
+    >>> only([1])
+    1
+    >>> only([1, 2])  # doctest: +IGNORE_EXCEPTION_DETAIL
```
pylang: I would given an example passing in an exception, as in the tests.  At first, I could not tell what `too_long` required, e.g. an `int`.

PR #286, opened 2019-04-10
Fix numeric range with step zero test

MSeifert04:
The order of the loop and with statement was wrong. The test did not execute the datetime test because the loop is inside the raises-context-manager and is thus exited when the previous test failed.


bbayles:
Thanks!

PR #289, opened 2019-05-02
Add map_accumulate

bbayles:
This PR adds `map_accumulate()`, a function similar to `map_reduce()`. Instead of gathering all items into memory and then reducing them, this function applies the reduction as each item is encountered.

A simple demonstration is emulating `collections.Counter`: https://github.com/erikrose/more-itertools/blob/a23188dc0a3f0be513023bf5b529bbe01c1410b4/more_itertools/more.py#L2267-L2273

A better one is keeping track of parts of a complicated data structure:
https://github.com/erikrose/more-itertools/blob/a23188dc0a3f0be513023bf5b529bbe01c1410b4/more_itertools/more.py#L2277-L2297

[Here](https://github.com/obsrvbl/flowlogs-reader/blob/63977dc787f289f8f62582a079efc67701cd9c8b/flowlogs_reader/aggregation.py) is a real-world example of this process in action. [Here](https://stackoverflow.com/a/2937417/353839) is another.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -51,6 +51,7 @@ __all__ = [
     'locate',
     'lstrip',
     'make_decorator',
+    'map_accumulate',
     'map_reduce',
     'numeric_range',
     'one',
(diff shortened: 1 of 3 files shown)


more_itertools/more.py
```diff
@@ -2254,6 +2255,70 @@ def map_reduce(iterable, keyfunc, valuefunc=None, reducefunc=None):
...
+    current value and the item.
+
+    Keeping a running total:
+
+        >>> iterable = ['ash', 'ape', 'bat', 'cat', 'cow']
+        >>> keyfunc = lambda item: item[0]  # Organize by first letter
+        >>> initfunc = lambda: 0
+        >>> updatefunc = lambda value, item: value + 1
+        >>> result = map_accumulate(iterable, keyfunc, initfunc, updatefunc)
+        >>> result == {'a': 2, 'b': 1, 'c': 2}
```
pylang: Save a line with >>> assert result == ...

pylang:
Nice.  Some feedback:

- The first example seems to demonstrate a `Counter` + predicate (novel).
- The second example reminds me of [this](https://stackoverflow.com/a/9044111/4531270) date range example compiled in a dict solve with `namedtuple`.
- `initfunc` sounds like the `default_factory` parameter in `defaultdict`

---
Here are the examples solved with `defaultdict`:


First example

```python
dd = ct.defaultdict(int)
for x in iterable:
    dd[x[0]] += 1
dd
# defaultdict(int, {'a': 2, 'b': 1, 'c': 2})
```


Second example:

```python
dd = ct.defaultdict(Tracker)
for d in iterable:
    k, t = d["key"], d["time"]
    dd[k].first_time = min(t, dd[k].first_time)
    dd[k].last_time = max(t, dd[k].last_time)
    dd[k].hits += 1

result = dict(dd)


assert result[1].first_time == dt.datetime(2019, 5, 1, 0, 0)
assert result[1].last_time == dt.datetime(2019, 12, 31, 0, 0)
assert result[1].hits == 2

assert result[2].first_time == dt.datetime(2019, 1, 1, 0, 0)
assert result[2].last_time == dt.datetime(2019, 1, 1, 0, 0)
assert result[2].hits == 1

```

So far, we can solve those examples with iteration while updating a `defaultdict`.  I value the abstraction, though imo I am +0 on adding it to the library.  I think some additional, compelling, real world examples that are hard to solve with `defaultdict` alone could easily sway me to +1.

PR #291, opened 2019-06-18
Adds function `iterchunked` (like `chunked`, but yields iterables instead of lists)

davebelais:
I'm not sure if this will be of any value for other users, however when working with very large datasets I frequently want my "chunk" to be an iterator rather than a list, so I've been using this implementation for a while. It works exactly as `chunked` aside from yielding iterators instead of lists.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -47,6 +47,7 @@ __all__ = [
     'intersperse',
     'islice_extended',
     'iterate',
+    'iterchunked',
     'last',
     'locate',
     'lstrip',
@@ -88,16 +89,15 @@ _marker = object()
 
 
 def chunked(iterable, n):
-    # type: (typing.Iterable, int) -> typing.Iterable[typing.Iterable]
-    """Break *iterable* into iterators of (at most) length *n*:
+    """Break *iterable* into lists of length *n*:
 
-        >>> list(list(chunk) for chunk in chunked([1, 2, 3, 4, 5, 6], 3))
+        >>> list(chunked([1, 2, 3, 4, 5, 6], 3))
         [[1, 2, 3], [4, 5, 6]]
 
     If the length of *iterable* is not evenly divisible by *n*, the last
-    returned iterable will be shorter:
+    returned list will be shorter:
 
-        >>> list(list(chunk) for chunk in chunked([1, 2, 3, 4, 5, 6, 7, 8], 3))
+        >>> list(chunked([1, 2, 3, 4, 5, 6, 7, 8], 3))
         [[1, 2, 3], [4, 5, 6], [7, 8]]
 
     To use a fill-in value instead, see the :func:`grouper` recipe.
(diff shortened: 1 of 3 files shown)


bbayles:
Thanks for the PR. Could you (a) fix the build failure, and (b) add some tests that show that the returned things are iterables? Specifically I'd want to see something like "get the first chunk, get the second chunk, consume the second chunk, show that the first chunk is not affected."

davebelais:
a) Looks like the build fail was because my lines were too long--yes, I'll fix that.
b) Good call, yes I'll do that.

pylang:
Good point @davebelais.  Subiterators are useful with big data.  

The return values on these tools has come to my attention as well.  The other day, I wanted to cast to a set some `partitions` from a permutation.  I could not because `partitions` also returns sublists, and the workaround was involved.

My only concern is that the precedent from this PRs approach may lead to future PRs of the same tools in the library that vary by output type alone, which can be hard to maintain and learn.

---

Is there a general way to control the return type of these tools, i.e. tools that return sublists (`chunked`, `partitions`, `split_*`, `unique_to_each`)?

Example:

Rather than make a host of similar tools that vary by return type:

```python
chunked(...)                # default
iterchunked(...)            # good w/memory
tuplechucked(...)           # good with sets + dicts
```

Maybe we could we add an option to the signature:

```
chunked(..., return_type=iter)
```

or make a return type converter:

```python
iterchunked = morph_type(chunked, iter)
iterchunked(...)
```
Or any other general approach.

more_itertools/more.py
```diff
@@ -111,6 +112,78 @@ def chunked(iterable, n):
...
+    To iterate over lists instead of sub-iterators, see the :func:`chunked` recipe.
+
+    """
+
+    # Make sure we are working with an iterator
+    if not (
+        hasattr(iterable, '__next__') and
+        callable(iterable.__next__)
+    ):
+        iterable = iter(iterable)
```
bbayles: Might it be simpler to just do `iter(iterable)` without these checks?

more_itertools/more.py
```diff
@@ -111,6 +112,78 @@ def chunked(iterable, n):
...
+
+    # Define functions for yielding sub-iterators
+
+    def yield_and_dequeue(queue, limit):
```
bbayles: Could we replace this function with `islice(queue, limit)`?

more_itertools/more.py
```diff
@@ -111,6 +112,78 @@ def chunked(iterable, n):
...
+        else:
+            chunk = iter(chunk)
+
+        return chunk
+
+    # Retrieve the first chunk of values (or `None`)
+    value = dequeue_and_yield(iterable, n)
+
+    # Only yield `value` so long as it is an iterator, as `None` indicates we've exhausted `iterable`
+    while value is not None:
```
bbayles: Elsewhere in the library we use the `_marker` object as a sentinel. This allows iterables to contain `None`.

more_itertools/more.py
```diff
@@ -111,6 +112,78 @@ def chunked(iterable, n):
...
+                break
+
+    def dequeue_and_yield(queue, limit):
+        # type: (Iterable, int) -> Iterable
+        """
+        This function first checks the next value in the queue to make sure we haven't already reached the end of the
+        iterator, *then* de-queues and yields at most `limit` number of items from the iterator--but only if there is at
+        least one upcoming value. This is to prevent the parent function from yielding empty iterators.
+        """
+        if not limit:
```
bbayles: Might it be better to check whether the argument is > 0 at the top of the function and omit this branch?

bbayles: Scratch that; I see the `limit - 1` below.

bbayles:
A few other tests I'll request:
* Empty iterable
* Iterable with `None` as an element
* Something like this 

```python
def gen():
    yield 0
    raise ValueError("!")
    yield from range(1, 10)

all_chunks = iterchunked(gen(), 4)
chunk = next(all_chunks)
```

davebelais:
...closing in favor of `ichunked`. Thanks!

Issue #72, opened 2016-11-12
Version of chunked that emits iterators and not lists


bbayles:
PR #56 and PR #58 both raise the idea of having a function that splits an iterator into a group of sub-iterators with a fixed length. That is, a version of `chunked()` that emits iterators instead of lists.

I was hoping to be able to modify `chunked()` to do this via a parameter or something, but I think performance would suffer. The simple version from #58 isn't viable.

```python
>>> for func in (original, ichunked_new, ichunked_pr58):
...     def stmt():
...         iterable = range(2000)  # Obviously performance will vary with iterable and n
...         n = 101
...         all_chunks = list(func(iterable, n))
...         assert len(all_chunks) == 20
... 
...     result = timeit(stmt, number=10000)
...     print(func.__name__, result)
original 0.7484618649759796
ichunked_new 0.971211633994244
ichunked_pr58 17.300433913012967
```

So I think a separate function (`ichunked`, I guess) is called for!

---

```python
from itertools import chain, islice, zip_longest
from more_itertools import consume, peekable
from timeit import timeit

def original(iterable, n):
    it = iter(iterable)
    while True:
        chunk = list(islice(it, n))
        if not chunk:
            return
        yield chunk


def ichunked_pr58(iterable, n, emit_lists=True):
    p = peekable(iterable)
    while p:
        chunk = islice(p, n)
        if emit_lists:
            yield list(chunk)
        else:
            yield chunk
            consume(chunk)

def ichunked_new(iterable, n, emit_lists=True):
    it = iter(iterable)
    while True:
        test_chunk = islice(it, n)
        try:
            item = next(test_chunk)
        except StopIteration:
            return
        chunk = chain([item], test_chunk)
        if emit_lists:
            yield list(chunk)
        else:
            yield chunk
            consume(test_chunk)


for func in (original, ichunked_new, ichunked_pr58):
    def stmt():
        iterable = range(2000)
        n = 101
        all_chunks = list(func(iterable, n))
        assert len(all_chunks) == 20

    result = timeit(stmt, number=10000)
    print(func.__name__, result)
```

bbayles:
Upon further reflection, this might be incoherent. If you don't cache the items from the iterable somewhere (i.e. in a list), you can't proceed past them in the iterator. You'll either wind up with something that spins forever if you try to `list()` it, or something that only works if you consume the items in the right order.

PR #292, opened 2019-06-22
ichunked, now more than ever

bbayles:
Re: issue #72 - this PR was inspired by @davebelais in PR #291.

As the issue states, a few people have tried to do a chunked-that-emits-iterators.

The version from PR #58, my version called `ichunked_new` in the issue, and the version from PR #291 all exhibit strange behavior when reading the chunks out of order.

This version handles that (previous chunks' elements are cached until they are needed), as well as 0-length iterables, iterables with `None` elements, etc.

I think I'm happy with this, 3 years after having given up. @davebelais, your thoughts? Re: @pylang's comment, I don't think this needs to merged with `chunked`; and I don't think the other "Grouping" itertools (except `grouper`) are really candidates for lazy versions?

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -14,6 +14,7 @@ These tools yield groups of items from a source iterable.
 **New itertools**
 
 .. autofunction:: chunked
+.. autofunction:: ichunked
 .. autofunction:: sliced
 .. autofunction:: distribute
 .. autofunction:: divide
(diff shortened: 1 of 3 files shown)


bbayles:
(`igrouper` would be `ichunked(padded(iterable, fillvalue, n, True), n)`)

davebelais:
Love it!

Issue #296, opened 2019-07-10
Set partitioning


ktbarrett:
I would like to the library include a function that implements partitioning of a set N-ways that does not preserve order (that's how it differs from `partitions`).

The 2-way set partitions of `range(4)` are
```python
((0,), (1, 2, 3))
((1,), (0, 2, 3))
((2,), (0, 1, 3))
((3,), (0, 1, 2))
((0, 1), (2, 3))
((0, 2), (1, 3))
((0, 3), (1, 2))
```

If this is desired, I already have an implementation of `integer_partitions` and `set_partitions` that yield lexicographically ordered partitions.

bbayles:
Please do post your implementations, yes.

ktbarrett:
See #297

PR #297, opened 2019-07-11
integer and set partitions

ktbarrett:
Implements lexically-ordered integer and set partitions generators. The generators can generate only k-partitions, or if no k is given, it will generate all number of partitions.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -61,7 +61,6 @@ __all__ = [
     'padded',
     'partitions',
     'set_partitions',
-    'integer_partitions',
     'peekable',
     'replace',
     'rlocate',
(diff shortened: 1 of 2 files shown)


more_itertools/more.py
```diff
@@ -2402,6 +2405,95 @@ def partitions(iterable):
...
+        for k in range(1, N + 1):
+            yield from helper(N, k, 1)
+    else:
+        yield from helper(N, k, 1)
+
+
+def set_partitions(iterable, k=None):
+    """
+    Generates k set partitions of iterable, or all partitions if k is not given
+    A set partition is a wasy to split a set into unique subsets that when
```
bbayles:
```suggestion
    A set partition is a way to split a set into unique subsets that when
```

bbayles:
Closes #296. Many thanks!

PR #301, opened 2019-07-27
New implementation of `set_partitions`

jferard:
See https://github.com/erikrose/more-itertools/issues/300 for discussion.

Remark: the `FrozenMultiset` class could be used to make some other tests more robust. Typical use case: the theoretical output is a collections of elements 1. with possible duplicates and 2. order doesn't matter. Obviously, an implementation has to choose one of the possible orders, but the tests should be able to ignore this order.

Example: `powerset`.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -1,3 +1,5 @@
+import warnings
+
 from collections import Counter, defaultdict, deque
 from collections.abc import Sequence
 from functools import partial, wraps
@@ -18,7 +20,7 @@ from itertools import (
     zip_longest,
 )
 from operator import itemgetter, lt, gt, sub
-from sys import maxsize, version_info
+from sys import maxsize
 from time import monotonic
 
 from .recipes import consume, flatten, powerset, take, unique_everseen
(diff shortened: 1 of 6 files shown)


bbayles:
Could you take a look at the `flake8` errors in the Travis output?

Python 3.4 doesn't work with some of your syntax, but I'm OK dropping it in the next version (see my [earlier comment](https://github.com/erikrose/more-itertools/pull/282#issuecomment-478063682)).

jferard:
Didn't know about `flake8`: nice tool. I installed it and will make an extensive use of it from now on!

> something that fails with 3.4 and not everything else.

I guess this is it.

more_itertools/more.py
```diff
@@ -2419,43 +2418,39 @@ def set_partitions(iterable, k=None):
...
-
-    def part_inds(inds, k, prev):
-        """Generates set partitions by index"""
-        if (k <= 1) and (less(prev, inds)):
-            yield (inds,)
+    L = list(iterable)
+    n = len(L)
+    if k is not None:
+        if k < 1:
+            raise ValueError(
```
bbayles: [Coveralls](https://coveralls.io/builds/24840087/source?filename=more_itertools%2Fmore.py#L2430) says this branch is not tested - could you test it?

more_itertools/more.py
```diff
@@ -2419,43 +2418,39 @@ def set_partitions(iterable, k=None):
...
+                "Can't partition in a negative or zero number of groups")
+        elif k > n:
+            return []
```
bbayles: This is also not tested yet. Also, could you make it `yield []` and then `return` instead?

bbayles:
Could you merge in master now that #282 is in?

more_itertools/more.py
jferard: Was some kind of typo: if k > n, there is group to yield, just the end of the generator (a bare `return`).

jferard:
Made a `git pull --rebase upstream master`. Hope it's ok.

PR #309, opened 2019-09-26
Bring types to the recipes

### LINKED ISSUE #262: Add type annotations

ilai-deutel:
This PR adds type hints ([PEP 484](https://www.python.org/dev/peps/pep-0484/)) to the recipes.
Type information is distributed following [PEP 561](https://www.python.org/dev/peps/pep-0561/).
The implementation of the recipes, as well as tests, are type-checked using `mypy` with strict options.

Fixes #262


bbayles:
Thanks for this - great work.

I might have some quibbles and questions as I read through this in details.  My first set are:
* Some of the annotations, particularly ones like `side_effect` are really un-aesthetic. Like, enough that I might rather have the type annotations in a separate file (contra my comment [here](https://github.com/erikrose/more-itertools/issues/262#issuecomment-462589592)).
* Given the, uh, lack of beauty, I think I'd rather the test files not get annotated (also `setup.py`)
* Would relaxing the strictness checks remove the need to do some of the `cast`s?

more_itertools/more.py
```diff
@@ -949,7 +1105,35 @@ def walk(node, level):
     yield from walk(iterable, 0)
 
 
-def side_effect(func, iterable, chunk_size=None, before=None, after=None):
+@overload
+def side_effect(
```
gabbard: @ilai-deutel : I was taking a quick look at `side_effect` since @bbayles mentioned it was "un-aesthetic".  At a glance it is not obvious to me why the `@overloaded` versions are necessary. Perhaps you could add comment about it for the benefit of future readers of the code?

more_itertools/more.py
```diff
@@ -949,7 +1105,35 @@ def walk(node, level):
...
+    after: Optional[Callable[[], object]] = ...,
+) -> Iterator[_T]:
+    ...
+
+
+def side_effect(
+    func: Union[Callable[[_T], object], Callable[[List[_T]], object]],
+    iterable: Iterable[_T],
+    chunk_size: Optional[int] = None,
+    before: Optional[Callable[[], object]] = None,
```
gabbard: Should this be `Optional[Callable[], Any]`, judging from the example of callbacks in https://www.python.org/dev/peps/pep-0484/#the-any-type ?

more_itertools/more.py
gabbard: Aha, I see - it is because you want to enforce that `chunk_size` is specified if and only if `func`  takes a `List` as its input.

more_itertools/more.py
```diff
@@ -949,7 +1105,35 @@ def walk(node, level):
...
+    chunk_size: None = ...,
+    before: Optional[Callable[[], object]] = ...,
+    after: Optional[Callable[[], object]] = ...,
+) -> Iterator[_T]:
+    ...
+
+
+@overload
+def side_effect(
+    func: Callable[[List[_T]], object],
```
gabbard: Is `List` unnecessarily constraining here?  Shouldn't any function which can take an `Iterable[_T]` work?  (personally, I very rarely write any function with an argument type of `List` - it's usually either `Iterable` or `Sequence`, depending on what I need to do with it)

ilai-deutel:
@bbayles sorry for the late reply.

> Some of the annotations, particularly ones like `side_effect` are really un-aesthetic. Like, enough that I might rather have the type annotations in a separate file (contra my comment [here](https://github.com/erikrose/more-itertools/issues/262#issuecomment-462589592)).

I tried to move the overloaded signatures to a separate `.pyi` file, but unfortuantely `mypy` would then ignore the in-file annotations. I can move everything to separate files if you prefer, but it makes type annotations harder to maintain.

> Given the, uh, lack of beauty, I think I'd rather the test files not get annotated (also `setup.py`)
 
Sure.

> Would relaxing the strictness checks remove the need to do some of the `cast`s?

I believe each `cast` is needed regardless of strictness options. An alternative is to use less explicit `# type: ignore` instead.

- [[1]](https://github.com/ilai-deutel/more-itertools/blob/31ae7d07190fc07a33808d94f13fbabe4720ffa6/more_itertools/more.py#L527) is needed to type decorators that do not have a fixed signature, see python/mypy#3157
- [[2]](https://github.com/ilai-deutel/more-itertools/blob/31ae7d07190fc07a33808d94f13fbabe4720ffa6/more_itertools/more.py#L941) and [[3]](https://github.com/ilai-deutel/more-itertools/blob/31ae7d07190fc07a33808d94f13fbabe4720ffa6/more_itertools/more.py#L945) have been removed
- [[4]](https://github.com/ilai-deutel/more-itertools/blob/31ae7d07190fc07a33808d94f13fbabe4720ffa6/more_itertools/more.py#L2551) is used to indicate that we know that `windowed(cycle(lst), len(lst))` will never yield `None` because `cycle(lst)` is infinite. This cannot be inferred from the type.
- The `cast` in [[5]](https://github.com/ilai-deutel/more-itertools/blob/31ae7d07190fc07a33808d94f13fbabe4720ffa6/more_itertools/more.py#L2783) tells mypy that within the `try ...` section, iterable is expected to have a length and be reversible
- In [[6]](https://github.com/ilai-deutel/more-itertools/blob/31ae7d07190fc07a33808d94f13fbabe4720ffa6/more_itertools/more.py#L2838), we know `windows` do not contain `None`, because we used `chain` previously to pad the iterable with enough marker. 
Instead of using `cast`, another solution is to use a `fillvalue=_marker` instead of `None`. This would give the right type to `windows`, but could be misleading (since we know the `fillvalue` won't be used).

```python
it = chain(iterable, [_marker] * (window_size - 1))  # type: Iterator[Union[_T, _Marker]]
windows = windowed(it, window_size, fillvalue=_marker)
```
- In [[7]](https://github.com/ilai-deutel/more-itertools/blob/31ae7d07190fc07a33808d94f13fbabe4720ffa6/more_itertools/recipes.py#L350), mypy consider both `iterable` and `n` as `Union[Iterable[_T], int]`, because of the deprecated signature. Without `cast`, `args = [iter(iterable)] * n` wouldn't make sense to mypy (you can't multiply a list with an iterable, or iterate over an integer).

more_itertools/more.py
ilai-deutel: `Any` and `object` are similar but slightly different. `Callable[[], object]` will be stricter because of this:

> When the type of a value is `object`, the type checker will reject almost all operations on it, and assigning it to a variable (or using it as a return value) of a more specialized type is a type error. On the other hand, when a value has type `Any`, the type checker will allow all operations on it, and a value of type `Any` can be assigned to a variable (or used as a return value) of a more constrained type.

If we take a function as an argument and call it, we can only use the return value in a way that would be compatible with any return type. For instance, we shouldn't be able to add an integer to the result, because the result might be of type `List[complex]` for instance.

```python
a: Callable[[], Any]
b: Callable[[], object]

if a():
    ...
if b():
    ...

a() + 5
b() + 5  #  error: Unsupported operand types for + ("object" and "int")  [operator]
```

more_itertools/more.py
ilai-deutel: `Callable[[List[_T]], object]` is actually less restrictive than `Callable[[Iterable[_T]], object]`, because any function that can take an iterable as an argument can take a list as an argument. Functions with an argument type of  `Iterable[_T]`, `Collection[_T]`, or any other supertype of `List[_T]` will be accepted.

```python
def f(a: Callable[[List[str]], object]):
    ...

def g(a: Callable[[Sequence[str]], object]):
    ...

def u(x: List[str]):
    ...

def v(x: Sequence[str]):
    ...

f(u)
f(v)
g(u)  #  Argument 1 to "g" has incompatible type "Callable[[List[str]], Any]"; expected "Callable[[Sequence[str]], object]"  [arg-type]
g(v)
```

In this example, both `u` and `v` are accepted as arguments of `f`, but only `v` is a valid argument for `g`.

bbayles:
Thanks for the detailed response!

> I can move everything to separate files if you prefer, but it makes type annotations harder to maintain.

Yeah, I think at this point I'd prefer that. I'm still turned off by the `cast`s as well.

more_itertools/more.py
gabbard: Got it! Thanks for spelling that out for me.

more_itertools/more.py
gabbard: Thanks!

ilai-deutel:
I removed type annotations for tests, and moved type annotations for `recipes` and `more` to separate stub files.

Unfortunately, this means that we don't have a way to test that the stubs are correct anymore: `mypy` doesn't test the stubs against the actual implementation (see python/mypy#5028), and the tests are no longer type-checked (since I removed the variable annotations and `# type: ignore` comments). Typed tests had helped me find issues with type annotations when I first added them, for instance the fact that the `validator` in `filter_except` is of type `Callable[[Any], object]`, not `Callable[[_T], object]`.

[`retype`](https://github.com/ambv/retype) could be a solution in the future, but it doesn't currently work with the `more_itertools` stub files (ambv/retype#20, ambv/retype#22).

bbayles:
Many thanks for this!

PR #310, opened 2019-10-01
Optimize partition w/ expensive predicate

stevecj:
Since calling the predicate might be expensive, call it ahead of the tee so it will not need to be evaluated separately in each branch of the tee, taking up to twice as much time as necessary, assuming an equal number of results will be evaluated through each of the resulting iterables.

### DIFF
--- a/more_itertools/recipes.py
+++ b/more_itertools/recipes.py
@@ -14,7 +14,6 @@ from itertools import (
     combinations,
     count,
     cycle,
-    filterfalse,
     groupby,
     islice,
     repeat,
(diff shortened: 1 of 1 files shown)


stevecj:
I know that the code in this branch is different in style from other code in the project because it uses generator expressions rather than reusing other functions. It was not obvious to me, however, how to accomplish the same thing more directly with reuse than with these generators.

I would welcome suggestions on how to make this code better aligned with other code in more-itertools if that is desirable.

bbayles:
Thanks for this!

MSeifert04:
I hope it's okay if I ask here: Is it okay if I also apply this change to my library `iteration_utilities`?

PR #317, opened 2019-10-15
Add repeat_last

d-ryzhikov:
Add `repeat_last(iterable)` function with the following functionality, which after the given *iterable* is exhausted keeps yielding its last element.
```python
>>> list(islice(repeat_last(range(3)), 5))
[0, 1, 2, 2, 2]
```


more_itertools/more.py
```diff
@@ -1202,6 +1203,26 @@ def padded(iterable, fillvalue=None, n=None, next_multiple=False):
             yield fillvalue
 
 
+def repeat_last(iterable):
```
bbayles: What do you think of having a `default` value to emit endlessly, if *iterable* is empty?

more_itertools/more.py
```diff
@@ -1202,6 +1203,26 @@ def padded(iterable, fillvalue=None, n=None, next_multiple=False):
...
+    """After the *iterable* is exhausted, keep yielding its last element.
+
+        >>> list(islice(repeat_last(range(3)), 5))
+        [0, 1, 2, 2, 2]
+    """
+    it = iter(iterable)
```
bbayles: I think we can simplify this implementation. What about

```python
item = _marker
for item in iterable:
    yield item

final = default if (item is _marker) else item
yield from repeat(final)
```

d-ryzhikov: Didn't notice that there was a marker and didn't want to introduce one. This implementations is much cleaner.

more_itertools/more.py
```diff
@@ -1202,6 +1203,26 @@ def padded(iterable, fillvalue=None, n=None, next_multiple=False):
             yield fillvalue
 
 
+def repeat_last(iterable, default=None):
```
MSeifert04: Shouldn't the default for default be `_marker`?

bbayles:
Excellent - many thanks!

PR #318, opened 2019-10-20
Move tests outside of package

jdufresne:
This avoids installing tests to the users site-packages directory upon
install from pip. For end users, these files are unnecessary on their
production systems and only result in additional code installed.

The tests will continue remain in the project at the top-level. They can
still be run with one of:

    $ tox
    $ python3 setup.py test
    $ python3 -m unittest

Tests are still distributed with the sdist through the user of
MANIFEST.in to allow distro packagers to run the tests.

As the package is no longer nested, can replace find_package() in
setup.py.


bbayles:
Yes, this is an improvement. Many thanks!

PR #320, opened 2019-10-20
Tidy up Travis configuration file

jdufresne:
Remove 'dist: "xenial"'. It is now the default and unnecesary.

Change to use pypy3 alias for testing against the latest PyPy3 release.


bbayles:
You're on a roll - thanks!

PR #321, opened 2019-10-20
Add trove classifiers

jdufresne:
These help new library users know, at a quick glance, if the library is
suitable for including in a project. These values are displayed on the
PyPI project page.

Documents support for Python 3.5, 3.6, 3.7 and 3.8.

Documents support for PyPy in addition to CPython.


bbayles:
I intentionally didn't set these, because (a) I wanted to signal "this library supports all current versions of Python 3," and (b) didn't want to keep changing the list as Python versions go in and out of style.

Open to being persuaded here, but I don't think I've ever actually filtered on a Python version classifier beyond "works on 2.x / works on 3.x" - have you?

jdufresne:
That's totally fine, up to you. I'll give my response and reasoning below, but really either choice is fine.

>  I wanted to signal "this library supports all current versions of Python 3,"

I understand this intention, but -- FWIW -- when I see projects without versions listed, I can't assume this. What I end up doing it going to their GitHub page and seeing what versions are tested on Travis or similar. Otherwise, I don't know if the project is stale or up to date on testing current Python versions. So, in the end, the attempt to make less work resulted in more for me. Ultimately, this is a minor thing.

>  didn't want to keep changing the list as Python versions go in and out of style.

I appreciate this as a goal, but I assume this is already happening to some extent for `tox.ini` and `.travis.yml`. So the effort save is just one additional place and not removed entirely. In the case of Python 3.8, you had a volunteer add it :slightly_smiling_face: and had the trove classifiers already existed I would have added it there too.

---

Thanks for the project. Again, either choice you make is fine, only trying to help.

bbayles:
OK, I'm convinced!

PR #323, opened 2019-10-28
Cleanup and improve tox.ini

jdufresne:
Replace '{envbindir}/python' with '{envpython}'
See: https://tox.readthedocs.io/en/latest/config.html#substitutions-for-virtualenv-related-sections

---

Drop "discover" argument. It is now the default for unittest.

https://docs.python.org/3/library/unittest.html#command-line-interface

> When executed without arguments Test Discovery is started:

---

Add the '{posargs}' placeholder to allow passing arguments to unittest
from the tox command line.

https://tox.readthedocs.io/en/latest/config.html#substitutions-for-positional-arguments-in-commands

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -2,4 +2,4 @@
 envlist = py{35,36,37,38}
 
 [testenv]
-commands = {envbindir}/python -m unittest discover -v
+commands = {envpython} -m unittest -v {posargs}


bbayles:
These look like improvements, thanks!

I'm inclined to remove Tox, but I said I wouldn't in 2016 and was going to quietly do it in 2021 or something. Would anyone care?

jdufresne:
tox allows contributors to test all supported Python versions with a single, familiar command, `tox`. The alternative would be running the same unittest command across different Python versions (or waiting for a Travis round trip).

tox would also allow the project to change its test runner, should it ever choose to. For example, one day you may wish to move the project to pytest. If that were to happen, contributors could still use the same generic entry point, `tox` to run the tests.

PR #325, opened 2019-10-29
Include stub files in setup.py

### LINKED ISSUE #324: Stub files not being installed
I just tried installing master to use the type annotations from https://github.com/erikrose/more-itertools/pull/309, but they are probably not being installed because the standard mechanism that gathers files to install doesn't include `.pyi` files.

To reproduce:
```bash
pip install --upgrade git+https://github.com/erikrose/more-itertools.git@583367c # current master at the time of the issue
ls $VIRTUAL_ENV/lib/python3.7/site-packages/more_itertools/
```
This isn't hard to fix, for a pure stub package I have [code that deals with this](https://github.com/fmagin/networkx-stubs/blob/master/setup.py#L14-L23) which according to the typing gitter is the best way to do it currently. Because this package should obviously include the regular `.py` files too some extra work is required which should be fairly trivial though. I might be able to look into this again later today if someone else wasn't quicker.

fmagin:
This is hacked together code for which I can't even quite remember where I originallyt got it from, but this at least includes the `.pyi` files when installing non editable. `pip install -e ./more-itertools` worked before because it just links the folder instead of copying files as far as I understand.
The files are few enough that if the code is really an issue, they could just be listed explicitly.

I quickly tested this by checking if PyCharm finds the stubs when installing non editable via  pip, because the issue was the `.pyi` files not being copied I assume that this works for mypy or anything else too.

Fixes #324

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,7 @@
 from re import sub
 
 from setuptools import setup
-
+import os
 
 def get_long_description():
     # Fix display issues on PyPI caused by RST markup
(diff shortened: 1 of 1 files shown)


bbayles:
Thanks! The wheel now includes the typing files.

graingert:
@bbayles any chance of publishing the wheel on pypi?

bbayles:
I'll publish a release this weekend.

PR #326, opened 2019-11-03
refactor unique_everseen

SergBobrovsky:
less code, minimum speed loss

### DIFF
--- a/more_itertools/recipes.py
+++ b/more_itertools/recipes.py
@@ -377,27 +377,16 @@ def unique_everseen(iterable, key=None):
     seenset_add = seenset.add
     seenlist = []
     seenlist_add = seenlist.append
-    if key is None:
-        for element in iterable:
-            try:
-                if element not in seenset:
-                    seenset_add(element)
-                    yield element
-            except TypeError:
-                if element not in seenlist:
-                    seenlist_add(element)
-                    yield element
-    else:
-        for element in iterable:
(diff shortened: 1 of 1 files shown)


bbayles:
Thanks for the PR!. Could you quantify the speed difference?

bbayles:
I get anywhere from 5% faster to 10% slower, depending on the key function:

```python
from timeit import Timer
from random import choice

from more_itertools.recipes import consume, unique_everseen, new_unique_everseen

pool = list(range(10000))
iterable = []
for i in range(1000):
    iterable.append(choice(pool))

results = []
for func in unique_everseen, new_unique_everseen:
    def stmt():
        consume(func(iterable, key=lambda x: str(x)))

    timer = Timer(stmt)
    n, result = timer.autorange()
    results.append(result)


print(results[1] / results[0] - 1)
```

SergBobrovsky:
I'm sorry, email is not my priority, so I'm late.
Anyway, You'd better know:

- pool = list(range(10000))
- iterable = []
- for i in range(1000):
-     iterable.append(choice(pool))
+ iterable = random.sample(range(10000), 1000)

- consume(func(iterable, key=lambda x: str(x)))
+ consume(func(iterable, key=str))

Be pythonic.
Best regards.

ср, 6 нояб. 2019 г. в 05:40, Bo Bayles <notifications@github.com>:

PR #327, opened 2019-11-03
refactor partition

SergBobrovsky:
# functional, smaller, faster up to 1.5 in example below executed on CPython

for part in partition(lambda x: x, [False, True] * 10 ** 7):
    for el in part:
        pass

### DIFF
--- a/more_itertools/recipes.py
+++ b/more_itertools/recipes.py
@@ -12,6 +12,7 @@ from collections import deque
 from itertools import (
     chain,
     combinations,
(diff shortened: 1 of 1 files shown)


bbayles:
Thanks for the PR. I'm going to decline this one - I ran some tests and couldn't see that this was consistently faster. Also, since we're not adding a feature, I'm disinclined to change the recipe that's listed in the docs.

Nice use of `compress`!

PR #331, opened 2019-11-09
refactor one

SergBobrovsky:


bbayles:
Skipping this one, as I think the existing implementation is a bit clearer. Thanks for the suggestion!

PR #335, opened 2019-11-09
refactor windowed part two

SergBobrovsky:
smaller, faster.


bbayles:
Thanks for the PR. This looks nice; if you can get it to pass the existing tests, I'll merge it. See https://github.com/erikrose/more-itertools/pull/63 for the original discussion.

PR #338, opened 2019-11-12
refactor divide

SergBobrovsky:
avoids unnecessary copying
smaller
a bit faster on short parts, example
```
divide(10 ** 6, range(10 ** 6))
```

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -5,6 +5,7 @@ from collections.abc import Sequence
 from functools import partial, wraps
 from heapq import merge
 from itertools import (
+    accumulate,
     chain,
     compress,
     count,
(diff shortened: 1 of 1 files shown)


more_itertools/more.py
```diff
@@ -1441,17 +1442,11 @@ def divide(n, iterable):
     """
     if n < 1:
         raise ValueError('n must be at least 1')
-
-    seq = tuple(iterable)
+    seq = iterable if hasattr(iterable, '__lengh__') else tuple(iterable)
```
bbayles: This won't do what we want for things like `dict`s. They have `__length__`, but won't allow for slicing.

Really what we want to check for is whether the iterable supports slicing. Checking whether something is an instance of `collections.abc.Sequence` could work here, or trying to retrieve a `slice(0)` and catching an exception.

PR #343, opened 2019-11-22
Improve difference

Jabbey92:
Initial issue: https://github.com/erikrose/more-itertools/issues/342

Changed use of `lambda` to `starmap`&`zip` for performance. (also improves code readability)

Changed `initial` parameter to `_marker` as not to limit `None` as initial value.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -1966,7 +1966,7 @@ def consecutive_groups(iterable, ordering=lambda x: x):
         yield map(itemgetter(1), g)
 
 
-def difference(iterable, func=sub, *, initial=None):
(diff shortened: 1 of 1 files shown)


more_itertools/more.py
```diff
@@ -2014,7 +2014,7 @@ def difference(iterable, func=sub, *, initial=None):
     if initial is not None:
         first = []
 
-    return chain(first, map(lambda x: func(x[1], x[0]), zip(a, b)))
+    return chain(first, starmap(func, zip(b,a)))
```
bbayles:
```suggestion
    return chain(first, starmap(func, zip(b, a)))
```

This will fix the flake8 error.

PR #345, opened 2019-11-26
Added efficient implementation of unique_powerset to recipes with test.

root-11:
Raymond Hettinger recommended me to submit this patch to you guys. I hope you like it?

# unique_powerset

The unique_powerset(iterable) returns the unique combinations of values
when presented with repeated values:

    unique_powerset([1,1,1,2,2,3]) --> [
        (), (1,), (1, 1), (1, 1, 1), (2,), (2, 2), (3,),
        (1, 2), (1, 1, 2), (1, 1, 1, 2),
        (1, 2, 2), (1, 1, 2, 2), (1, 1, 1, 2, 2),
        (1, 3), (1, 1, 3), (1, 1, 1, 3),
        (2, 3), (2, 2, 3), (1, 2, 3),
        (1, 1, 2, 3), (1, 1, 1, 2, 3),
        (1, 2, 2, 3), (1, 1, 2, 2, 3),
        (1, 1, 1, 2, 2, 3)
    ] # 24 records

This should be viewed in contrast to the powerset 
(see [1](https://docs.python.org/3/library/itertools.html#itertools-recipes))
which would generate repeated values:

    powerset([1,1,1,2,2,3]) --> [
      (),
      (1,), (1,), (1,),  # <-- duplicates.
      (2,), (2,),
      (3,),
      (1, 1), (1, 1), (1, 1),
      (1, 2), (1, 2), (1, 2), (1, 2), (1, 2), (1, 2),
      (1, 3), (1, 3), (1, 3),
      ... cut for brevity ...
      (1, 1, 1, 2, 2), (1, 1, 1, 2, 3), (1, 1, 1, 2, 3),
      (1, 1, 2, 2, 3), (1, 1, 2, 2, 3), (1, 1, 2, 2, 3)
    ] # 63 records.

The application of the unique_powerset is plenty, but most dominantly in
the knapsack problem where a number of items must be matched to the
capacity limit of the knapsack. Each combination of items determine the
utilisation of the knapsack and the best should be selected, yet items
of the same value can be substituted without further consideration

As the number of duplicate values grow, the number of redundant options
grows exponentially if using powerset.
In the example above the powerset generates 63 vs the 24 unique options
generated in the unique_powerset.

The assertion `set(powerset(iterable)) == unique_powerset(iterable)` must
always be true, and whilst the former method is available, powerset of
any iterables longer than 20 items, become intolerable except for the most
patient programmers.


bbayles:
Neat implementation. Could you fix the conflicts with master? I'll have a few more comments.

more_itertools/tests/test_recipes.py
```diff
@@ -344,6 +344,50 @@ def test_combinatorics(self):
         )
 
 
+class UniquePowersetTests(TestCase):
```
bbayles: Could you add a test for what's expected with an empty iterable?

more_itertools/tests/test_recipes.py
```diff
@@ -344,6 +344,50 @@ def test_combinatorics(self):
...
+    """ Tests for ``unique_powerset``"""
+
+    def test_unique_powerset(self):
+        elements = [1, 1, 1, 2, 2, 3]
+        pss = set(mi.powerset(elements))
+
+        self.assertEqual(len(pss), 24)
```
bbayles: I think we can do without this check here - we're checking the length of `ups` below, and that `pss` and `ups` are the same.

more_itertools/tests/test_recipes.py
root-11: done in [9eba55](https://github.com/root-11/more-itertools/commit/9eba5565496d59305c3578a4e242ad069373ba1b)

more_itertools/tests/test_recipes.py
root-11: done in [9eba55](https://github.com/root-11/more-itertools/commit/9eba5565496d59305c3578a4e242ad069373ba1b)

root-11:
> Neat implementation. Could you fix the conflicts with master? I'll have a few more comments.

I'm stuck on this. The content is identical on both forks, but git claims there's a conflict.

Any hint on how to resolve it? SOs [hints](https://stackoverflow.com/questions/31973318/github-cant-automatically-merge) took me nowhere.

bbayles:
Something like this should work:

```
git remote add upstream https://github.com/erikrose/more-itertools.git
git fetch --all
git merge upstream/master
git push origin HEAD
```

root-11:
This worked.

> git remote add upstream https://github.com/erikrose/more-itertools.git
> git fetch --all
> git merge upstream/master
> git push origin HEAD:master

more_itertools/recipes.py
```diff
@@ -346,6 +347,116 @@ def powerset(iterable):
...
+    for r in range(1, len(blocks) + 1):
+        for clusters in combinations(blocks, r):
+            # each 'cluster' is now an element from the powerset of the
+            # unique elements --> (1,) (2,) (3,) (1,2) (1,3) (2,3) (1,2,3)
+
+            # first we set indices for the accessing the first element in
+            # the clusters values:
+            c_index = [0 for _ in clusters]
+            # this allows us to increment each index in values of each block.
+            # Hereby c_index = [0,1,0] on the cluster (1,2,3) becomes [1,2,2,3].
```
bbayles:
```suggestion
            # Hereby c_index = [0,1,0] on the cluster (1,2,3) becomes [1,2,2,3]
```

more_itertools/recipes.py
```diff
@@ -346,6 +347,116 @@ def powerset(iterable):
...
+    yield ()
+    # first we summarize the iterable into blocks of identical values. Example:
+    # [1,1,1,2,2,3] -->
+    # d = {
+    #     1: [[1],[1,1],[1,1,1]],
+    #     2: [[2],[2,2]],
+    #     3: [[3]]
+    #     }
+    d = {i: iterable.count(i) for i in set(iterable)}
+    blocks = {i: [] for i in set(iterable)}
```
bbayles: Here's an issue I noticed: this exhausts `iterable` once on line 411. If it's an iterator and not a sequence, then the second time `set(iterable)` is attempted it will fail.

bbayles: We can outsource the creation of these dicts to the `collections` module:

```python
    blocks = defaultdict(list)
    for k, v in Counter(iterable).items():
        for i in range(1, v + 1):
            blocks[k].append([k] * i)
```

bbayles:
Some more requests:
* This isn't something brought over from the `itertools` docs, so it (and its tests) should go in `more.py` (and `test_more.py`).
* It probably makes more sense to call this `distinct_powerset`, so it goes along with `distinct_combinations` and `distinct_permutations`.
* Could you tell me the merits of this implementation over the one below? If it uses a better algorithm to generate distinct combinations, should we update that function instead?

Shorter implementation:

```python
def distinct_powerset(iterable):
    s = list(iterable)
    return chain.from_iterable(
        distinct_combinations(s, r) for r in range(len(s) + 1)
    )
```

Thanks!

root-11:
> Could you tell me the merits of this implementation over the one below?

I can confirm that the proposal:

```
def distinct_powerset(iterable):
    s = list(iterable)
    return chain.from_iterable(
        distinct_combinations(s, r) for r in range(len(s) + 1)
    )
```

produces the exact same result, and I'm very happy to see code used, rather than my original proposal. 

> If it uses a better algorithm to generate distinct combinations, should we update that function instead?

Yes!

> It probably makes more sense to call this distinct_powerset, so it goes along with distinct_combinations and distinct_permutations.

I agree. As the name `set` has been (ab)used in `powerset` (which technically only produces a set if the input items are unique), using **distinct** seems appropriate.

Good feedback. Thank you.

root-11:
I do apologize for reverting on my conclusion from earlier, but the runtime penalty of generating each combination followed by throwing the duplicates away in `unique_everseen` **is significant**, due to the factorial nature of combinations, where the **throw-away : retain** factor is ~3:1.

The analysis is simple:
```
print("len of iterable: time in milliseconds")
for source in range(10,30,1):
    iterable = list(range(source))
    start = time.time()
    c = 0
    for i in distinct_powerset(iterable):
        c+= 1
    t = time.time() - start

    start = time.time()
    c = 0
    for i in unique_powerset(iterable):
        c+= 1
    t2 = time.time() - start
    print(f"{len(iterable)}: distinct_powerset {round(t*1000)} | unique_powerset to {round(t2*1000)}")
```

The result:

```
len of iterable: time in milliseconds
10: distinct_powerset 12 | unique_powerset to 4
11: distinct_powerset 24 | unique_powerset to 9
12: distinct_powerset 53 | unique_powerset to 20
13: distinct_powerset 115 | unique_powerset to 40
14: distinct_powerset 242 | unique_powerset to 84
15: distinct_powerset 572 | unique_powerset to 171
16: distinct_powerset 1090 | unique_powerset to 344
17: distinct_powerset 2229 | unique_powerset to 708
18: distinct_powerset 4555 | unique_powerset to 1462
19: distinct_powerset 10034 | unique_powerset to 3158
20: distinct_powerset 20587 | unique_powerset to 6377
21: distinct_powerset 43650 | unique_powerset to 13370
22: distinct_powerset 89703 | unique_powerset to 27363
23: distinct_powerset 188352 | unique_powerset to 57026
24: distinct_powerset 394471 | unique_powerset to 117548
```

I really like the brevity of the code proposed as `distinct_powerset`, but as it is significantly slower I'd recommend to keep the code proposed earlier.

bbayles:
Hmm, I'm confused. 

> the runtime penalty of generating each combination followed by throwing the duplicates away in `unique_everseen`

My proposed implementation doesn't use `unique_everseen`; it uses `distinct_combinations`. That function does not generate and throw away duplicates.

It looks like your version of `unique_powerset`  is indeed faster than mine. But I'm wondering if we ought to take the "generate distinct combinations" portion out of your implementation, and replace the one in `distinct_combinations`. Does that make sense?

root-11:
I line 2642 of `more.py`, `unique_everseen` is used in `distinct_combinations` (recursively) on master.

My understanding of python is that each recursive stack call has its own memory allocation, so unique_everseen must repeat it's work at each recursive entry into distinct_combinations. I think that's where the duplicate work happens.

bbayles:
Got it - thanks! So can a modification of [this block](https://github.com/erikrose/more-itertools/pull/345/files#diff-61f057e7be1481445851e17b801f7ea5R419-R457) (with a fixed `r`) be used in place of the existing `distinct_combinations` implementation?

root-11:
Yes. I think so.

more_itertools/recipes.py
```diff
@@ -384,7 +384,7 @@ def unique_powerset(iterable):
 
     As the number of duplicate values grow, the number of redundant options
     grows exponentially if using powerset.
-    In the example above the powerset generates 63 vs the 23 unique options
+    In the example above the powerset generates 63 vs the 24 unique options
```
plazer1, line 390: `powerset` generates 64 subsets, not 63

bbayles:
Closing this one due to age.

I was still interested in this:
> It looks like your version of unique_powerset is indeed faster than mine. But I'm wondering if we ought to take the "generate distinct combinations" portion out of your implementation, and replace the one in distinct_combinations. Does that make sense?

But I think https://github.com/more-itertools/more-itertools/pull/483 has superseded that.

PR #350, opened 2019-12-02
add __init__.pyi stub Fixes #349

### LINKED ISSUE #349: mypy annotations don't work with root package

keisheiled:


keisheiled:
@erikrose perhaps it's worth adding mypy check to see if the types are valid with mypy.

(and merging them inline with the code)

keisheiled:
https://github.com/google/pytype/tree/master/pytype/tools/merge_pyi can merge the types back in

bbayles:
I'm willing to believe this fixes #349, but could you tell me why [this](https://github.com/erikrose/more-itertools/blob/master/.travis.yml#L19) automated check doesn't work?

We won't be putting the types in with the code.

keisheiled:
that check isn't checking anything because you have no typed defs,  `--disallow-untyped-defs` it will show that failure.

As far as I'm aware there's no way currently to get mypy to check code with separate annotations

graingert:
> that check isn't checking anything because you have no typed defs, `--disallow-untyped-defs` it will show that failure.
> 
> As far as I'm aware there's no way currently to get mypy to check code with separate annotations, could you explain to me why you do not want to apply the type annotations to the code?

https://github.com/python/mypy/pull/6695

correct, you have to merge the typed into the code for them to be checked

bbayles:
I'll merge this to solve the issue in #349 .

I'm declining to retrofit types onto the existing recipes for aesthetic reasons.

I've subscribed to the relevant mypy issues and will be watching to see when that project adds support for checking separate annotations.

Thanks!

keisheiled:
@bbayles can you make a release for this?

bbayles:
Yes, I will.

bbayles:
https://pypi.org/project/more-itertools/8.0.1/

keisheiled:
@bbayles thanks for the release! it still didn't work though as the .pyi file is missing from the whl

This should add it back in https://github.com/erikrose/more-itertools/pull/352

PR #356, opened 2019-12-15
Add nth_or_last

d-ryzhikov:


more_itertools/more.py
```diff
@@ -180,6 +181,23 @@ def last(iterable, default=_marker):
...
+        2
+        >>> nth_or_last([0, 1], 2)
+        1
+        >>> nth_or_last([], 0, 'some default')
+        'some default'
+
+    If *default* is not provided and there are no items in the iterable,
+    raise ``ValueError``.
+    """
+    return last(islice(iterable, n + 1), default=default)
```
pylang: Clever implementation!  `islice` yields the full elements even if `n` > `len(seq)`, so you are guaranteed to get the last element.  Nice.

PR #360, opened 2019-12-24
Adding Travis-Ci Support For Arm64

ossdev07:
Travis-CI has added support for ARM64. Added ARM64 job in Travis-CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
+arch:
+    - amd64
+    - arch64
 language: "python"
 
 python:
@@ -5,7 +8,12 @@ python:
     - "3.6"
     - "3.7"
     - "3.8"
-    - "pypy3"
+
+matrix:
+    include:
+        - arch: amd64
+          python: pypy3
+          env:
 
 install:
     - "pip install ."


bbayles:
Thanks! Do we need the blank `env` on line 16? Or can that be removed?

MSeifert04:
Just a minor point: This almost doubles the amount of CI (workers and time). Given that this package doesn't use C extensions it's probably fine to just run one Python test against arch64.

.travis.yml
```diff
@@ -1,11 +1,19 @@
+arch:
+    - amd64
+    - arch64
```
MSeifert04: The rest of the file adds a blank line between these definitions, so I think it would be consistent to add a blank line between arch and language as well.

.travis.yml
```diff
@@ -1,11 +1,19 @@
...
     - "3.6"
     - "3.7"
     - "3.8"
-    - "pypy3"
+
+matrix:
+    include:
+        - arch: amd64
+          python: pypy3
+          env:
```
MSeifert04: As noted by @bbayles this empty variable should not be needed.

PR #363, opened 2019-12-31
Make numeric_range an iterable

jferard:
The `numeric_range` function is meant to be

> An extension of the built-in ``range()`` function whose arguments can be any orderable numeric type.

But `range(...)` is an iterable, while `numeric_range(...)` is an iterator:

    >>> r = range(5)
    >>> list(r) + list(r)
    [0, 1, 2, 3, 4, 0, 1, 2, 3, 4]
    >>> from more_itertools import numeric_range
    >>> nr = numeric_range(5)
    >>> list(nr) + list(nr) # nr was exhausted
    [0, 1, 2, 3, 4]

I think that `numeric_range` should be an iterable. (Note that this semantic difference has some effects: `range` has an efficient implementation of `len`, `reversed`, `r[...]`, but those are features of an iterable, not an iterator. Hence we can't add these features to `numeric_range`.)

 This PR is a basic implementation. Main drawback: we need division and modulo. I don't know if there are numeric entities in Python that can't be divided (and would need a fallback).

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -1,5 +1,7 @@
 import warnings
 from collections import Counter, defaultdict, deque
+
+from collections import Counter, defaultdict, deque, abc
 from collections.abc import Sequence
 from functools import partial, wraps
 from heapq import merge, heapify, heapreplace, heappop
@@ -21,6 +23,7 @@ from math import exp, floor, log
 from operator import gt, itemgetter, lt, sub
 from random import random, randrange, uniform
 from operator import itemgetter, sub
+from operator import itemgetter, sub, gt, lt
 from sys import maxsize
 from time import monotonic
 
@@ -1611,7 +1614,7 @@ def groupby_transform(iterable, keyfunc=None, valuefunc=None):
     return ((k, map(valuefunc, g)) for k, g in res) if valuefunc else res
 
 
-class numeric_range:
+class numeric_range(abc.Sequence, abc.Hashable):
     """An extension of the built-in ``range()`` function whose arguments can
     be any orderable numeric type.
 
@@ -1677,7 +1680,7 @@ class numeric_range:
             self._start, self._stop, self._step = args
         else:
             if argc == 0:
-                err_msg = 'numeric_range expected 1 argument, got {}'
+                err_msg = 'numeric_range expected at least 1 argument, got {}'
             else:
                 err_msg = 'numeric_range expected at most 3 arguments, got {}'
             raise TypeError(err_msg.format(argc))
@@ -1720,7 +1723,8 @@ class numeric_range:
             return numeric_range(start, stop, step)
         else:
             raise TypeError(
-                "numeric range indices must be integers or slices, not str")
+                'numeric range indices must be '
+                'integers or slices, not {}'.format(type(key).__name__))
 
     def __hash__(self):
         if self:
@@ -1729,15 +1733,11 @@ class numeric_range:
             return self._EMPTY_HASH
 
     def __iter__(self):
-        elem = self._start + self._zero
+        values = (self._start + (self._step * n) for n in count())
         if self._growing:
-            while elem < self._stop:
-                yield elem
-                elem += self._step
+            return takewhile(partial(gt, self._stop), values)
         else:
-            while elem > self._stop:
-                yield elem
-                elem += self._step
+            return takewhile(partial(lt, self._stop), values)
 
     def __len__(self):
         if self._growing:
@@ -1753,9 +1753,9 @@ class numeric_range:
         if distance <= 0:
             return 0
 
-        r = distance % step
-        q = int(distance / step)
-        return q + int(r != self._zero)
+        assert distance > 0 and step > 0
+        q, r = divmod(distance, step)
+        return int(q) + int(r != self._zero)
 
     def __reduce__(self):
         return numeric_range, (self._start, self._stop, self._step)
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -1687,6 +1687,12 @@ class numeric_range:
             raise ValueError('numeric_range() arg 3 must not be zero')
         self._growing = self._step > self._zero
 
+    def __bool__(self):
+        if self._growing:
+            return self._start < self._stop
+        else:
+            return self._start > self._stop
+
     def __contains__(self, elem):
         if self._growing:
             if self._start <= elem < self._stop:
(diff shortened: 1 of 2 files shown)


more_itertools/more.py
```diff
@@ -1590,7 +1590,7 @@ def groupby_transform(iterable, keyfunc=None, valuefunc=None):
     return ((k, map(valuefunc, g)) for k, g in res) if valuefunc else res
 
 
-def numeric_range(*args):
+class numeric_range:
```
bbayles: Would it be possible inherit from `collections.abc.Iterable` and `collections.abc.Sequence`? `range()` objects are marked as instances of each. Could add a test for this too.

more_itertools/more.py
```diff
@@ -1634,35 +1634,148 @@ def numeric_range(*args):
...
+            self._stop, = args
+            self._start = type(self._stop)(0)
+            self._step = 1
+        elif argc == 2:
+            self._start, self._stop = args
+            self._step = 1
+        elif argc == 3:
+            self._start, self._stop, self._step = args
+        else:
+            if argc == 0:
```
bbayles: Could you just move this into the `if / elif` structure above?

more_itertools/more.py
```diff
@@ -1634,35 +1634,148 @@ def numeric_range(*args):
...
+            raise TypeError(
+                "numeric range indices must be integers or slices, not str")
+
+    def __hash__(self):
+        if self:
+            return hash((self._start, self._get_by_index(-1), self._step))
+        else:
+            return self._EMPTY_HASH
+
+    def __iter__(self):
```
bbayles: I think this could be re-worked.

First, doing repeated `+=` operations is going to introduce accumulating floating point error.

Second, I think `__iter__` should return `self` and `__next__` should return the next `_get_by_index` value (meaning, keep track of `n` like [here](https://github.com/erikrose/more-itertools/blob/c11195abbc3ca7408d77ac051712162b455b2306/more_itertools/more.py#L1660-L1665)).

more_itertools/more.py
```diff
@@ -1634,35 +1634,148 @@ def numeric_range(*args):
...
+            start = self._stop
+            stop = self._start
+            step = -self._step
+
+        distance = stop - start
+        if distance <= 0:
+            return 0
+
+        r = distance % step
+        q = int(distance / step)
```
bbayles: Why not `divmod` here?

more_itertools/more.py
```diff
@@ -1634,35 +1634,148 @@ def numeric_range(*args):
...
+        elif isinstance(key, slice):
+            start = self._start if key.start is None \
+                else self._get_by_index(key.start)
+            stop = self._stop if key.stop is None \
+                else self._get_by_index(key.stop)
+            step = self._step if key.step is None else key.step * self._step
+            return numeric_range(start, stop, step)
+        else:
+            raise TypeError(
+                "numeric range indices must be integers or slices, not str")
```
bbayles: `str` is hardcoded here - this should probably check the type of what was passed in, right?

more_itertools/more.py
jferard: Good idea, but inherit from  ̀collections.abc.Iterable` is unnecessary since a `Sequence` is already an `Iterable`. I will mark `numeric_range` as  ̀collections.abc.Sequence` and  ̀collections.abc.Hashable`.

more_itertools/more.py
jferard: You mean something like:

```
    elif argc == 0:
        raise TypeError('numeric_range expected at least 1 argument, got {}'.format(argc)
    else:
        raise TypeError('numeric_range expected at most 3 arguments, got {}'.format(argc)
```
?

(I added "at least" to make the message clearer).

more_itertools/more.py
jferard: > First, doing repeated += operations is going to introduce accumulating floating point error.

Totally right.

> Second, I think __iter__ should return self and __next__ should return the next _get_by_index value (meaning, keep track of n like here).

I have a serious doubt here: if `__iter__` returns `self` you have an iterator (that is also an iterable). The issue is that a `numeric_range` becomes stateful, whereas I think it should be stateless (iterable but not iterator) as the regular `range`. Related issues: what is the length of an stateful object? when are stateful objects equal? ...

more_itertools/more.py
jferard: Because I initially thought that `5.0 // 1.5` would return `3` and not `3.0`. The code was:

        r = distance % step
        q = distance // step

And then I added a `int` cast and removed the second slash. I'll use `divmod`. Note that there is a subtle difference:

```
>>> divmod(5.0, -1.5)
(-4.0, -1.0)
>>> (5.0//-1.5, 5.0%-1.5)
(-4.0, -1.0)
```

But:
```
>>> (int(5.0/-1.5), 5.0%-1.5)
(-3, -1.0)
```

We do not have this problem here because `distance` and `step` are both positive. I propose an `assert` to document this:

        assert distance > 0 and step > 0

more_itertools/more.py
jferard: My mistake!

more_itertools/more.py
bbayles: Yeah, exactly.

more_itertools/more.py
bbayles: Ah, `next(range(10))` doesn't work - missed that. Anyway, should avoid the floating point error by using a counter, like the current implementation.

more_itertools/more.py
bbayles: Thanks for the explanation. No need for the assert, but could add this as a comment.

more_itertools/more.py
jferard: Done.

more_itertools/more.py
jferard: Done.

more_itertools/more.py
jferard: Done.

more_itertools/more.py
jferard: Done.

more_itertools/more.py
jferard: Fixed.

more_itertools/more.py
```diff
@@ -1654,35 +1654,145 @@ def numeric_range(*args):
...
+    def count(self, value):
+        return int(value in self)
+
+    def index(self, value):
+        if self._growing:
+            if self._start <= value < self._stop:
+                q, r = divmod(value - self._start, self._step)
+                if r == self._zero:
+                    return int(q)
+        else:
```
bbayles: I don't think this branch is tested - could you add something that verifies this?

more_itertools/more.py
```diff
@@ -1654,35 +1654,145 @@ def numeric_range(*args):
...
+
+        distance = stop - start
+        if distance <= 0:
+            return 0
+
+        # distance > 0 and step > 0: regular euclidean division
+        q, r = divmod(distance, step)
+        return int(q) + int(r != self._zero)
+
+    def __reduce__(self):
```
bbayles: This method doesn't appear to be tested - could we add a pickle test?

more_itertools/more.py
jferard: Done.

more_itertools/more.py
jferard: Done.

jferard:
> A couple more requests here, plus one more: could you add some tests of the new methods with Decimal, Fraction, and datetime objects to see what breaks?

I added a lot of tests and fixed some issues. See commit message 37e6b6f for details.

bbayles:
Thanks for this!

PR #369, opened 2020-01-03
Version 8.1.0

bbayles:
This PR includes updates for version 8.1.0.

@MSeifert04, I hooked up https://github.com/c4urself/bump2version here. What do you think?


MSeifert04:
I know that ["cookiecutter-pylibrary" template](https://github.com/ionelmc/cookiecutter-pylibrary) uses bumpversion so that's probably a good fit. I always handle the version manually (probably not sensible but a habit I haven't changed yet), but I also don't have such an active library like `more-itertools` 😄

more_itertools/more.py
```diff
@@ -2783,19 +2782,27 @@ def _sample_weighted(iterable, k, weights):
 def sample(iterable, k, weights=None):
```
pylang: I would consider adding a `seed=None` parameter.  Then you can set a seed to get deterministic samples.

PR #372, opened 2020-01-09
Expand sliced to allow sliceable objects that don't use `bool` for emptiness

sswingle:
I was surprised when I tried to use `sliced` with a numpy array, and it failed.  The issue turned out to be that `sliced` calls `bool` on each slice to check if it's empty or not, so it knows when to stop.  By changing it to call `len` instead, we can use `sliced` with numpy arrays and other objects that don't use `bool` to indicate emptiness.

I've also added a test for `sliced` that uses a numpy array.  However this does add a dependency on numpy for our tests, so maybe we shouldn't keep this new test.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -1051,7 +1051,7 @@ def sliced(seq, n):
     For non-sliceable iterables, see :func:`chunked`.
 
     """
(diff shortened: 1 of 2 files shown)


bbayles:
Merged in #373. Thanks!

PR #374, opened 2020-01-12
Fix typing for always_iterable

blueyed:
This uses the typing from `isinstance`, which is used with this argument
(handling `None` before).


bbayles:
Thanks - do you have a reference for "typing from isinstance" handy? Also, what the error is currently being generated?

blueyed:
> Thanks - do you have a reference for "typing from isinstance" handy?

https://github.com/python/typeshed/blob/a06abc5dff2928b7111452fe9f44a922585ebde0/stdlib/2and3/builtins.pyi#L1203


> Also, what the error is currently being generated?

> src/_pytest/python_api.py:687: error: Argument 2 to "always_iterable" has incompatible type "Tuple[Type[type], Tuple[Type[bytes], Type[str]]]"; expected "Union[type, Tuple[type, ...], None]"

pytest uses `BASE_TYPE = (type, STRING_TYPES)`, where `STRING_TYPES` is a tuple already.

blueyed:
@bbayles would be great to have a new release for this, and the other typing fixes.

bbayles:
@blueyed , I've released this to PyPI.

PR #375, opened 2020-01-12
mypy: improve config for tests

blueyed:


bbayles:
Could you summarize why these changes are improvements?

blueyed:
I'm not too convinced myself, but:

1) `[mypy-more_itertools.tests.*]` appears to be wrong, since tests are in `tests` directly.
2) `ignore_errors` seems to be not helpful when you want to do type checking explicitly.
3) `disallow_untyped_defs = false` is used since tests are not typed.

Altogether this allows for `mypy tests`, which still results in a lot of errors, but e.g. showed https://github.com/erikrose/more-itertools/pull/377.

PR #376, opened 2020-01-12
Improve flake8 config

blueyed:


bbayles:
I may update this in a different branch. Thanks for the suggestion!

PR #378, opened 2020-01-26
Update and fix typing

ilai-deutel:
This PR updates the type stubs for new methods (`buckets.__iter__`, `numeric_range.*`) and fixes type hints for `sample` and `grouper`.


bbayles:
Thanks!

more_itertools/more.pyi
```diff
@@ -199,7 +204,10 @@ def sort_together(
 def unzip(iterable: Iterable[Sequence[_T]]) -> Tuple[Iterator[_T], ...]: ...
 def divide(n: int, iterable: Iterable[_T]) -> List[Iterator[_T]]: ...
 def always_iterable(
-    obj: object, base_type: Union[type, Tuple[Union[type, Tuple[Any, ...]], ...], None] = ...
+    obj: object,
+    base_type: Union[
+        type, Tuple[Union[type, Tuple[Any, ...]], ...], None
+    ] = ...,
```
blueyed: @ilai-deutel 
Since a lot of this seems to just about wrapping for line-length I've wanted to mention that it is typically OK for .pyi files to have long lines, and a special flake8 config can be used to handle that (https://github.com/erikrose/more-itertools/pull/376).
Also (larger) style changes are better off in a separate commit then also.
(just for info / my 2 cents)

ilai-deutel: I used [black](https://github.com/psf/black) to format the `.pyi` with the same parameters as I used when I initially created the `.pyi` files, which are the ones used for the `.py` files (`black -S -l 79` as in #306). [typeshed](https://github.com/python/typeshed) uses a line length of 130 for stub files and it was suggested to become the default in python/black#341, but the project maintainer decided otherwise.
79 characters for regular Python files and 130 for stub files sounds reasonable to me, what do you think @bbayles? Also, if `black` is enforced across the project, there could be `black` checks in the travis config in addition to `flake8`.

bbayles: I'm in favor of re-formatting everything with `black`, and enforcing it in Travis. I'll do that for a future release.

Issue #381, opened 2020-01-30
Add a maxsplit, reverse parameters to split_


jferard:
# maxsplit
more-itertools has various `split_` functions (`split_at`, `split_after`, `split_before`, `split_when`, `split_into`) but none of these functions has a way to stop the split after a given number of elements. Of course, we can take the first elements of a `split_...` iterator and swtich to the original iterator:

    >>> import more_itertools as mi, itertools
    >>> s = '0,1,2,3,4,5,6,7,8,9'
    >>> it1 = iter(s)
    >>> it2 = mi.split_after(it1, lambda x: x == ",")
    >>> list(itertools.islice(it2, 3)) # it *2*
    [['0', ','], ['1', ','], ['2', ',']]
    >>> list(it1) # it *1*
    ['3', ',', '4', ',', '5', ',', '6', ',', '7', ',', '8', ',', '9']

But it would be more convenient to write, as we do with [`str.split`](https://docs.python.org/3/library/stdtypes.html#str.split):

    >>> import more_itertools as mi, itertools
    >>> s = '0,1,2,3,4,5,6,7,8,9'
    >>> it2 = mi.split_after(s, lambda x: x == ",", maxsplit=3)
    >>> list(it2) # it *2*
    [['0', ','], ['1', ','], ['2', ','], ['3', ',', '4', ',', '5', ',', '6', ',', '7', ',', '8', ',', '9']]

(The last parameter could be an iterator to remain lazy).

# reverse
We could have another parameter, named `reverse` to mimic [`str.rsplit`](https://docs.python.org/3/library/stdtypes.html#str.rsplit):

    >>> it2 = mi.split_after(s, lambda x: x == ",", maxsplit=3, reverse=True)
    >>> list(it2) # it *2*
    [['0', ',', '1', ',', '2', ',', '3', ',', '4', ',', '5', ',', '6', ','], ['7', ','], ['8', ','], ['9']]

(Or create `rsplit_at`, `rsplit_after`, ... but that would be a little cumbersome)

Implementation should not be difficult.

bbayles:
I'd entertain a PR for the `maxsplit` additions, and a separate one for `reverse`.

jferard:
@bbayles Would you allow:
* a `keep_separator` parameter for `split_at` (useful for the implementation)
* the name `rsplit` for the parameter instead of `reverse` because `reverse` might be confusing
?

Examples (handwritten):

[...]

PR #382, opened 2020-02-01
Maxsplit parameter for "split_" functions

jferard:
See #381. 

The implementation is *very* straightforward, but may not be optimal: when maxplit reaches `0`, we could return a `list(...)` with the iterator. E.g.:

```
def split_at(iterable, pred, maxsplit=-1):
    if maxsplit == 0:
        yield list(iterable)
        return

    buf = []
    it = iter(iterable)
    for item in it:
        if pred(item) and maxsplit != 0:
            yield buf
            if maxsplit == 1: # we want a last chunk.
                yield list(it)
                return
            buf = []
            maxsplit -= 1
        else:
            buf.append(item)
    yield buf
```

It should be faster for very long iterators, but I didn't `timeit`. (All tests pass with this implementation.)

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -1081,10 +1081,18 @@ def split_at(iterable, pred, maxsplit=-1):
         >>> list(split_at(range(10), lambda n: n % 2 == 1, maxsplit=2))
         [[0], [2], [4, 5, 6, 7, 8, 9]]
     """
+    if maxsplit == 0:
+        yield list(iterable)
+        return
+
     buf = []
-    for item in iterable:
-        if pred(item) and maxsplit != 0:
+    it = iter(iterable)
+    for item in it:
+        if pred(item):
             yield buf
+            if maxsplit == 1:
+                yield list(it)
+                return
             buf = []
             maxsplit -= 1
         else:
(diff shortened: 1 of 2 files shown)


jferard:
Additional information. `split_at` is the naive implementation (current state of the PR) and `split_at1` is the more "sophisticated" implementation above:

```
from timeit import timeit as ti
s = 'abcdcba'+'a'*100
print(ti(lambda: list(split_at(s, lambda x: x == 'b', maxsplit=1))))
# 12.111127081000177
print(ti(lambda: list(split_at1(s, lambda x: x == 'b', maxsplit=1))))
# 1.421707098999832
```

If you agree, I will update the implementation.

more_itertools/more.py
```diff
@@ -1064,28 +1064,43 @@ def sliced(seq, n):
...
         >>> list(split_at(range(10), lambda n: n % 2 == 1))
         [[0], [2], [4], [6], [8], []]
+
+    At most *maxsplit* splits are done. If *maxsplit* is not specified or -1,
+    then there is no limit on the number of splits:
+
+        >>> list(split_at(range(10), lambda n: n % 2 == 1, maxsplit=2))
+        [[0], [2], [4, 5, 6, 7, 8, 9]]
     """
+    if maxsplit == 0:
```
bbayles: Could you make this `if not maxsplit` so we can catch `None` also?

jferard: I tested this:

```
>>> "1,2".split(",", maxsplit=0)
['1,2']
>>> "1,2".split(",", maxsplit=None)
TypeError: 'NoneType' object cannot be interpreted as an integer
```

I think we should stick to `str.split` behavior, and check the type of `maxsplit` like this:

```
>>> maxsplit = 0
>>> if not isinstance(maxsplit, int): raise TypeError("'{}' object cannot be interpreted as an integer".format(maxsplit.__class__.__name__))
... 
>>> maxsplit = None
>>> if not isinstance(maxsplit, int): raise TypeError("'{}' object cannot be interpreted as an integer".format(maxsplit.__class__.__name__))
... 
TypeError: 'NoneType' object cannot be interpreted as an integer
```

bbayles: Good call. Thanks!

PR #384, opened 2020-02-06
Typing: Maxsplit parameter for "split_" functions

ilai-deutel:
Small changes to `more.pyi`, following #382.


bbayles:
Many thanks!

PR #385, opened 2020-02-08
Rsplit

jferard:
See #381
There where different options: a) use `rlocate`, b) try to insert the `rsplit` logic directly in the code or c) merge the first chunks to keep only the last chunks. Options a) and b) had a lot of edges cases, so I decided to:

1. create all the chunks;
2. merge all chunks excepted `maxsplit`.

I added two helper functions.

(I used the parameter name `rsplit`, but it is easy to change this).

### DIFF
--- a/tests/test_more.py
+++ b/tests/test_more.py
@@ -27,6 +27,7 @@ from time import sleep
 from unittest import skipIf, TestCase
 
 import more_itertools as mi
+from more_itertools.more import _aggregate_slice
 
 
 def load_tests(loader, tests, ignore):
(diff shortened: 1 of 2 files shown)


bbayles:
Thanks! It will be a few days before I can review this thoroughly.

bbayles:
This is good work, so thanks for the PR.

The reversal logic isn't quite clear to me. I'll spend some more time with it, but I'm wondering if it might be better to just do the `keep_separators` option and introduce a helper function in a different PR.

tests/test_more.py
```diff
@@ -1112,6 +1113,55 @@ def test_seperators(self):
             self.comp_with_str_split(s, delim, maxsplit)
```
bbayles: Back on the earlier `keep_separator` PR, I should have asked this: your function does this:

```python
>>> from more_itertools import split_at
... 
... list(split_at('abccba', lambda x: x == 'a', keep_separator=True))
[[], ['a'], ['b', 'c', 'c', 'b'], ['a'], []]
```

Is that really what we want, the blanks at the ends?

jferard: Sorry for the late answer. The closest example I found is:

```
>>> import re
>>> re.split("(a)", 'abccba')
['', 'a', 'bccb', 'a', '']
```

I think it's a strong argument in favour of the current behavior.

bbayles:
I'm going back and forth about the reversal stuff - my problem, not yours! I'm going to pull your `keep_separator` part out and try to think about the `rsplit` parts independently.

bbayles:
I'm going to close this one due to age. The reversal stuff is a nice addition, but the complexity it adds to the implementation is more than I want to take on. Thanks!

PR #395, opened 2020-03-08
Update more.py

SergBobrovsky:
I decided to reinvent distinct_permutations on uniform **lazy** approch
For r != None twice faster then previos code, else a bit slowly.
And now I can do this:
```
for t in islice(distinct_permutations('abcdefghijklmnopqrstuvwxyz'), 0, None, 10 ** 6): # !!!
    print(''.join(t))
```

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -546,71 +546,83 @@ def one(iterable, too_short=None, too_long=None):
 
 
 def distinct_permutations(iterable, r=None):
-    """Yield successive distinct permutations of the elements in *iterable*.
-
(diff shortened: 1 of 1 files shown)


pylang:
@SergBobrovsky it is great to see you're excited about `more_itertools`.  You contributions are appreciated.  

Is there a way you can submit one PR per change?  Over the past day, you have 13 PRs spread across two tool contributions.  You see, GitHub sends emails each time a PR is made, which is now getting spammy.   I'm pretty sure there's a way to update an open PR, which might be what you want.

PR #399, opened 2020-03-08
distinct_permutations

SergBobrovsky:
I decided to reinvent distinct_permutations on uniform lazy approch
For r != None twice faster then previos code, else a bit slowly.
And now I can do this:

for t in islice(distinct_permutations('abcdefghijklmnopqrstuvwxyz'), 0, None, 10 ** 6): # no memory eating
    print(''.join(t))

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -546,71 +546,85 @@ def one(iterable, too_short=None, too_long=None):
 
 
 def distinct_permutations(iterable, r=None):
-    """Yield successive distinct permutations of the elements in *iterable*.
+    """
+    Generates unique permutations of the elements *iterable*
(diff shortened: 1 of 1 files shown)


SergBobrovsky:
> I'd prefer to keep the docstring that was there previously, however.

As You wish. Anyway, docstring of itertools.permutations does not contains any formula - and I hope, this is an argument )

tommyod:
@SergBobrovsky . Curious spectator here. What's the reference for this algorithm?

SergBobrovsky:
@tommyod 
>I'm curious - what's the reference for this algorithm?
[https://en.wikipedia.org/wiki/Permutation#Generation_in_lexicographic_order](url)

PR #407, opened 2020-03-09
refactor windowed part 3

SergBobrovsky:
smaller, faster, successfully pass tests )
flake8 is evil (

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -672,8 +672,8 @@ def unique_to_each(*iterables):
     return [list(filter(uniques.__contains__, it)) for it in pool]
 
 
-def windowed(seq, n, fillvalue=None, step=1):
-    """Return a sliding window of width *n* over the given iterable.
+def windowed(iterable, width, fillvalue=None, step=1):
+    """Return a sliding window of *width* over the given iterable.
 
         >>> all_windows = windowed([1, 2, 3, 4, 5], 3)
         >>> list(all_windows)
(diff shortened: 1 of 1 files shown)


more_itertools/more.py
```diff
@@ -672,8 +672,8 @@ def unique_to_each(*iterables):
     return [list(filter(uniques.__contains__, it)) for it in pool]
 
 
-def windowed(seq, n, fillvalue=None, step=1):
-    """Return a sliding window of width *n* over the given iterable.
+def windowed(iterable, width, fillvalue=None, step=1):
```
bbayles: I'd prefer if the parameter names didn't change.

more_itertools/more.py
```diff
@@ -694,43 +694,28 @@ def windowed(seq, n, fillvalue=None, step=1):
...
+                yield window
+        if 0 < len(window) < width:
+            yield chain(window, (fillvalue,) * (width - len(window)))
+        elif 0 < i < min(step, width):
+            window.extend((fillvalue,) * i)
+            yield window
+
+    if width >= 0 < step:
+        return map(tuple, helper(**locals()) if width else ((),))
+    raise ValueError('width must be >= 0\nstep must be > 0')
```
bbayles: This loses the specific error messages of the original implementation.

more_itertools/more.py
```diff
@@ -694,43 +694,28 @@ def windowed(seq, n, fillvalue=None, step=1):
...
+                i = step
+                yield window
+        if 0 < len(window) < width:
+            yield chain(window, (fillvalue,) * (width - len(window)))
+        elif 0 < i < min(step, width):
+            window.extend((fillvalue,) * i)
+            yield window
+
+    if width >= 0 < step:
+        return map(tuple, helper(**locals()) if width else ((),))
```
bbayles: Is it really necessary to inline `locals` here? What's the performance impact?

more_itertools/more.py
SergBobrovsky: I do not mind that you take only the helper() and implement it in the previous version of the windowed, keeping in it everything that is dear to you.

more_itertools/more.py
SergBobrovsky: The discussion of **n** versus **width** seems like a waste of time for me, if you believe that somewhere there is a dude who passes a named parameter **n** to the windowed() - your right to save this argument name.

more_itertools/more.py
SergBobrovsky: This is my jamb, I was tired there ) Only **step** is critical, so changing the signature of the helper to a **helper(step)** will be right.

SergBobrovsky:
I don't mind to you improve pr as you wish.

PR #415, opened 2020-03-31
Add `zip_equal`

frankier:
`zip_equal` zips together equal length iterators or else throws an UnequalIterablesException. A custom exception class is used to make it easier to catch if need be.

This functionality has been discussed a few times in the past. See https://mail.python.org/pipermail/python-ideas/2018-July/052462.html for example. Here's another implementation in the wild: https://github.com/bioinformed/vgraph/blob/5866e6a686ce98e8d02f6f97f6ab025c89f123b9/vgraph/iterstuff.py#L343 . A quick google reveals that zip_equal is more common than zip_exact.


frankier:
Also here: https://stackoverflow.com/questions/32954486/zip-iterators-asserting-for-equal-length-in-python

more_itertools/more.py
```diff
@@ -1411,6 +1413,41 @@ def stagger(iterable, offsets=(-1, 0, 1), longest=False, fillvalue=None):
...
+
+def zip_equal(*iterables, names=None):
+    """``zip`` the input *iterables* together, but throw an
+    ``UnequalIterablesException`` if any of the *iterables* terminate before
+    the others. The exception will give the indices which terminated too early.
+    Alternatively *names* can be passed and it will give the name corresponding
+    to the early terminating indices.
+
+    """
+    sentinel = object()
```
bbayles, line 1440: We already have `_marker` in this module; we can use that instead of a new sentinel.

more_itertools/more.py
```diff
@@ -1411,6 +1413,41 @@ def stagger(iterable, offsets=(-1, 0, 1), longest=False, fillvalue=None):
...
+            depleted = "Columns: {}".format(
+                ", ".join(names[idx] for idx in idxs)
+            )
+        super().__init__(
+            "Iterables have different lengths."
+            "{} depleted before rest.".format(depleted)
+        )
+
+
+def zip_equal(*iterables, names=None):
```
bbayles, line 1432: I think the `names` feature is overkill. Would you mind removing it?

more_itertools/more.py
```diff
@@ -1411,6 +1413,41 @@ def stagger(iterable, offsets=(-1, 0, 1), longest=False, fillvalue=None):
...
+    the others. The exception will give the indices which terminated too early.
+    Alternatively *names* can be passed and it will give the name corresponding
+    to the early terminating indices.
+
+    """
+    sentinel = object()
+    if names is not None and len(iterables) != len(names):
+        raise ValueError("Number of iterable and names didn't match")
+
+    for combo in zip_longest(*iterables, fillvalue=sentinel):
```
bbayles, line 1444:
```suggestion
    for combo in zip_longest(*iterables, fillvalue=_marker):
```

more_itertools/more.py
```diff
@@ -1411,6 +1413,41 @@ def stagger(iterable, offsets=(-1, 0, 1), longest=False, fillvalue=None):
...
+        finished_idxs = [idx for idx, c in enumerate(combo) if c is sentinel]
```
bbayles, line 1445: Like `names`, I think this is probably too much - let's just go with `if _marker in combo`?

more_itertools/more.py
```diff
@@ -1411,6 +1413,41 @@ def stagger(iterable, offsets=(-1, 0, 1), longest=False, fillvalue=None):
...
+        if finished_idxs:
+            raise UnequalIterablesException(finished_idxs, names)
```
bbayles, line 1447: I'm OK with this being its own `Exception` class, but I think it might be overengineered right now. How about just:

```python
class UnequalIterablesError(ValueError):
    pass
```

bbayles: `Error` seems to be more of a standard suffix than `Exception` - see [here](https://docs.python.org/3/library/exceptions.html#exception-hierarchy).

more_itertools/more.py
frankier: I have used "is" instead since it can not be overloaded as __eq__ can, so we know for sure that we actually have _marker rather than a tricky/misbehaving object (_marker is only used with is elsewhere also) but I've removed the rest

bbayles:
Many thanks!

PR #417, opened 2020-04-03
Adding an aggregator function

nayyarv:
When writing code, I tend to prefer using `yield's` and `yield from` since it means I don't have to maintain my own list, append and then join. This does push the aggregation to the calling code which can be annoying when it's the same thing every time and pushes the complexity/boilerplate to the caller. So I've written a wrapper that does this for me.

for example

```python

def hello_world():
    string_list = []
    string_list.append("Hello")
    string_list.append("World")
    return " ".join(string_list)

```
is now
```python
@mi.aggregator(" ".join)
def string_yielder():
    yield "Hello"
    yield "World!"
```

This also comes in use when I have to hit a db/external resource in a paginated way but need to return a `np.array` or `pd.DataFrame`, again I can just yield and then the aggregator does the work, keeping the code clean and very readable.

I'm not in love with the name and more than happy to update the doc + any other things if you have some suggestions - this is sort of a minimal PR to gauge interest before I put in more effort to get it release worthy.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -27,6 +27,7 @@ from .recipes import consume, flatten, powerset, take, unique_everseen
 
 __all__ = [
     'adjacent',
+    "aggregator",
     'always_iterable',
     'always_reversible',
     'bucket',
@@ -99,6 +100,31 @@ __all__ = [
 ]
 
 _marker = object()
+
+def aggregator(aggregator_func):
+    """Combines the output of a generator with *aggregator_func*
+
+        >>> @aggregator(" ".join)
+        ... def string_yielder():
+        ...     yield "Hello"
+        ...     yield "World!"
+        "Hello World!
+
+    Allows the usage of yields in the code without needing to convert to list allowing
+    for more expressive and cleaner code instead of ``list.append``
+    This is particularly nice when interfacing with code that recieves data in a
+    paginated fashion but you'd like to return a ``pd.DataFrame`` or ``np.array``
+
+    This can also be chained
+    """
+    def wrapper(func):
+        @wraps(func)
+        def inner(*args, **kwargs):
+            return aggregator_func(func(*args, **kwargs))
+
+        return inner
+
+    return wrapper
 
 
 def chunked(iterable, n):
--- a/tests/test_more.py
+++ b/tests/test_more.py
@@ -3573,3 +3573,65 @@ class SampleTests(TestCase):
 
         # The observed largest difference in 10,000 simulations was 4.337999
         self.assertTrue(difference_in_means < 4.4)
+
+
+class AggregatorTests(TestCase):
+
+    def test_strjoin(self):
+        @mi.aggregator(" ".join)
+        def hello_world():
+            yield "Hello"
+            yield "World"
+
+        self.assertEqual(hello_world(), "Hello World")
+
+    def test_str_join_on_list(self):
+        @mi.aggregator(" ".join)
+        def hello_world():
+            return ["Hello", "World"]
+
+        self.assertEqual(hello_world(), "Hello World")
+
+    def test_listify(self):
+        @mi.aggregator(list)
+        def hello_world():
+            yield "Hello"
+            yield "World"
+
+        self.assertEqual(hello_world(), ["Hello", "World"])
+
+    def test_chained(self):
+        # test, not a recommended way to do things!
+        @mi.aggregator(tuple)
+        @mi.aggregator(chain.from_iterable)
+        def number_lists():
+            yield [1, 2, 3]
+            yield [4, 5, 6]
+
+        self.assertEqual(number_lists(), (1, 2, 3, 4, 5, 6))
+
+    def test_chained_2(self):
+        # indicates an apply_each might be useful?
+        @mi.aggregator(".".join)
+        @mi.aggregator(lambda x: map(str, x))
+        @mi.aggregator(chain.from_iterable)
+        def number_lists():
+            yield [1, 2, 3]
+            yield [4, 5, 6]
+
+        self.assertEqual(number_lists(), "1.2.3.4.5.6")
+
+    def test_conversion(self):
+        # test, not a recommended way to do things!
+        @mi.aggregator(set)
+        def number_set():
+            return range(7)
+
+        self.assertEqual(number_set(), {0, 1, 2, 3, 4, 5, 6})
+
+    def test_reduction(self):
+        @mi.aggregator(sum)
+        def summer():
+            return range(7)
+
+        self.assertEqual(summer(), 21)


nayyarv:
Firstly I think `aggregator` is better named `collector`, but I'm open to naming suggestions.

Secondly, I've been thinking about `make_decorator`, how it relates to this PR and why I haven't used it in my code, while mostly glossing over it

Firstly the doc suggests doing something like this

```python
listifier = mi.make_decorator(list)

@listifier()
def hw():
    yield "Hello"
    yield "World"

print(hw())
['Hello', 'World']
```

However, I don't like adding things to the global scope, I'd rather define everything on one line. This means I have the info asap and don't have odd style mismatches where I define a decorator on one line and use it instantly - it looks more like a variable definition than a function definition and as such looks a bit strange in most style guides. For example, I would almost never do this with `functools.partial` in global scope since I think it would end up confusing more than helpful, especially for just a single use.

Instead I'd probably do this

```python
@mi.make_decorator(list)()
def hw():
    yield "Hello"
    yield "World"

# or canonically
@mi.make_decorator(islice)(1, 10, 2)
def nums():
    yield from range(20)
```

This syntax is awkward, the `make_decorator` is not as obvious as `listifier` was and doesn't really give a reader any intuition. Additionally, for the `listifier` example, I didn't add the extra function call `()` since it wasn't common decorator syntax, it didn't work, so I gave up and moved on.

Furthermore, while `make_decorator` is powerful and works for nearly every case (with the result_index kwarg), it doesn't provide for a simple entry point for most usage and I suspect it's not heavily used (I asked a friend if there was an aggregator/collector decorator and he suggested this PR).

As a result, I think it makes sense to provide a simple entry point that might take care of most usage of the `make_decorator` while leaving `make_decorator` around for any users that need to use it.

```python
@mi.collector(list)
def hw():
    yield "Hello"
    yield "World"

@mi.collector(islice, 1, 10, 2)
def nums():
    yield from range(20)

@mi.collector(sum)
@mi.collector(islice, 1, 10, 2)
def nums():
    yield from range(20)
```

I do feel like this approach is better - the name `collector` instantly tells me that this function is trying to collect the output of `hw` into some other data structure, and this syntax is similar to `functools.partial` which I feel anyone mucking around with decorators should be familiar with. It's also simple enough for most use cases (as the tests in this PR should show) and I can guarantee at least 1 user who'd sprinkle this around quite a lot in their code (I think this use case is good enough for the standard library tbh).

bbayles:
Thanks for the writeup. I'm inclined not to merge this as-is - `make_decorator` can do this, as you say.

I would welcome:
* An example in the `make_decorator` docstring that shows how to do the `''.join` trick
* A top level function called `str_join` that is just `make_decorator(''.join)`

If you're up for that, I would merge either or both of those.

nayyarv:
As a library, providing an easy entry point to the code is important and I think `make_decorator` fails by providing code examples that contravene most style guides and being unintuitive enough that I took multiple reads of the doc to work out that `make_decorator` could do what I wanted. 
I did a very informal search for [`more_itertools language:Python make_decorator`](https://github.com/search?q=more_itertools+language%3APython+make_decorator&type=Code) on github and it appears that nearly every result is just someone who's committed their venv to the repo (as opposed to something like `flatten` which has a lot of uses). This matches with the anecdotal evidence that two heavy users of this library didn't realise how to use `make_decorator` to provide the functionality of `collector`

I think the value of `collector` is this ease of use and intuitiveness rather than additional functionality. Conceptually it is similar to the relationship between `subprocess.run` vs `subprocess.Popen`. I use `run` unless I must use `Popen` and similarly I'd use `collector` instead of `make_decorator` when possible. Constraining it to accept just a single function should give it enough functionality while stilling retaining readability and simplicity while still allowing for the usage of lambdas for more complex uses

I can see the argument that this potentially creates two ways of doing things and to that I respond
1. I believe `make_decorator` is probably rarely used and providing a better API to encourage the code to be used is better than leaving it in since `make_decorator` was already there. FWIW  `collector` with a lambda can replace `make_decorator` in every way and if we were to limit it to one, I feel like `make_decorator` should be the one removed.
2. Multiple ways of doing things is ok. As `run` vs `Popen` has already shown, or how `requests` is used more than `urllib`, the easier to use code will always be used the most. 


Finally let me provide a few more usage examples of how I currently use the `collector` decorator that go beyond my string usage (which was a minimal example), which should hopefully suggest that `str_join` is already insufficient.

```python
@mi.collector(pd.DataFrame)
@mi.collector(mi.flatten)
def query_to_datafrmae(db, query):
    token = db.make_query(query)
    while token:
        token, data = db.get_data(token)
        yield data

@mi.collector("\n| ".join)
def aws_cw_insight(group, regexes, predicates):
    yield "fields @message"
    yield f"filter @logstream = {group}"
    for reg in regexes:
        yield f"filter @message like /{reg}/"
    yield from predicates

@mi.collector(np.array)
@mi.collector(lambda x: random.sample(x, len(x))) # hella gratuitous, but a power example
def get_data():
    with open("data_file", 'r') as f:
        for line in f:
            yield process(line)

```

bbayles:
Well, thanks for this writeup too. My position is still this: I'll accept a PR to add a documentation example and/or a top-level function for `str_join`.

nayyarv:
So, as it turns out, I gave some bad info. Sorry about that, I did a copy paste without checking, going off of my python mental model

So the above `make_decorator` example I gave is actually invalid

```python
@mi.make_decorator(list)()
def hw():
    yield "Hello"
    yield "World"

File "<ipython-input-7-e1b4220dd9fd>", line 1
    @mi.make_decorator(list)()
                            ^
SyntaxError: invalid syntax
```

Decorators can only be variables or functions as far as I can tell. I did some experiments (things like trying `@Test().hi` and ast decompositions) but I really don't have any thing I can point to regarding this syntax. I suspect that since decorators are applied lazily, something that creates state on definition wouldn't be allowed. 

Hence if we wanted to define a decorator on one line, we would need this PR.

pylang:
@nayyarv I appreciate your contribution.  I learned a few things, including more people are using `make_decorator`!  \o/   GG @bbayles!

@nayyarv I would comment your points, but I looks like the decision is made on this particular suggestion.  I like your insight.  I'm curious to see novel tools you might create in the future.

nayyarv:
@pylang, I'm not sure how you came to that conclusion re `make_decorator` being used.

This improved search with the `NOT numeric_range` added on : [`more_itertools make_decorator language:Python NOT numeric_range`](https://github.com/search?q=more_itertools+make_decorator+language%3APython+NOT+numeric_range&type=Code) in order to try and exclude situations where the more_itertools package was committed. It returns only 3 results. (I tried to do NOT filename:more.py and variants, but it doesn't appear to work).

As opposed to [`more_itertools flatten language:Python not numeric_range`](https://github.com/search?q=more_itertools+flatten+language%3APython+not+numeric_range&type=Code) which has 6500+ results.

Anyways to sum up my views:
>  An example in the make_decorator docstring that shows how to do the ''.join trick. 

This is hardly a trick, `"".join` is a function and all usual rules apply. I don't think this adds much, especially to a long document and especially to a function with almost no users.

> A top level function called `str_join` that is just make_decorator(''.join) 

I would rather we didn't pollute the namespace with such specialised functions. Given that this formulation wouldn't even allow a choice of which string to join on, I don't think it'll be used at all and shouldn't be in a library like this as it sets a precedent of specialised decorators (why not `@to_set` or `@listify`).

bbayles:
Closing this out - thanks again for the discussion.

PR #422, opened 2020-05-03
Check lengths of iterables in zip_equal for better speed and error message

alexmojaki:
Hi!

This PR checks (when possible) the lengths of the iterables passed to `zip_equal`. This allows it to give a more informative error message for differing lengths, allowing easier debugging. It also means it can just defer to `zip` without iterating in Python space, which is significantly faster. Here's a benchmark comparing zip_equal in master with zip:

 ```python
import timeit
from collections import deque
from more_itertools import zip_equal

x1 = list(range(1000))
x2 = list(range(1000, 2000))


def my_timeit(stmt):
    print(timeit.repeat(stmt, globals=globals(), number=10000, repeat=3))


def consume(iterator):
    deque(iterator, maxlen=0)


my_timeit("consume(zip(x1, x2))")
my_timeit("consume(zip_equal(x1, x2))")
```

Output:

```
[0.151994071, 0.16300702100000003, 0.14365304500000003]
[2.094478611, 2.043564243, 2.0213984409999997]
```

So zip is about 13 times faster, and this PR takes advantage of that for the common case of zipping lists or other sequences.

### DIFF
--- a/tests/test_more.py
+++ b/tests/test_more.py
@@ -1678,21 +1678,35 @@ class ZipEqualTest(TestCase):
     def test_unequal(self):
         short = [0, 1]
         long = [2, 3, 4]
-        with self.assertRaises(mi.UnequalIterablesError, msg="Iterables have different lengths: [2, 3]"):
+        with self.assertRaises(
+                mi.UnequalIterablesError,
+                msg="Iterables have different lengths: [2, 3]",
+        ):
             list(mi.zip_equal(short, long))
 
-        with self.assertRaises(mi.UnequalIterablesError, msg="Iterables have different lengths: [3, 2]"):
+        with self.assertRaises(
+                mi.UnequalIterablesError,
+                msg="Iterables have different lengths: [3, 2]",
+        ):
             list(mi.zip_equal(long, short))
 
-        with self.assertRaises(mi.UnequalIterablesError, msg="Iterables have different lengths"):
+        with self.assertRaises(
+                mi.UnequalIterablesError,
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by bbayles** [comment]
Thanks for the PR. I like the idea of using normal `zip` if it's possible. I'd rather not have the "length <= 10" thing however - is this trying to avoid doing `min` and `max` on the input iterables? It's a bit odd because the code already runs through them multiple times.

**Comment by alexmojaki** [comment]
It's to show the exact lengths at each position in the common case while also accounting for the possibility of having a million iterables.

**Comment by bbayles** [comment]
Could we avoid an unnecessary pass through the iterables by just using `len` directly, without the `Sized` check, and switching strategies when we hit one that raises an exception?

**Comment by bbayles** [comment]
Thanks! I was able to get this down to a single pass of the iterables and still always report the mismatching lengths.

File: tests/test_more.py [hunk 1]
```diff
@@ -1668,20 +1668,46 @@ class ZipEqualTest(TestCase):
...
-        with self.assertRaises(mi.UnequalIterablesError):
+        with self.assertRaises(
+                mi.UnequalIterablesError,
+                msg="Iterables have different lengths: [3, 2]",
+        ):
             list(mi.zip_equal(long, short))
 
+        with self.assertRaises(
+                mi.UnequalIterablesError,
+                msg="Iterables have different lengths",
```
bbayles, line 1695: FYI - this doesn't do what you intended. That is, it doesn't check the exception's argument for this string.

Issue #428, opened 2020-05-21
islice_extended with actual slices


cool-RR:
Here's an idea I suggested to python-ideas, and I was told it could be a better fit here. How about instead of this: 

`islice_extended(iterable, 7, 20)`

We'll just have: 

`islice_extended(iterable)[7:20]`

Advantages:
1. More familiar slicing syntax.
2. No need to awkwardly use `None` when you're interested in just specifying the end of the slice without specifying the start, i.e. `islice_extended(x)[10:]` instead of `islice_extended(x, 10, None)`.
3. Doesn't require breaking backwards compatibility.

bbayles:
Great idea. I'll mess around with an implementation.

bbayles:
See what you think of [this implementation](https://github.com/more-itertools/more-itertools/compare/islice-with-slices?expand=1)?

Link to the original discussion:
https://mail.python.org/archives/list/python-ideas@python.org/thread/QVTGZD6USSC34D4IJG76UPKZRXBBB4MM/

cool-RR:
Wow, this is some fast service ;)

The implementation looks like a good start. One thing I'd change is that I think `islice_extended(x)[a:b]` should return an `islice_extended` object, so that could be sliced again with the same syntax. The `_islice_extended` logic could just be be on the class.

Also: I don't like the name `islice_extended`, I think it's too long. I don't think it'll be bad to call it `islice`, despite the fact it clashes with itertools. I think it's fair as it can be used as a drop-in replacement for `itertools.islice`.

Since I see you're receptive to new features, I wonder whether there are any tools in my [cute_iter_tools](https://github.com/cool-RR/python_toolbox/blob/master/python_toolbox/cute_iter_tools.py) that could be combined into `more-itertools`.

bbayles:
Re: the name, I'm not keen on changing it. I recommend `from more_itertools import islice_extended as islice` for using it as a replacement for `itertools.islice`.

I like your idea of making the slices slice-able. See what you think of https://github.com/more-itertools/more-itertools/pull/429 ?

cool-RR:
Regarding the name, I understand. 

Regarding the implementation, it looks good to me.

PR #429, opened 2020-05-23
Slice support for islice_extended 

bbayles:
Re: https://github.com/more-itertools/more-itertools/issues/428, this PR adds support for slicing to `islice_extended`.

You can still use the 2 argument (`iterable, stop`), 3 argument (`iterable, start, stop`), and 4 argument forms (`iterable, start, stop step`). And the returned objects are themselves slice-able.

```python
... from more_itertools import islice_extended
... 
... it = map(str, count())  # Infinite iterable
... it = islice_extended(it)  # Now it's slice-able
... it = it[10:]  # Start from index 10
... it = it[:10]  # End after 10 items
... it = it[::2]  # Skip every other item
... list(it)
0: ['10', '12', '14', '16', '18']
```


cool-RR:
Woohoo!

PR #430, opened 2020-05-27
Improve tests for islice_extended

cool-RR:


tests/test_more.py
```diff
@@ -2668,7 +2668,20 @@ def test_slicing(self):
         first_slice = mi.islice_extended(iterable)[10:]
         second_slice = mi.islice_extended(first_slice)[:10]
         third_slice = mi.islice_extended(second_slice)[::2]
-        self.assertEqual(list(third_slice), ['10', '12', '14', '16', '18'])
+        self.assertSequenceEqual(list(third_slice),
```
bbayles, line 2671: This change seems unnecessary - could you revert it?

tests/test_more.py
```diff
@@ -2668,7 +2668,20 @@ def test_slicing(self):
...
+    def test_slicing_extensive(self):
+        iterable = range(10)
+        options = (None, 1, 2, 7, -1)
+        for start, stop, step in product(options, options, options):
+            sliced_tuple_0 = tuple(mi.islice_extended(iterable)
+                                   [start:stop:step])
+            sliced_tuple_1 = tuple(mi.islice_extended(iterable, start, stop,
+                                                      step))
+            sliced_range = tuple(iterable[start:stop:step])
+            self.assertSequenceEqual(sliced_tuple_0, sliced_range)
```
bbayles, line 2683: Could you make these `assertEqual`? From [the docs](https://docs.python.org/3/library/unittest.html#unittest.TestCase.addTypeEqualityFunc):

> The list of type-specific methods automatically used by assertEqual() are summarized in the following table. Note that it’s usually not necessary to invoke these methods directly.
> ...
> assertSequenceEqual(a, b)

tests/test_more.py
```diff
@@ -2668,7 +2668,20 @@ def test_slicing(self):
...
         second_slice = mi.islice_extended(first_slice)[:10]
         third_slice = mi.islice_extended(second_slice)[::2]
-        self.assertEqual(list(third_slice), ['10', '12', '14', '16', '18'])
+        self.assertSequenceEqual(list(third_slice),
+                                 ['10', '12', '14', '16', '18'])
+
+    def test_slicing_extensive(self):
+        iterable = range(10)
+        options = (None, 1, 2, 7, -1)
+        for start, stop, step in product(options, options, options):
```
bbayles, line 2677: Could you use `self.subTest` here, as the [test above](https://github.com/more-itertools/more-itertools/pull/430/files#diff-eb9b02233f873d0b01c09f7c2a396636R2657) does?

PR #432, opened 2020-05-29
Add mark_first() and mark_last()

kalekundert:
These functions yield values from an iterable, adding a boolean that indicates whether the current value is the first/last in the iterable.  An example makes this pretty clear:

```python
>>> list(mi.mark_first('ABC'))
[(True, 'A'), (False, 'B'), (False, 'C')]
>>> list(mi.mark_last('ABC'))
[(False, 'A'), (False, 'B'), (True, 'C')]
```

I think these functions are generally useful because it's pretty common to want to do something different on the first or the last iteration through an iterable.  While it's not particularly hard to identify the first/last iteration manually, these functions make the intent of the code much more clear.


bbayles:
Thanks for the PR.

Could you give me an example of how `mark_first` might be used that's shorter or easier than `enumerate`?

```python
for i, item in enumerate(iterable):
    if item == 0:
        do_first(item)
    else:
        do_rest(item)
```

I'll grant that there might be a stronger case for `mark_last`, though I probably would encourage something like:

```python
it = islice_extended(iterable, 0, -1)
for item in it:
    do_rest(item)
do_last(next(iterable)
```

kalekundert:
I won't try to argue that `mark_first()` is substantially better than `enumerate()`.  It just felt incomplete to have `mark_last()` and not `mark_first()`, so I put it in.

I do think that `mark_last()` is far better than the `islice_extended()` syntax, though.  The big reason is that the last/non-last cases often do a lot of the same things, and having an `is_last` variable makes this much simpler to implement.  For example, I think the snippet below is a pretty plausible use case that couldn't easily be written using `islice_extended()` (without refactoring the loop body into a handful of helper functions):

```python
for is_last, item in mark_last(iterable):
    some()
    lines()
    of()
    code()

    if is_last:
        do_something()
    else:
        do_something_else()

    more()
    lines()
    of()
    code()
```

It may also be worth pointing out that [for-loops in Jinja](https://jinja.palletsprojects.com/en/2.11.x/templates/\#for) have a feature similar to this.

bbayles:
I take your point about the complex loop. Also, my `islice_extended` trick doesn't actually work, as the last item is consumed by the cache for the negative index.

So I like something like `mark_last`, but as you say that seems to imply a `mark_first`, and that one seems superfluous. How about a single wrapper that does both?

```python
from more_itertools import spy

def mark_ends(iterable):
    it = iter(iterable)
    head, it = spy(it, 3)
    size = len(head)
    if size == 0:
        return
    elif size == 1:
        yield head[0], True, True
        return
    elif size == 2:
        yield head[0], True, False
        yield head[1], False, True
        return

    yield next(it), True, False
    pairs = pairwise(it)
    for a, b in pairs:
        yield a, False, False
    yield b, False, True
```

kalekundert:
I like that, especially since I'm sure there would be times where it would be useful to have both markers.  The only thing I'd do differently would be to order the return values like so: `(is_first, is_last, item)`.  That seems to be more of the convention for "augment"-style iterators, e.g. `enumerate()`, `count_cycle()`, etc.

I'll try to update the PR (and fix the tests) tomorrow.

bbayles:
Want to take a shot at adding type stubs? That's what Travis is complaining about.

Something like:

```
def mark_ends(iterable: Iterable[_T]) -> Iterator[Tuple[bool, bool, _T]]: ...
```

README.rst
```diff
@@ -42,6 +42,8 @@ Python iterables.
 | Augmenting             | `count_cycle <https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.count_cycle>`_,                                                                                                                |
 |                        | `intersperse <https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.intersperse>`_,                                                                                                                |
 |                        | `padded <https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.padded>`_,                                                                                                                          |
+|                        | `mark_first <https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.mark_first>`_,                                                                                                                  |
```
bbayles, line 45: Should update these for the combined function also.

docs/api.rst
```diff
@@ -82,6 +82,8 @@ These tools yield items from an iterable, plus additional data.
 .. autofunction:: count_cycle
 .. autofunction:: intersperse
 .. autofunction:: padded
+.. autofunction:: mark_first
```
bbayles, line 85: Here as well.

kalekundert:
I'm not sure what to do about the `black` error.  I tried running `black` on the files it's complaining about, but it introduces tons of changes beyond the code I worked on.  I'm also not sure why Travis is even running `black`, since it doesn't seem to be anywhere in `.travis.yml`...

bbayles:
Looks like you branched off before #431 was in master. I merged your changes. Many thanks!

Issue #307, opened 2019-08-29
Problematic combination of `consecutive_groups` and `last`


kriomant:
```
>>> from more_itertools import consecutive_groups, last
>>> list(map(list, consecutive_groups([1,2,10,11])))
[[1, 2], [10, 11]]
>>> list(last(consecutive_groups([1,2,10,11])))
[11]
```

bbayles:
Ah, tricky. This is similar to the behavior of [groupby](https://docs.python.org/3/library/itertools.html#itertools.groupby).

Its docs note:
> The returned group is itself an iterator that shares the underlying iterable with `groupby()`. Because the source is shared, when the `groupby()` object is advanced, the previous group is no longer visible. So, if that data is needed later, it should be stored as a list...

Using `last()` is like consuming the returned iterable out of order, so we see some strange behavior. I will add a note to the docs.

bbayles:
Added in PR #38. Thanks!

PR #440, opened 2020-06-20
Add groupfunc to groupby_transform

bbayles:
This PR adds the *groupfunc* argument to `groupby_transform`.

The docs for `groupby` have this caveat:
> The returned group is itself an iterator that shares the underlying iterable with groupby(). Because the source is shared, when the groupby() object is advanced, the previous group is no longer visible. So, if that data is needed later, it should be stored as a list.

This has come up here before, e.g. [issue 307](https://github.com/more-itertools/more-itertools/issues/307#issuecomment-526356782).

The `groupby` argument makes it easy to do that `list`-ifying, plus whatever else you might want to do.


bbayles:
This does become somewhat close to `map_reduce`:

```python
>>> from more_itertools import map_reduce
... list(map_reduce(range(50), lambda k: 10 * (k // 10), lambda v: v + 1, sum).items())
[(0, 55), (10, 155), (20, 255), (30, 355), (40, 455)]
```

```python
>>> from more_itertools import groupby_transform
... list(groupby_transform(range(50), lambda k: 10 * (k // 10), lambda v: v + 1, sum))
[(0, 55), (10, 155), (20, 255), (30, 355), (40, 455)]
```

But order matters for `groupby_transform`; it's more efficient when the input is sorted.

Issue #441, opened 2020-06-24
Callback to iterator utility: cb_to_iter


frankier:
Thought I would ask about this one here since it's something I keep using, and is completely to do with iterators but I'm not sure if it might be a bit controversial/out of scope.

Some libraries (e.g. HDF5py, lxml) do iteration via callbacks. It seems to be particularly common in fairly straight forward wrappers of C/C++ APIs. The idea of converting it to an iterator has come up before https://stackoverflow.com/questions/9968592/turn-functions-with-a-callback-into-python-generators . I myself have used this repeatedly (I have been copy pasting it between projects) so that I can get something out I can start applying iterator combinators on.

Here is the straw man. 

```
class AbortThread(BaseException):
    pass


def cb_to_iter(f):
    def iter(*args, **kwargs):
        from threading import Thread
        from queue import Queue

        q = Queue()
        job_done = object()
        abort = False

        def cb(*cb_args, **cb_kwargs):
            if cb_kwargs:
                cb_args = cb_args + (cb_kwargs,)
            if len(cb_args) == 1:
                q.put(cb_args[0])
            else:
                q.put(cb_args)
            q.join()
            if abort:
                raise AbortThread

        def task(*args, **kwargs):
            try:
                f(*(args + (cb,)), **kwargs)
                q.put(job_done)
            except AbortThread:
                pass

        thread = Thread(target=task, args=args, kwargs=kwargs)
        thread.start()

        try:
            while True:
                next_item = q.get(True)
                if next_item is job_done:
                    break
                yield next_item
                q.task_done()
        finally:
            abort = True
            try:
                q.task_done()
            except ValueError:
                pass
            thread.join(5)

    return iter
```

One big question mark here is: is it okay to start spawning threads? The main justification has to be that it's the only way to achieve this, and this is the only way to start applying iterator combinators when faced with this type of API short of just collecting everything into a list.

It would also be possible to use an async iterator, but that's a whole different kettle of fish.

bbayles:
Definitely a useful thing, though the way one has to do this is quite... un-aesthetic. Can you paste some examples of your code in action, e.g. with the scipy `fmin` from SO?

The `thread.join(5)` in the `finally` is unfortunate, but I suppose it's either that or risk leaking resources?

bbayles:
Here's my take:

<details><summary>callback_iter</summary>

```python
def callback_iter(func, yield_result=False, wait_seconds=0.1):
[...]

PR #444, opened 2020-06-27
Add callback_iter wrapper

bbayles:
Re: Issue #441, this PR adds `callback_iter`. It wraps functions that use callbacks for processing intermediate results.

@frankier, let me know what you think? I'll leave some comments here to explain the design.

---

__Before__:

```python
>>> from timeit import Timer
... from time import sleep
... 
... all_items = []
... callback = lambda *args, **kwargs: all_items.append((args, kwargs))
... timer = Timer(stmt='import time; time.sleep(0.01)')
... timer.autorange(callback=callback)
... all_items
[((1, 0.010457161999966047), {}),
 ((2, 0.020233963999999105), {}),
 ((5, 0.050860954999961905), {}),
 ((10, 0.10240066400001524), {}),
 ((20, 0.20350842099992406), {})]
```

__After__:

```python
>>> from timeit import Timer
... from more_itertools import callback_iter
... 
... timer = Timer(stmt='import time; time.sleep(0.01)')
... with callback_iter(timer.autorange) as it:
...     all_items = list(it)
... 
... all_items
[((1, 0.010082001999990098), {}),
 ((2, 0.02018048600007205), {}),
 ((5, 0.051050625000016225), {}),
 ((10, 0.10201740200000131), {}),
 ((20, 0.2054319899998518), {})]
```


more_itertools/more.py
```diff
@@ -3240,3 +3245,131 @@ def is_sorted(iterable, key=None, reverse=False):
...
+        if not self.done:
+            raise RuntimeError('Function has not yet completed')
+
+        return self._future.result()
+
+    def _reader(self):
+        q = Queue()
+
+        def callback(*args, **kwargs):
+            if self._aborted:
```
bbayles, line 3344: `self._aborted` gets set in when the `with` block exits ([here](https://github.com/more-itertools/more-itertools/pull/444/files#diff-a7f69885eda5d871c519f326bf410e62R3318)), which should allow for callbacks to be aborted.

```python
>>> from timeit import Timer
... from more_itertools import callback_iter
... 
... timer = Timer(stmt='import time; time.sleep(0.01)')
... with callback_iter(timer.autorange) as it:
...     all_items = next(it)
... 
... it.result
# ...
# more_itertools.more.AbortThread: canceled by user
```

more_itertools/more.py
```diff
@@ -3240,3 +3245,131 @@ def is_sorted(iterable, key=None, reverse=False):
...
+            try:
+                item = q.get(timeout=self._wait_seconds)
+            except Empty:
+                pass
+            else:
+                q.task_done()
+                yield item
+
+            if self._future.done():
+                break
```
bbayles, line 3363: I think this should handle partial consumption gracefully. Suppose you only consume one item, i.e. execution never gets past the yield on line 3360.

If the thread running the target function completes before the `with` block exits, then the queue will have items in it, but it will go out of scope.

If the thread running the target function isn't complete when the `with` block exits, it will hit the `AbortThread` exception above.

more_itertools/more.py
```diff
@@ -3240,3 +3245,131 @@ def is_sorted(iterable, key=None, reverse=False):
...
+
+        remaining = []
+        while True:
+            try:
+                item = q.get_nowait()
```
bbayles, line 3368: Since we only break out of the above loop when the target thread is finished, there should be no reason to wait on items in the queue - everything that could have been produced has already been produced.

more_itertools/more.py
```diff
@@ -3240,3 +3245,131 @@ def is_sorted(iterable, key=None, reverse=False):
...
+            except Empty:
```
bbayles, line 3369: As above, the target thread is finished by the time we get here - so when the queue is empty, it's empty for good, and we can break out.

more_itertools/more.py
```diff
@@ -3240,3 +3245,131 @@ def is_sorted(iterable, key=None, reverse=False):
...
+                break
+            else:
+                q.task_done()
+                remaining.append(item)
```
bbayles, line 3373: We could yield these as we encounter them, but we wouldn't empty the queue in the "partial consumption of results" case.

Issue #453, opened 2020-08-06
last() can try to use reversed()


brianmaissy:
Hey, first of all, I'm new new to more-itertools, and wanted to say thanks! I'm currently in the process of migrating a codebase to use this instead of some of its own hand-rolled utils.

What follows is a suggestion for a possible improvement, based on behavior that one of those utils supported. If you like the idea I'm happy to open a PR.

If I call `last()` on an iterable which doesn't support indexing but does have a `__reversed__()` method, the entire iterable will be consumed. Is there a reason `last()` doesn't try to call `reversed()`?

Here's a toy example of the type of iterable which would benefit from that behavior:

```
class ReversibleSlowRange:
    def __init__(self, start, stop, step=1):
        self.start = start
        self.stop = stop
        self.step = step
        self.current = start

    def __iter__(self):
        return self

    def __next__(self):
        if self.current == self.stop:
            raise StopIteration()
        time.sleep(1)
        result = self.current
        self.current += self.step
        return result

    def __reversed__(self):
        return self.__class__(self.stop-self.step, self.start-self.step, -self.step)
```

I don't think this is relevant for any of the builtins or types from the standard library, and there might not be too many real-world examples of data structures like this, so if this is rejected for falling into the category of "bloat", I understand. Just thought I would suggest it in case it might be useful to somebody.

brianmaissy:
For what it's worth, as of python 3.8 dicts fall into this category, so if someone is doing `last(some_dict)` in order to get the last key it would be faster if `last()` used `reversed()`. Still probably an uncommon use case, but the fact that it's relevant to a builtin type might make the difference.

bbayles:
What do you think of [this implementation](https://github.com/more-itertools/more-itertools/compare/last-reversed?expand=1)?

One issue I ran into on my machine: due to [Python issue 38525](https://bugs.python.org/issue38525), version 3.8.0 will crash when trying to reverse an empty `dict`. That was fixed in [3.8.1](https://docs.python.org/3.8/whatsnew/changelog.html#python-3-8-1-release-candidate-1), but unfortunately Ubuntu packages 3.8.0 in [18.04](https://packages.ubuntu.com/bionic-updates/python3.8).

brianmaissy:
The implementation looks good.

[...]

PR #454, opened 2020-08-10
Let last() use reversed()

bbayles:
This PR updates `last()` such that it:
* Tries to access iterable's last (`[-1]`) item directly
*  If that doesn't work, tries to reverse the iterable and return its first item
* If that doesn't work, feeds the iterable into a length-1 deque and returns the only item

Closes https://github.com/more-itertools/more-itertools/issues/453.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -1,7 +1,7 @@
 import warnings
 
 from collections import Counter, defaultdict, deque, abc
-from collections.abc import Reversible, Sequence
+from collections.abc import Sequence
 from concurrent.futures import ThreadPoolExecutor
 from functools import partial, wraps
 from heapq import merge, heapify, heapreplace, heappop
@@ -23,7 +23,7 @@ from math import exp, floor, log
 from queue import Empty, Queue
 from random import random, randrange, uniform
 from operator import itemgetter, sub, gt, lt
-from sys import maxsize
+from sys import hexversion, maxsize
 from time import monotonic
 
 from .recipes import (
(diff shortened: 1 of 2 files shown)


more_itertools/more.py
```diff
@@ -180,18 +180,19 @@ def last(iterable, default=_marker):
     raise ``ValueError``.
     """
     try:
-        try:
-            # Try to access the last item directly
+        if isinstance(iterable, Sequence):
```
brianmaissy, line 183: it might be safer and more robust to stick with the try/except approach, in case someone has a strange object that supports `getitem` but not `len`. Might also want to test this case.

more_itertools/more.py
```diff
@@ -180,18 +180,19 @@ def last(iterable, default=_marker):
...
             return iterable[-1]
-        except (TypeError, AttributeError, KeyError):
-            # If not slice-able, iterate entirely using length-1 deque
-            return deque(iterable, maxlen=1)[0]
-    except IndexError as e:  # If the iterable was empty
+        # Work around https://bugs.python.org/issue38525
+        elif hasattr(iterable, '__reversed__') and (hexversion != 0x030800F0):
```
brianmaissy, line 186: same here, I would try/except rather than checking `hasattr`

more_itertools/more.py
bbayles: I don't like the try/except approach, because of the `__getitem__(-1)` ambiguity - do I mean the sequence index or the mapping key? So this was intentional - non-sequences fall back to the `consume all but the last item` behavior.

more_itertools/more.py
bbayles: I would have done `isinstance(iterable, Reversible)`, but `Reversible` isn't in Python 3.5. But is [effectively](https://github.com/python/cpython/blob/3.8/Lib/_collections_abc.py#L305) a check for a `__reversed__` method.

more_itertools/more.py
brianmaissy: ok sounds good. I considered getting rid of the special case for sequences, because `reversed()` also works on sequences that don't have a `__reversed__` method (and thus removing a special case), but in any case we would have to deal with the 3.8.0 bug first, so the common case would still have to go through a preliminary check anyway.

PR #455, opened 2020-08-16
Added pitchfork partitioning

MarcinKonowalczyk:
I found myself in need of this functionality recently, and I couldn't find an appropriate `more_itertools` solution, so I thought I'll submit what I've come up with.

The idea is to be able to partition a sequence into three segments: the prefix, the middle part (which, in lack of a better name I've called `hay`) and the suffix. `pitchfork` yields all such splits. From the doctoring, this is a pitchfork partition of width two:

    >>> for p in pitchforked('ABCDEF', 2):
    ...     print(*map(lambda x: ''.join(x),p))
     AB CDEF
    A BC DEF
    AB CD EF
    ABC DE F
    ABCD EF 

The reason `windowed` doesn't cut it, is because I want to be able to construct the entire sequence from the parts on each iteration.

I've done my best to add it to the documentation correctly, and add tests (which all pass).

Hope you like it. 🙂

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -62,7 +62,6 @@ These tools yield windows of items from an iterable.
 .. autofunction:: substrings
 .. autofunction:: substrings_indexes
 .. autofunction:: stagger
-.. autofunction:: pitchforked
 
 ----
 
@@ -198,6 +197,7 @@ These tools yield combinatorial arrangements of items from iterables.
 .. autofunction:: circular_shifts
 .. autofunction:: partitions
 .. autofunction:: set_partitions
+.. autofunction:: pitchforked
 
 ----
 
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -62,6 +62,7 @@ These tools yield windows of items from an iterable.
 .. autofunction:: substrings
 .. autofunction:: substrings_indexes
 .. autofunction:: stagger
+.. autofunction:: pitchforked
 
 ----
 
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -110,6 +110,7 @@ __all__ = [
     'UnequalIterablesError',
     'zip_equal',
     'zip_offset',
+    'pitchforked'
 ]
 
 _marker = object()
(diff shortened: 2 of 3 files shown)


bbayles:
Thanks for the writeup. The implementation is surprisingly short.

I'd be inclined to merge this if: (1) we can articulate a problem this is suited to solving and (2) name the function in a way people looking to solve it would think of. Any Stack Overflow links that might be relevant?

MarcinKonowalczyk:
### Re: 1)
I found myself in need this functionality for the language I'm writing. I have the program in the form of an abstract (binary) syntax tree, represented as a tuple of nodes. The nodes can be strings (leaves) or tuples (subtrees). I wanted to go through each node (or each pair, or triple) and construct a new tree based on some alteration of these parts. (See [this file](https://github.com/MarcinKonowalczyk/psll-lang/blob/master/psll.py) under the "optimisation" section.)

I recognise this is an esoteric usage, but I believe this is the crux of the problem this function would be trying to solve: _"Partition a sequence in a more controlled manner than `partitioned`, and for each such partition reconstruct the original sequence with some alteration"_ What comes to mind is:
- Exchange each pair of neighbouring rows in a grid (maybe in a matrix, as part of some linear algebra algorithm or proof)
- Find English words which are also valid when a pair of neighbouring letters is swapped around
```
from english_words import english_words_set

for word in english_words_set:
    if len(word)>2: # No short words
        for p,h,s in pitchforked(word,2):
            if h[0]!=h[1]: # No double letters
                new_word = ''.join(p) + h[1] + h[0] + ''.join(s)
                new_word = new_word.lower()
                if new_word in english_words_set:
                    print(word,new_word)
```
(Some interesting ones are: 'egocentric' <-> 'geocentric', 'altitude' <-> 'latitude', 'cup' <-> 'cpu' and 'electro' <-> 'elector'
- Something like that for finding and replacing a sequence in a string (if one is allergic to regex):
```
string = "Quick brown fox jumped over a lazy fox"
a, b = "fox", "box"
while True:
    for j,(p,h,s) in enumerate(pitchforked(string,len(a))):
        if ''.join(h)==a:
            print(f"Found a match for '{a}' at position {j}")
            string = ''.join(p) + b + ''.join(s)
            break # Start again
    else: # No matches
        break # Break out of the for loop
print(string)
```

###  Re: 2)

Yeah, I know... 😅 It was originally called `partitioned` until I realised that is already a thing in `more_itertools`. Honestly, I think this function should be called `partitioned` and the one that is currently in the package should be something like `every_partition`, but that kind of name change would be very very bad given how many people are using `more_itertools`. 

`pitchforked` was something a friend suggested having heard the description of what the function does. Essentially the logic was that it splits a sequence into three sections, like a pitchfork would a bay of hay (given sufficient imagination I guess). The other idea I had was: `trisect`...?

bbayles:
Thanks for the explanation. What do you think of this implementation and name?

```python
def windowed_complete(iterable, n):
    """
    Yield ``(beginning, middle, end)`` tuples, where:
    * Each ``middle`` has *n* items from *iterable*
    * Each ``beginning`` has the items before the ones in ``middle``
    * Each ``end`` has the items after the ones in ``middle``

    >>> iterable = range(7)
    >>> n = 3
    >>> for beginning, middle, end in windowed_complete(iterable, n):
    ...     print(beginning, middle, end)
    () (0, 1, 2) (3, 4, 5, 6)
    (0,) (1, 2, 3) (4, 5, 6)
    (0, 1) (2, 3, 4) (5, 6)
    (0, 1, 2) (3, 4, 5) (6,)
    (0, 1, 2, 3) (4, 5, 6) ()

    """
    seq = tuple(iterable)
    size = len(seq)

    for i in range(size - n + 1):
        beginning = seq[:i]
        middle = seq[i: i + n]
        end = seq[i + n:]
        yield beginning, middle, end
```

MarcinKonowalczyk:
@bbayles, I like it! I've done some rudimentary `cProfile` testing and this implementation is about 2.7 times faster than the duque one. I think it's still missing the checks on the value of `n` though, because for `n<0` and `n>len(iterable)` the output is nonsense. I'm ok with the name. It makes much more of a clear link to `windowed` which I think is a good thing.

I've made those changes and will push a commit in a moment. I quite like the `every_partition` example, so I kept that (just made it much shorter, but hopefully still clear).

A bit of an unfortunate thing which I've noticed is that `n > size' check occurs *after* the iterable has been consumed. I don't see a clear way around it though, and I think that exiting with a clear error is the best thing. I' have made a note of this behaviour in the doctoring.

bbayles:
Great, glad this performs well. I'll merge this shortly, and will probably make a few changes to the docstring. I like your all-lengths example, but it's a little esoteric. Thanks!

Issue #456, opened 2020-08-18
New feature proposal: all_unique


brianmaissy:
I frequently find myself using the following idiom to check if all the elements of a collection are unique:

```python
if len(set(elements)) == len(elements):
    ...
```

Things that bother me about this:
1. repetitive idiom, but seems too small to extract to a utility function
2. doesn't support general iterables
3. inefficient, because it doesn't fail fast

I was wondering if a generalized `all_unique` function useful addition to this library? It would be an elegant complement to `all_equal`. There are many possible implementations, but I was imagining something along these lines (very similar to the implementation of `unique_everseen`):

```python
def all_unique(iterable, key=None):
    seenset = set()
    seenlist = []

    for element in map(key, iterable) if key else iterable:
        try:
            if element in seenset:
                return False
            else:
                seenset.add(element)
        except TypeError:
            if element in seenlist:
                return False
            else:
                seenlist.append(element)

    return True
```

What do you think? If you like it I'm happy to open a PR.

bbayles:
I dig it. Could you go ahead and make a PR?

brianmaissy:
sorry for the delay, just got around to it

brianmaissy:
resolved

PR #466, opened 2020-09-21
add new all_unique itertool

brianmaissy:
Implements #456


more_itertools/more.py
```diff
@@ -3447,3 +3448,40 @@ def windowed_complete(iterable, n):
...
+    Sequences with a mix of hashable and unhashable items can be used, but the
+    function will be slower for unhashable items. This is similar to
+    :func:`unique_everseen`, see details there.
+    """
+    seenset = set()
+    seenset_add = seenset.add
+    seenlist = []
+    seenlist_add = seenlist.append
+    for element in map(key, iterable) if key else iterable:
+        try:
```
jtwool, line 3477: Do we want to use `if isinstance(element, Hashable)` instead of the `try`/`except`?
Would require importing `Hashable` from `collections.abc`.

more_itertools/more.py
```diff
@@ -3447,3 +3448,40 @@ def windowed_complete(iterable, n):
...
+        >>> all_unique('ABCb')
+        True
+        >>> all_unique('ABCb', str.lower)
+        False
+
+    Sequences with a mix of hashable and unhashable items can be used, but the
+    function will be slower for unhashable items. This is similar to
+    :func:`unique_everseen`, see details there.
+    """
+    seenset = set()
```
jtwool, line 3472: To make the reasoning more clear, I'd favor using the names `seen_hashable` (`seenset`) and `seen_unhashable` (`seenlist`) instead.

more_itertools/more.py
```diff
@@ -3447,3 +3448,40 @@ def windowed_complete(iterable, n):
...
+    seenset_add = seenset.add
+    seenlist = []
+    seenlist_add = seenlist.append
+    for element in map(key, iterable) if key else iterable:
+        try:
+            if element in seenset:
+                return False
+            else:
```
jtwool, line 3480: We can remove these `else` statements at the end because there are `return` statements inside the `if` blocks.

jtwool (commented):
Three comments for clarity/flow.

more_itertools/more.py
brianmaissy: I copied the implementation from `unique_everseen`. Do we maybe want to ensure that it works the same way? If not I don't mind changing it to the hashable check

more_itertools/more.py
brianmaissy: again, copied the implementation. do you think the improvement in clarity outweighs the benefit of the implementation being as similar as possible to `unique_everseen`?

more_itertools/more.py
brianmaissy: ok

more_itertools/more.py
jtwool: Hmm. Maybe the suggested changes make more sense for subsequent PR that changes the implementation of both functions?

bbayles (approved):

PR #470, opened 2020-10-11
Fix nth product

N8Brooks:
This is to:
- Format with black
- Fix mypy failures
- Fix imports in recipes.py for nth_product
- Fix unit tests for nth_product

There are two extra commits and a few discarded revisions. I was checking how the various checks did. 
The unique_powerset got picked up in the diff. I'm not sure why.


bbayles:
I like the function. Perhaps you could submit this as a clean PR, though? Check out master, create a branch, and then drop your new code in. Thanks!

N8Brooks:
Thanks! Will do!

N8Brooks:
Updated!

bbayles:
Merged in 1361a19d86db5f9f1c8fb73b90fb8f28678a4364

PR #471, opened 2020-10-25
Add combinatoric caclulations and nth_permutation

N8Brooks:
Copies of math combinatoric calculations
  - comb
  - perm
  - prod
  - These could be take from the math library in python 3.8, but would not work in 3.6 or 3.7
  - math.prod is ~2x as fast as more_itertools.prod while comb and perm are only slightly faster
  - These could probably be removed whenever support for 3.6 and 3.7 is dropped and then the math ones could be used

 Defined combinatoric calculations
  - comb_w_r - length of combinations with replacement
  - perm_w_r - length of permutations with replacement
  - pwr - Length of a powerset with n items

nth_permutation
 - Retrives permutation n of an iterable with r items
 - It translates the index to factorial base which is then translated to the appropriate permutation using the pool.
 - Uses more_itertools.perm
 - The return function is not ideal. I think it is O(n * r). With a b-list or similar data structure it could be O(sqrt(n) * r)
 - itertools.permutations uses r=None, but since index should come as the final argument (like nth_combination), the r=None was replaced with just r for nth_permutation.


bbayles (changes requested):
Many thanks for the PR. I'm inclined to take `nth_permutation` and `prod` for this library and to defer the rest. Could you modify this PR with those changes?

N8Brooks:
I think that's a good idea. What do you think about dropping prod as well? The math library has had it since 3.8 and it's much faster than using reduce(operator.mul). Whenever more itertools drops support for 3.7 I could go back and change these functions to use the math counterpart (math.prod for nth_product and math.perm for nth_permutation).

bbayles:
Oh, I was hoping you'd keep `prod` - I use that all the time and I'm tired of using the `reduce(mul...` construct.

This library has had some backports for older Python versions in the past, e.g. `collate` (for `heapq.merge`) and `accumulate` (for `itertools.accumulate`). I've usually removed them once all the supported Python versions have the corresponding function.

Thanks!

N8Brooks:
Yeah, I'll just add it as a separate pull request. I think that would be better so that there is one change per pull request :-D

pylang:
I've sat on this recommendation some time, but these contributions seem related.


`itertools` has `combinations()` , `combinations_with_replacement()` and `permulations()`.  It lacks an explicit `permutations_with_replacement()`.  It turns out `product()` is the same thing.

```python
def permutations_with_replacement(iter_, r):
    yield from itertools.product(iter_, repeat=r)
```

This wasn't obvious to me at first, and frankly, i'm not sure I'll always remember this fact.  So I figured there might be some value in making an explicit function.  Although I realize this is just a wrapper to an existing tool, but there are [existing wrappers](https://more-itertools.readthedocs.io/en/stable/_modules/more_itertools/more.html#lstrip) in the library.

N8Brooks:
Cool! I didn’t know they were the same thing!

bbayles (approved):

pylang:
Ok, a clarification on `permutations_with_repetition`.  A reason why we don't see this term very much may be because the this is not an official term.  Technically, it's not even a [permutation according to discrete mathematics](https://en.wikipedia.org/wiki/Permutation), requiring a "bijection to itself" (one-to-one mapping of unique elements).  Since it's not a real permutation, I can see why it may be excluded from `itertools`. 

Alternative names are [`n-tuple`](https://en.wikipedia.org/wiki/Permutation#Permutations_with_repetition), but i think that could be confusing too.  It's sad that the name doesn't work out since these kind of questions occur a lot.

- [cartesian product](https://stackoverflow.com/questions/52192855/cartesian-product-of-two-lists-in-python)
- [latest](https://stackoverflow.com/questions/64987503/using-iterators-to-get-a-particular-sequence-of-list-from-1-2-3)
- [general search: permutation with replacement python](https://stackoverflow.com/search?q=permutation+with+replacement+python)

N8Brooks:
Okay, that clears it up, thanks!

Issue #468, opened 2020-10-09
Another iterator suggestion.


jenstroeger:
Iterators are very useful for comprehensions, and today I came across a problem that just happened to be solveable with [prepend()](https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.prepend) but there is no function that returns an iterator for the more general case.

I have a bunch of values, some scalar and some iterable, and I need to iterate over the scalars and the first level of the iterables:
```python
def didum(*args):
    for value in args:
        if isinstance(value, (str, bytes,)):
            yield value
            continue
        if not value:
            yield value
            continue
        try:
            yield from value
        except TypeError:
            yield value
```
This one’s useful because it doesn’t create temporary data structures (like prepend() does). You can then call this function to iterate over its argument values:
```
>>> list(didum(1, 2, "foo"))
[1, 2, 'foo']
>>> list(didum(b"bar", [1, 2, 3], 4))
[b'bar', 1, 2, 3, 4]
>>> list(didum([], "", 0, None))
[[], '', 0, None]
>>> list(didum())
[]
```
The existing prepend() can then be implemented as:
```
def prepend(value, iterator):
    return didum(value, iterator)  # Don’t create a [value] list.
```
I guess it’s a combination of generalizing prepend() and [flatten()](https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.flatten). I have no name for it though 😉

If this is interesting for you, I’d be happy to submit a PR. If this exists already, I guess I’ve missed it.

Ricocotam:
I was actually going to post a similar iterator that I called `flatten_iflist_iterate`

```python
def flatten_iflist_iterate(input_list):
    for item in input_list:
        if type(item) == list:
            for i in flatten_iflist_iterate(item):
                yield i
        else:
            yield item
```
I think your code is actually better but we need a recursion to handle nested lists : 

```python
>>> a = [1, 2, [3, 4, [5, [6], 7]], 8, 9]
>>> list(flatten_iflist_iterate(a))
[1, 2, 3, 4, 5, 6, 7, 8, 9]
>>> list(didium(a))
[1, 2, 3, 4, [5, [6], 7], 8, 9]
```

jenstroeger:
@Ricocotam, I actually didn’t want to recurse into nested iterators and only skim the first level.
```
>>> a = [1, 2, [3, 4, [5, [6], 7]], 8, 9]
>>> list(didum(a))
[...]

PR #475, opened 2020-11-30
Add new `value_chain` iterator.

jenstroeger:
See also issue https://github.com/more-itertools/more-itertools/issues/468. This is the non-recursive implementation of the discussed iterator. Happy to add the `collapse` parameter, though.

Also, considering that `value_chain()` is a more general case of `prepend()`, would you like to tail-call:
```python
def prepend(value, iterator):
    return value_chain(value, iterator)
```


more_itertools/more.py
```diff
@@ -33,6 +33,7 @@
     powerset,
     take,
     unique_everseen,
+    prepend,
```
jenstroeger, line 36: Hmm. @bbayles, if I remove the `prepend` then Doctest fails, if I add `prepend` then flake8 fails. What now?

more_itertools/more.py
```diff
@@ -3552,3 +3554,39 @@ def nth_permutation(iterable, r, index):
...
+    of :func:`prepend`:
+
+        >>> list(prepend(0, [1, 2, 3]))
+        [0, 1, 2, 3]
+        >>> list(value_chain(0, [1, 2, 3]))
+        [0, 1, 2, 3]
+
+    """
+    for value in args:
+        if isinstance(value, (str, bytes,)):
```
jenstroeger, line 3583: We should probably add the other scalar types here, too: `int`, `float`, `bool`.

more_itertools/more.py
bbayles: It shouldn't be imported here; it should be imported in the docstring before [this line](https://github.com/more-itertools/more-itertools/pull/475/files#diff-d7dff201bd69500a573b7e96c52f8e625242d769a04621224112faf6e7bc63fdR3576).

jenstroeger: Duh. Commit https://github.com/more-itertools/more-itertools/pull/475/commits/254133ff846af44c99150b645bd73e4999bd06a3.

more_itertools/more.py
jenstroeger: Commit https://github.com/more-itertools/more-itertools/pull/475/commits/0d67cde0543810d8ca75005af2d9c1e94a0c7dd9.

jenstroeger:
Ok so that one failure is a notification from [black](https://github.com/psf/black) which would reformat this:
```python
if isinstance(value, (str, bytes, int, float, bool,)):
```
to this:
```python
if isinstance(
    value,
    (
        str,
        bytes,
        int,
        float,
        bool,
    ),
):
```
This is just plain ugly, IMO, but it’s not my code either so I’m happy to comply.

bbayles:
Go ahead and make the fomatter happy, if you could please.

jenstroeger:
Done 🤓

bbayles:
Merged with some changes:
* Added to API docs
* The `int`, `float`, `bool` code that was added in 0d67cde0543810d8ca75005af2d9c1e94a0c7dd9 weren't tested and were hard to explain - I've left those out.
* Removed references to the `prepend` function - it's a simple recipe from the `itertools` docs that we bring in here for completeness.

Many thanks!

PR #477, opened 2020-12-03
Fixed 404ing link of 'Tour of Python Itertools'

Evantm:


bbayles:
Thanks - but the original link still works for me! You see differently?

Evantm:
image
The blog uses an API which uses a non standard port (https://martinheinz.dev:1234/api/v1/posts/16) which is blocked by many firewalls.

bbayles:
Hmm, odd choice. I'd like to keep the original link, but yours as an alternate is helpful.

Issue #451, opened 2020-07-23
distinct_combinations() crashing with large r value


Minabsapi:
For a personal project I wanted to use this handy function to avoid unnecessary duplicates, turns out it won't work when using it against a large iterable and `r` value (respectively of size 2000 and value 1000).

I'd suggest either to rewrite the function to avoid recursions as much as possible or warn (in the documentation or at runtime) that `r` should not exceed the maximum recursion call system limit.

bbayles:
Any details you could share about: (a) the iterable, (b) the calculation or operation you're doing with it, (c) what error you got?

Minabsapi:
Thanks for your reply
1. The iterable is a 1000-elements long list of integers
2. I am iterating through this generator function, doing some comparisons and testing some conditions. So nothing intensive or eventually impacting the function
3. The error is, as suggested in my initial message, a recursion error:
```
File "/home/minabsapi/SomeProject/main.py", line 69, in <module>
    for combination in more_itertools.distinct_combinations(sorted(working_group), len(local_fraction[0])):
  File "/home/minabsapi/.local/lib/python3.8/site-packages/more_itertools/more.py", line 3080, in distinct_combinations
    for suffix in distinct_combinations(pool[i + 1 :], r - 1):
  File "/home/minabsapi/.local/lib/python3.8/site-packages/more_itertools/more.py", line 3080, in distinct_combinations
    for suffix in distinct_combinations(pool[i + 1 :], r - 1):
  File "/home/minabsapi/.local/lib/python3.8/site-packages/more_itertools/more.py", line 3080, in distinct_combinations
    for suffix in distinct_combinations(pool[i + 1 :], r - 1):
  [Previous line repeated 993 more times]
  File "/home/minabsapi/.local/lib/python3.8/site-packages/more_itertools/more.py", line 3079, in distinct_combinations
    for i, prefix in unique_everseen(enumerate(pool), key=itemgetter(1)):
  File "/home/minabsapi/.local/lib/python3.8/site-packages/more_itertools/recipes.py", line 389, in unique_everseen
    iterable, keys = tee(iterable)
RecursionError: maximum recursion depth exceeded while calling a Python object
```

bbayles:
What's the density of repetitions in the iterable, and what's your `r`? I'm wondering if this weren't recursive if you'd just run out of memory.

```
from more_itertools import distinct_combinations, ilen

n = 500
ilen(distinct_combinations('A' * n + 'B' * n, n))
```

[...]

PR #483, opened 2020-12-10
Iterative solution for distinct_combinations

knutdrand:
Should scale slightly better than recursive solution, issue #451


bbayles:
Excellent! Many thanks.

bbayles (approved):

PR #484, opened 2020-12-13
add key argument to sort_together() function

### LINKED ISSUE #480: allow passing your own key function to sort_together
I recently came across a situation in which I wanted to use `sort_together`, but I also wanted to sort by a custom key.
What do you think about exposing a `key` argument to `sort_together`?

I think it would also be nice for the user's key function to take only the parameters corresponding to the `key_list`, so the common case of sorting by the first is the easiest, and so the two features of `key` and `key_list` work together.

Something to this effect:

```python
def sort_together(iterables, key_list=(0,), key=lambda *x: x, reverse=False):
    return list(
        zip(
            *sorted(
                zip(*iterables), key=lambda x: key(*always_iterable(itemgetter(*key_list)(x))), reverse=reverse
            )
        )
    )


list_of_x = [1, 2, 3]
list_of_y = [4, 5, 6]

sort_together((list_of_x , list_of_y), key=lambda x: x*2)
sort_together((list_of_x , list_of_y), key_list=(1,), key=lambda y: y*2)
sort_together((list_of_x , list_of_y), key_list=(0, 1), key=lambda x, y: x*y)
```

Happy to open a PR if you like the idea.

brianmaissy:
Closes #480

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -1595,9 +1595,13 @@ def sort_together(iterables, key_list=(0,), key=None, reverse=False):
     function. Its arguments are the elements of the iterables corresponding to
     the key list::
 
-        >>> iterables = [(3, 1, 2), (1, 0, 2), ('c', 'b', 'a')]
-        >>> sort_together(iterables, key_list=(0, 1), key=lambda x, y: x - y)
-        [(2, 1, 3), (2, 0, 1), ('a', 'b', 'c')]
+        >>> names = ('a', 'b', 'c')
+        >>> lengths = (1, 2, 3)
+        >>> widths = (5, 2, 1)
+        >>> def area(length, width):
+        ...     return length * width
+        >>> sort_together([names, lengths, widths], key_list=(1, 2), key=area)
+        [('c', 'b', 'a'), (3, 2, 1), (1, 2, 5)]
 
     Set *reverse* to ``True`` to sort in descending order.
 
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -1569,7 +1569,7 @@ def zip_offset(*iterables, offsets, longest=False, fillvalue=None):
     return zip(*staggered)
 
 
-def sort_together(iterables, key_list=(0,), reverse=False):
+def sort_together(iterables, key_list=(0,), key=None, reverse=False):
     """Return the input iterables sorted together, with *key_list* as the
     priority for sorting. All iterables are trimmed to the length of the
     shortest one.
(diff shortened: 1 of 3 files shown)


more_itertools/more.py
```diff
@@ -1591,18 +1591,43 @@ def sort_together(iterables, key_list=(0,), reverse=False):
...
-                zip(*iterables), key=itemgetter(*key_list), reverse=reverse
+    if key is None:
+        # if there is no key function, the key argument to sorted is an
+        # itemgetter
+        key_argument = itemgetter(*key_list)
+    else:
+        # if there is a key function, call it with the items at the offsets
+        # specified by the key function as arguments
+        key_list = list(key_list)
+        if len(key_list) == 1:
```
bbayles, line 1616: Thanks for breaking this out. I'm still a little puzzled - does the length-1 case actually fail and need special handling? If so, why?

brianmaissy: If we were to use the code from the other case, the `itemgetter()` would return a single element (rather than a tuple), and when we try to pass them to the key function with `*` (iterable unpacking) it would raise a `TypeError` that "argument after * must be an iterable".

The essence of the problem is that the degenerate case of `itemgetter()` with a single argument returns a single value, whereas when given multiple arguments it returns a tuple.

bbayles: OK, got it - thanks.

```python
>>> from operator import itemgetter
... 
... get_1 = itemgetter(1)
... get_1([1, 2, 3])
... 
... get_1_2 = itemgetter(1, 2)
... get_1_2([1, 2, 3])
1: (2, 3)
```

more_itertools/more.py
```diff
@@ -1591,18 +1591,43 @@ def sort_together(iterables, key_list=(0,), reverse=False):
         >>> sort_together(iterables, key_list=(1, 2))
         [(2, 3, 1), (0, 0, 1), ('a', 'c', 'b')]
 
+    To sort by a function of the elements of the iterable, pass a *key*
+    function. Its arguments are the elements of the iterables corresponding to
+    the key list::
+
+        >>> iterables = [(3, 1, 2), (1, 0, 2), ('c', 'b', 'a')]
+        >>> sort_together(iterables, key_list=(0, 1), key=lambda x, y: x - y)
```
bbayles, line 1599: This example is a little confusing. What's being subtracted in the last item?

brianmaissy: The lists are being sorted by the difference between the value in the first list and the second list at each index (since the key list is 0 and 1, x and y are elements of the first and second lists respectively).

It would raise a TypeError on that input, because it would be subtracting a number from a string.

bbayles: Could you put a different example in the docstring? This one is hard (for me) to reason about - it's sort of a fluke that not all the items get touched by the key function?

brianmaissy: Yeah sure, I'll find a better example.

Can you explain what you mean by "not all the items get touched by the key function"? You mean that the key function gets only two arguments? That's because of the key list.

Maybe I'll add a second example to demonstrate the interaction of the key list and the key function? Or maybe it would be clearer if they didn't interact, and the key function always got all of the elements as it's arguments.

brianmaissy: I made the example clearer, let me know what you think

bbayles (approved):

Issue #487, opened 2020-12-17
Provide a way for time_limited() to indicate why it ended


roysmith:
I have a perfect use for time_limited(), but I'd also like to be able to tell the end user, "Data stream truncated after N seconds, results are incomplete"  As far as I can tell, there's no way determine if the iteration ended because of a timeout or because the underlying iterator was exhausted.

Maybe add an attribute to the returned iterator?  I could envision something like:

```
iterator_2 = time_limited(n, iterator_1)
my_data = list(iterator_2)
if iterator_2.timed_out:
    print("Data stream truncated after N seconds, results are incomplete")
```
but I'm not actually sure that's implementable.

bbayles:
I like it! I think it is implementable, if `time_limited` were turned into a class. I'll play with it. Great suggestion!

PR #490, opened 2020-12-18
Add timed_out property for time_limited

bbayles:
Re: https://github.com/more-itertools/more-itertools/issues/487, this PR adds a means of determining whether `timed_out`, well, timed out.

The test failures are due to https://github.com/actions/setup-python/issues/171; I'll merge after that is fixed.


more_itertools/more.py
```diff
@@ -3035,34 +3035,49 @@ def set_partitions_helper(L, k):
         yield from set_partitions_helper(L, k)
 
 
-def time_limited(limit_seconds, iterable):
+class time_limited:
     """
     Yield items from *iterable* until *limit_seconds* have passed.
+    If the iterable is not exhausted in the time limit, the ``timed_out``
```
roysmith, line 3041: I think:

`If the time limit is reached before the iterable is exhausted, the ``timed_out`` attribute will be ``True``
`

would be clearer.

PR #492, opened 2020-12-26
Add product_index, combination_index, and permutation_index functions

N8Brooks:
- These are the inverses of nth_product, nth_combination, and nth_permutation respectively. 
 - If there are multiple such indexes of element, then the first index is returned similar to the index methods for lists or tuples. 
 - The implementation of combination_index is kind of weird ... I thought it would be better to have an iterable to index stuff and find the length of the iterable because it is only needed once. On the other hand, with product_index, finding the index of x and the length is needed every loop and so I thought it was better to cast it to a tuple. 
 - With the nth_combination and nth_permutation functions, r is required in order to find the element it returns. Here, r is taken implicitly using the length of the element given. 
 - Some python errors are relied upon. For example if the pool.index(x) fails in product_index. Maybe these places should be wrapped to raise a more descriptive error?
 - Ideally the list in permutation_index would be a b-list data structure or something similar to remove items in log(n) time.


bbayles:
Nice implementations! I'm divided on whether to include these in the library, however. They're certainly related to the `combinations`, `permutations`, and `product` functions. But they're not really functions that take iterables as input or give them as output, nor ones that augment other iterables, etc.

If you've got any references to Stack Overflow or the Python mailing lists that have people using `itertools` needing these functions, that would help me decide. Thanks!

N8Brooks:
They take iterables as input, but the output is strictly an integer. This is similar to some other things in the library such as ilen. 

As for stack overflow, I came across a few posts: 
* https://stackoverflow.com/questions/32884220/python-itertools-product-get-item-index
* https://stackoverflow.com/questions/14455634/find-the-index-of-a-given-combination-of-natural-numbers-among-those-returned
* https://math.stackexchange.com/questions/1363239/fast-way-to-get-a-position-of-combination-without-repetitions
* https://codereview.stackexchange.com/questions/136612/permutation-index-python

I think it could be a strong addition, because sometimes you may need an nth_permutation along with a permutation_index. Suppose there could be genetic traveling salesperson solution where two permutations of cities are combined using the average of their indexes?

bbayles:
OK, thanks. Could you make `element` the first argument in each of the functions - right now `product` is different?

Things to consider for later: a version for `combinations_with_replacement`, support for different `r` values.

N8Brooks:
Yup! Added another commit. Yeah ... combinations_with_replacement should have an appropriate function for nth and index. Unfortunately, I have no idea how to do that. I'll add it eventually unless someone beats me to it :D

more_itertools/more.py
```diff
@@ -3683,3 +3686,77 @@ def value_chain(*args):
             yield from value
         except TypeError:
             yield value
+
+
+def product_index(element, *args):
+    """Equivalent to ``list(product(*args)).index(element)``
+
+    The products of *args* can be ordered lexicographically.
```
pylang, line 3694: Most docstrings in this library have a clear example.  I've notice these (and the *_nth combinatorics actually) are lacking them.  Example:

```python
>>> list(product(1, "abc", [1, 2, 3]))
?
```

Please include them to maintain consistency and clarity.

pylang (changes requested):

N8Brooks:
The comments were updated with examples as well as a description of errors raised. Checks failed for different reasons whether or not the itertools functions in the comments were imported. Not sure how to proceed.

more_itertools/more.py
```diff
@@ -3597,6 +3600,17 @@ def nth_product(index, *args):
     The products of *args* can be ordered lexicographically.
     :func:`nth_product` computes the product at sort position *index* without
     computing the previous products.
+
+        >>> nth_product(8, range(2), range(2), range(2), range(2))
+        (1, 0, 0, 0)
+
+    The equivalent being:
+
+        >>> list(product(range(2), range(2), range(2), range(2)))[8]
```
bbayles, line 3609:
```suggestion
        >>> from itertools import product
        >>> list(product(range(2), range(2), range(2), range(2)))[8]
```

bbayles:
My suggestion should fix it; apply to the other examples as well?

FWIW I usually rewrite the docstrings before merging in, but it would be good if we can make sure CI passes before I do tht.

Issue #502, opened 2021-02-08
Idea: An iterator wrapper that counts items yielded so far


krzysieq:
Sometimes I find that I have an iterable that I want to consume, but as a side effect I want to count the number of items yielded:
```python
counting_iterator = CountingIterator(my_iterable)
consume(counting_iterator)
counting_iterator.counter  # -> number of `next` calls on the iterator
```

roysmith:
Isn't that just len(my_iterable)?

krzysieq:
Perhaps it's most useful to think about cases where iteration over `my_iterable` is expensive (so the `len` call is wasteful) or `my_iterable` is an iterator that can't be re-iterated over once exhausted (so code that wants to separately call `len` and `consume` would need to cache elements of the iterator in some auxiliary store).

bbayles:
`ilen` does this, yes?

bbayles:
Oh, I was missing the "so far" aspect. I get it now.

krzysieq:
You're right, `len` doesn't work with an arbitrary iterable so the argument above really should have been about `ilen`.

pylang:
I can imagine this working with `make_decorator` and using it to count yielded items, similar to a classic `@count_calls` decorator, which tallies how many times a function is called.

bbayles:
I like this idea - check out https://github.com/more-itertools/more-itertools/pull/503 and let me know what you think?

krzysieq:
Thanks, that looks great.

PR #503, opened 2021-02-09
Add countable()

bbayles:
Re: https://github.com/more-itertools/more-itertools/issues/502, this PR adds `countable`, a wrapper for iterables that provides an `items_seen` counter.


more_itertools/more.py
```diff
@@ -3789,3 +3790,35 @@ def permutation_index(element, iterable):
...
+
+class countable:
+    """Wrap *iterable* and keep a count of how many items have been consumed.
+
+    The ``items_seen`` attribute starts at ``0`` and increments as the iterable
+    is consumed:
+
+        >>> iterable = map(str, range(10))
+        >>> it = countable(iterable)
+        >>> it.items_seen
```
pylang, line 3803: This name `items_seen` ... Seems alright now, but I'm wondering if someday I'd prefer a Counter.  If so, `items_seen` could mean a dict of observations, rather than an int.  Maybe `count_seen` or  `count_justseen`?

krzysieq: A full Counter could certainly be useful in some cases, but can I ask that this wrapper never uses a Counter _by default_? Besides the question of hashability of the items, the wrapper loses quite a bit of its appeal to me if it were to use O(n) memory.

PR #504, opened 2021-02-27
Dirty partition

zmej-serow:
Added optional parameter `side_effect` to `partition` function. I find it quite useful and readable.


zmej-serow:
Not sure why `black` fails on 3.6.
I've run black locally and it tells me "All done! ✨ 🍰 ✨"...

bbayles:
Thanks for the PR! I'm inclined not to include this in the library, because I think it opens Pandora's box: adding a side effect-like feature to each of the functions.

> Not sure why black fails on 3.6.
> I've ran black --check . locally and it tells me "All done! sparkles cake sparkles"...

Usually this is due to having a different version of `black` from what the CI system is using.

zmej-serow:
> I'm inclined not to include this in the library, because I think it opens Pandora's box: adding a side effect-like feature to each of the functions.

I thought about it, too. Probably you're right, there is only one angel by my side: as `partition` creates *two* generators, there is a necessity to wrap them *both* with the side-effect function. On the other hand, if we incorporate side-effect in `partition`, the whole thing will be much more elegant.

Your thoughts?

> > Not sure why black fails on 3.6.
> Usually this is due to having a different version of `black` from what the CI system is using.

Oh, probably yes. I'm using 19.10b0, and CI?

bbayles:
> Probably you're right, there is only one angel by my side: as partition creates two generators, there is a necessity to wrap them both with the side-effect function

Thanks - I did consider that. But I don't think saving one line `x, y = partition(iterable); x, y = side_effect(x, ...), side_effect(y, ...)` is worth the baggage here.

Appreciate the PR - do feel free to propose other changes!

zmej-serow:
> I don't think saving one line `x, y = partition(iterable); x, y = side_effect(x, ...), side_effect(y, ...)` is worth the baggage here.
> 
> Appreciate the PR - do feel free to propose other changes!

Thanks! Probably I will. It's really a gem, the more_itertools package.
Appreciate the discussion!

Issue #508, opened 2021-03-10
Inconsistent behaviour of the split_before function for an empty collection


TiunovNN:
I've noticed that the `split_before` function never returns an empty iterator:
```python
>>> list(split_before([], lambda x: bool(x)) 
[[]]
```
In the meantime, the `split_after` function works correctly:
```python
>>> list(split_after([], lambda x: bool(x)))
[]
```
It looks like inconsistent behavior

PR #509, opened 2021-03-10
Fix split_before for an empty collections.

TiunovNN:
### Issue reference

more-itertools/more-itertools#508

### Changes

If the function `split_before` gets an empty collection it will return an empty iterator.
```python
>>> list(split_before([], pred=lambda x: bool(x)) ) 
[]
```


bbayles (approved):

bbayles:
Thanks!

Issue #438, opened 2020-06-17
Deprecate `zip_equal()`?


pylang:
[PEP 618](https://www.python.org/dev/peps/pep-0618/) has been [accepted](https://mail.python.org/archives/list/python-dev@python.org/message/NLWB7FVJGMBBMCF4P3ZKUIE53JPDOWJ3/), adding `strict=False` to `zip()`.  `zip(strict=True)` will throw an exception if lengths are unequal.  

In light of this PEP, should `zip_equal()` be deprecated?  If so, it may be helpful to warn users early of a future deprecation.

bbayles:
Surprised this made it in. I'll mark it with `DeprecationWarning` when Python 3.10 is released.

PR #511, opened 2021-03-14
Add a DeprecationWarning for 3.10+

bbayles:
Re: https://github.com/more-itertools/more-itertools/issues/438, this PR adds a `DeprecationWarning` for `zip_equal` when running on Python 3.10 and higher.


bbayles:
Note to self: the test failure is related to building `mypy` on the latest 3.10 alpha. This will probably be fixed when wheels get published [here](https://pypi.org/project/mypy/#files).

PR #513, opened 2021-04-27
Add Excel-style column count iterator.

jenstroeger:
Implements issue https://github.com/more-itertools/more-itertools/issues/514. You’ve probably seen Excel and other spreadsheet applications’ column counts: `A`, `B`, `C`, …, `Z`, `AA`, `AB`, … and so forth:

excel

This PR implements a count iterator much similar to Python’s [`itertools.count()`](https://docs.python.org/3/library/itertools.html#itertools.count) but which returns said column counts:
```python
>>> from more_itertools import xcount
>>> count = xcount(start="Y")
>>> next(count)
'Y'
>>> next(count)
'Z'
>>> next(count)
'AA'
>>> next(count)
'AB'
```
I left the PR as a draft because I still have two questions:

 1. There aren’t any counters in the `more_itertools` package which is why I added the new iterator under “Others”. Perhaps a “Counters” row in README.rst would be better?
 2. Not sure I like `xcount()` as a function name, but `excel_count()` wasn’t any better either. I’m very open to suggestions 😉


jenstroeger:
@bbayles, it looks like the tests are failing because _other_ code currently in the repo would be reformatted by black:
```
Using configuration from /.../more-itertools/pyproject.toml.
--- more.pyi	2021-04-27 05:46:32 +0000
+++ more.pyi	2021-04-27 05:53:30.653016 +0000
@@ -28,10 +28,11 @@
 _V = TypeVar('_V')
 _W = TypeVar('_W')
 _T_co = TypeVar('_T_co', covariant=True)
 _GenFn = TypeVar('_GenFn', bound=Callable[..., Iterator[object]])
 _Raisable = Union[BaseException, 'Type[BaseException]']
+
 @type_check_only
 class _SizedIterable(Protocol[_T_co], Sized, Iterable[_T_co]): ...
 
 @type_check_only
 class _SizedReversible(Protocol[_T_co], Sized, Reversible[_T_co]): ...
would reformat more.pyi
All done! ✨ 🍰 ✨
1 file would be reformatted.
```
and
```
Using configuration from /.../more-itertools/pyproject.toml.
--- test_more.py	2021-04-27 05:00:18 +0000
+++ test_more.py	2021-04-27 05:53:56.174343 +0000
@@ -635,11 +635,11 @@
             lambda: mi.one(it),
         )
 
 
 class IntersperseTest(TestCase):
-    """ Tests for intersperse() """
+    """Tests for intersperse()"""
 
     def test_even(self):
         iterable = (x for x in '01')
         self.assertEqual(
             list(mi.intersperse(None, iterable)), ['0', None, '1']
would reformat test_more.py
All done! ✨ 🍰 ✨
1 file would be reformatted.
```

jenstroeger:
Thank you @bbayles for the update to master, I rebased this PR.

> I'd like to understand the use case for this. In the issue template we have:
> ```
> 
> ```

I work a lot with XML documents, and increasingly they contain tables; often times the rows and columns and cells in those tables are without an `id` attribute and so my script adds that. And because Microsoft Office Word and Libre/OpenOffice Writer and Excel-type applications all enumerate their columns from `A`, `B`, …, `Z`, `AA`, … I needed the counter.

The `more_itertools` seemed like a good place for it.

> […] this is a good use for `itertools.product`. Here is an example of generating the same sequence:

Agreed 👍🏼 

So, what do you want to do? Happy to use your code if you’d like to add this counter to the `more_itertools` package in which case my two questions above are still open.

bbayles:
Thanks for the info. I'm not inclined to add this to `more-itertools`, but it does seem like it would be a good addition to a library focused on MS Excel. I'm not sure what the best library would be for that - don't most people use `pandas` these days?

jenstroeger:
> […] it does seem like it would be a good addition to a library focused on MS Excel.

It’s not an Excel feature: various applications that deal with tables (e.g. Word or Writer) seem to have adopted the convention of counting columns that way; I don’t know if Excel was the first to do that.

For my purpose, I wouldn’t want to take a `pandas` dependency only to get access to a column counter. I’ll just add to on of our local util packages then.

Feel free to close this PR.

PR #522, opened 2021-05-26
Add unanimous()

kalekundert:
This PR adds a `unanimous()` function, which checks to see if all of the values in an iterable are the same and returns that value if so.  This is complementary to `all_equal()`, which makes the same check but does not return the corresponding value.  

Here's a brief example:
```
>>> from more_itertools import unanimous
>>> unanimous([1, 1, 1])
1
>>> unanimous([1, 1, 2])
  ...
ValueError: found multiple values: 1, 2
```
Let me know if this seems like a reasonable contribution, and if any changes are needed.


bbayles:
Thanks for the PR. I'm inclined to think this is a little too similar to `all_equal` to be justified, but if you have any of this I might change my mind!

```
<!-- Link to a discussion on, e.g. Stack Overflow, showing where the new tool would have been useful -->
<!-- Or link to a GitHub issue in another repository -->
<!-- Or a discussion on the python-ideas mailing list -->
<!-- Alternatively, describe a real problem you had to solve with this itertool -->
```

kalekundert:
I use `unanimous()` a lot in the context of writing scripts that output laboratory protocols.  Unfortunately these kinds of scripts don't lend themselves to simple examples, but hopefully I can still explain how they benefit from `unanimous()`.  The following snippet is from a protocol that involves loading heterogeneous samples of DNA/RNA molecules onto a gel matrix to be separated by size (this is called [gel electrophoresis](https://en.wikipedia.org/wiki/Gel_electrophoresis)).  If all the samples have the same volume, I assign that volume to an attribute of the `gel` object that knows how to display it.  If any of the samples have different volumes, I instruct the `gel` object to not display any volumes and I add an extra step to the protocol describing how much of each sample to load:

```python
try:
    gel.load_volume_uL = unanimous(
            x.load_volume_per_lane_uL
            for x in self.samples
    )
except ValueError:
    gel.load_volume_uL = 0

    protocol += pl(f"Load the {n:sample/s} in the gel as follows:", u := ul())
    for sample in self.samples:
        num_lanes = f" in each of {m} lanes" if (m := sample.num_lanes) != 1 else ""
        u += f"{sample.load_volume_per_lane_uL:.2g} µL {sample.name}{num_lanes}"
```

If I were to write this snippet with `all_equal()`, I'd have to add some extra logic to go back and get the volume in question after checking to see if all the volumes are the same.   This logic wouldn't be so bad in this case, because the value I want is just `self.samples[0].load_volume_per_lane_uL`, but it's easy to imagine cases that wouldn't be so simple:

- If I only had an iterator/generator rather than a list, I'd have to use something like `tee()` or `spy()` to get the value in question before `all_equal()` consumes the iterable.
- If the comprehesion in the above snippet did some expensive calculation that I didn't want to repeat, I'd have to assign it to a temporary variable instead of using it anonymously.
- If I wanted to use some default value in the case that the iterable is empty, I'd have to add more extra logic to deal with that.  `all_equal()`—like `all()`—simply returns `True` when given an empty iterable.  This extra logic would be especially complicated and inefficient if the iterable was an interator/generator (e.g. you'd basically have to convert it into a list first, which defeats the purpose of using iterators/generators).

Overall, I'd argue that if you use `all_equal()` to figure out the value shared by all elements of an iterable, you'll run into a lot of corner cases and the right approach will depend a lot on the specific iterable you're working with.  More conceptually, it just feels wrong to have to do to any amount of extra effort to get a value that `all_equal()` clearly knows.  Of course, `all_equal()` can't return that value because it has to return a boolean, but that's why I think `unanimous()` complements `all_equal()` well.  By using exceptions to indicate whether or not every element has the same value, `unanimous()` communicates two pieces of information to the caller rather than just one.

bbayles:
> If I only had an iterator/generator rather than a list, I'd have to use something like tee() or spy() to get the value in question before all_equal() consumes the iterable.

What's wrong with this? That is why those functions exist.

Some more questions:  What would  `unanimous()` return for an empty iterable? `all_equal` returns `True`.

Also, what about a `return_value` keyword argument for `all_equal`?

kalekundert:
> What's wrong with this? That is why those functions exist.

That's true.  The point I was trying to make is that if you weren't specifically thinking about iterators/generators, you'd probably write the code in a way that wouldn't support them.

> Some more questions: What would unanimous() return for an empty iterable?

It raises an exception unless a default is provided:
```python
>>> unanimous([])
  ...
ValueError: empty iterable
>>> unanimous([], default=False)
False
```

> Also, what about a return_value keyword argument for all_equal?

I don't think that's a good idea.  I'm not sure exactly what you have in mind, but I assume that `return_value=True` would change the return type of `all_equal` to something like (`is_all_equal, value)`.  The problems with this are:

- The return type of a function shouldn't depend its arguments, as discussed [here](https://mail.python.org/pipermail/python-ideas/2016-May/040193.html).  This is basically shoehorning two functions into one, and it makes the API less clear.
- There's no reasonable value for the `value` argument if the iterable isn't all equal.  `unanimous()` gets around this by raising an exception in cases where it doesn't make sense to return something.

bbayles:
After some thought, I'm going to pass on this - the recommended alternative is `spy` and `all_equal`.

Thanks for the PR and writeups!

PR #526, opened 2021-06-02
Add type overloads for zip_equal and zip_offset

### LINKED ISSUE #525: zip_equal and zip_offset lose type information
The standard-library type definitions for `zip` ([2.x](https://github.com/python/typeshed/blob/master/stdlib/%40python2/builtins.pyi#L1030), [3.x](https://github.com/python/typeshed/blob/master/stdlib/builtins.pyi#L1310)) have special cases for 1-5 iterable arguments, and a fallback for 6 or more arguments or if called with varargs. This gives better typing information on common cases where one wants to zip a fixed number of iterables.

The type definitions for `zip_equal` and `zip_offset` currently only have a case for varargs which loses type information about the result: we don't know either the length of output tuples or types of individual elements.

maffoo:
### Issue reference
Fixes #525 

### Changes
This adds overloaded type definitions for `zip_equal` and `zip_offset`, following the standard-library type definitions for `zip` ([2.x](https://github.com/python/typeshed/blob/master/stdlib/%40python2/builtins.pyi#L1030), [3.x](https://github.com/python/typeshed/blob/master/stdlib/builtins.pyi#L1310)). This creates special cases for 1-5 iterable arguments, and a fallback for 6 or more arguments or if called with varargs. This gives better type information in common situations when zipping a fixed number of iterables.

### DIFF
--- a/more_itertools/more.pyi
+++ b/more_itertools/more.pyi
@@ -24,6 +24,12 @@ from typing_extensions import ContextManager, Protocol, Type, overload
 
 # Type and type variable definitions
 _T = TypeVar('_T')
+_T1 = TypeVar('_T1')
+_T2 = TypeVar('_T2')
+_T3 = TypeVar('_T3')
+_T4 = TypeVar('_T4')
+_T5 = TypeVar('_T5')
+_T6 = TypeVar('_T6')
 _U = TypeVar('_U')
 _V = TypeVar('_V')
 _W = TypeVar('_W')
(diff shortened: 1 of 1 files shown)


bbayles:
Thanks for the PR.

The argument for following what the standard library does is strong. However, I think I object on aesthetic grounds. Repeating the same bit of line noise six times? This is Python?

I could see the argument for "one, two, many" perhaps - if you want to resubmit that, I'll merge it?

maffoo:
I agree it's ugly, but it's also "just" type definitions so doesn't pollute the code itself. But I'll defer to you if you'd like to keep just the first two type specializations. The particular case where I ran into this in my code was zipping two iterables, so keeping just the two-arg case would be sufficient for what I need right now. And I suspect zipping two iterables is the most common case so that might cover other people's needs as well.

maffoo:
Note: mypy added type defs for zip up to 5 args here: https://github.com/python/typeshed/pull/1041. GvR commented there that this could eventually be superseded by variadic generics: https://github.com/python/typing/issues/193.

bbayles:
> But I'll defer to you if you'd like to keep just the first two type specializations.

Yes please!

bbayles (approved):

bbayles (changes requested):
Check out the CI failures on the type changes?

bbayles (approved):

jjjamie:
FWIW, I'd also prefer to keep the typedefs for >2 iterables. This fixes Pycharm linter errors for zipping two iterables of different types but it'll still complain about 3+.

I would prefer some extra (well motivated, correct ;)) noise in a `pyi` file to false errors highlighted in yellow every time I read my code. (Variadic generics haven't appeared in the 4 years since the GvR comment on https://github.com/python/typeshed/pull/1041 - not sure it's worth waiting for them!)

Issue #527, opened 2021-06-18
Add transpose function


qalshidi:
### Description
Transposes an iterable of iterables. I found myself needing a function like this a lot.

### References
https://stackoverflow.com/questions/6473679/transpose-list-of-lists

### Examples
```
>>> list(transpose([[1, 2, 3], [4, 5, 6]]))
[(1, 4), (2, 5), (3, 6)]
```

pylang:
Is `transpose()` something other than as `more_itertools.interleave`?

https://more-itertools.readthedocs.io/en/latest/api.html#more_itertools.interleave

qalshidi:
> Is `transpose()` something other than [`more_itertools.interleave_longest`](https://more-itertools.readthedocs.io/en/latest/api.html#more_itertools.interleave_longest)?

Definitely. Transpose returns an exchange of rows and columns. Interleave returns a single row. They definitely have different uses and the implementations are different.

PR #528, opened 2021-06-18
Add transpose() function

qalshidi:
### Issue reference
more-itertools/more-itertools #527

### Changes
Adds a new itertools function to transpose iterable of iterables. Useful for quite a few cases that is not necessarily numerical which wouldn't call for `numpy`.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -106,6 +106,7 @@ __all__ = [
     'substrings',
     'substrings_indexes',
     'time_limited',
+    'transpose',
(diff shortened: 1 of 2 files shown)


bbayles:
Thanks for the writeup and PR.

I'm not inclined to add this function, as I think `zip_longest` is pretty much suitable for this purpose - is the "inner type" change really so general a need?

Thanks again!

qalshidi:
Okay thanks for considering

Issue #530, opened 2021-07-06
interleave_evenly


mbugert:
### Description

Interleave multiple iterables with known length so that their elements are evenly distributed throughout the output sequence.

### References
Use case: Recently, I had multiple datasets A, B, C, ... of vastly different size on which I wanted to train a neural network.
Without going too much into detail here, training works best if the elements from these datasets are distributed evenly in the data stream fed to the network.

So why `interleave_evenly` when there are `interleave` and `interleave_longest`? Consider the following example:
```python3
>>> a, b, c = ['a']*20, ['b']*5, ['c']*3
>>> ''.join(interleave(a, b, c))
'abcabcabc'                      # data is trimmed

>>> ''.join(interleave_longest(a, b, c))
'abcabcabcababaaaaaaaaaaaaaaa'   # uneven distribution

>>> ''.join(interleave_evenly([a, b, c], [20, 5, 3]))
'aabacaaabaaacabaaaabaacaabaa'   # all the data, evenly distributed
```

The solution essentially is a multi-dimensional extension of [Bresenham's line drawing algorithm](https://en.wikipedia.org/wiki/Bresenham%27s_line_algorithm). The iterable lengths are needed to achieve the desired distribution. Passing the lengths separately looks a bit unnecessary in the examples given that the iterables are lists, but in my use case, the datasets were too big to fit in memory, so there was a good reason to go with iterators + precomputed lengths.

### Examples
```python
>>> iterables = [1, 2, 3, 4, 5], ['a', 'b']
>>> lengths = [5, 2]
>>> list(interleave_evenly(iterables, lengths))
[1, 2, 'a', 3, 4, 'b', 5]

>>> iterables = [[1, 2, 3], [4, 5], [6, 7, 8]]
>>> lengths = [3, 2, 3]
>>> list(interleave_evenly(iterables, lengths))
[1, 6, 4, 2, 7, 3, 8, 5]
```

PR #531, opened 2021-07-06
Add interleave_evenly function and tests

mbugert:
### Issue reference

more-itertools/more-itertools#530

### Changes
Add `interleave_evenly` implementation plus unit tests.

For some unknown reason, the newly added tests all failed with `AttributeError: module 'more_itertools' has no attribute 'interleave_evenly'` in my pycharm, but when I moved the `InterleaveEvenlyTests` class into a separate file, it worked.


FinalSh4re:
You didn't add your function to \_\_all\_\_, this should be the reason why the method isn't found.

https://github.com/more-itertools/more-itertools/blob/7b89d6ead4725fa1160e466ab1ebb55747de55aa/more_itertools/more.py#L38

mbugert:
Right, thanks for pointing out. I've added it and also updated `api.rst`.

bbayles:
One question before I merge this: is there any reason not to compute the lengths for the user, rather than making them pass them in? None of the tests use custom lengths.

mbugert:
Good question.
The idea was to make it compatible with any iterable: in my original use case, I interleaved iterators which were `itertools.combinations`, i.e. iterators which produce a known number of items but are too expensive to keep in memory as lists. I agree that it's ugly with plain lists though.

I can change it so that iterable lengths are inferred automatically (with `len()` on those iterables which support it), making `lengths` an optional kwarg.

mbugert:
Looks much cleaner now. WDYT?

bbayles:
Thanks!

Issue #532, opened 2021-07-10
repeat_each


FinalSh4re:
I hope I didn't overlook any already existing itertool that might cover this.

### Description
The iterator repeats each element of the passed iterable _n_ times.

### References
[https://stackoverflow.com/questions/24225072/repeating-elements-of-a-list-n-times](https://stackoverflow.com/questions/24225072/repeating-elements-of-a-list-n-times)

### Examples
I usually implement it a little differently from the version in the SO thread:

    def repeat_each(iterable, n=2):
        # repeat_each('ABC', 3) -> A A A B B B C C C
        return chain(*map(repeat, iterable, repeat(n)))

#### Possible variation
A possible variation of the above iterator would be incrementally repeated elements. However since there are probably only niche applications for this I didn't want to open a separate dedicated issue.

    def repeat_incrementally(iterable, start=1, step=1):
        # repeat_incrementally('ABC', 1, 3) -> A B B B B C C C C C C C
        return chain(*map(repeat, iterable, count(start, step)))

bbayles:
Feel free to open a PR with your implementation - looks good!

PR #533, opened 2021-07-10
repeat_each itertool

FinalSh4re:
### Issue reference
#532 
I decided to combine both functions from the issue, since the `repeat_incrementally()` function also covers the `repeat_each()` one's functionality by setting the step size to 0. However, since I believe having constant repeats is more commonly used I set the incremental step default to 0 (constant repeating of each element).

### Changes
Adding `repeat_each` itertool.

This is my first PR, please let me know if anything is missing or needs to be changed. I tried to mimic previous PRs made by other people, but I might have missed something. Especially the stub file I'm not sure about. Thanks!


bbayles:
Oh, I should have specified: I like this without the `step` argument. If you can drop back to the simpler version, I'd be happy to merge that.

bbayles (changes requested):
* See the comment about about the `step` argument
* Please do fix the failing tests

FinalSh4re:
I changed it back to the initial version from the issue. As for the failing tests, how do I add the missing dependencies it's asking for? Sorry, I'm completely new to this.

bbayles:
The issue is this:

``
would reformat more_itertools/more.pyi
would reformat more_itertools/more.py
Oh no! 💥 💔 💥
2 files would be reformatted, 9 files would be left unchanged.
```

You should be able to fix by installing `black` (`pip install black`) and then running it (`black .`).

bbayles:
OK, last requests: could you write tests for what you expect to happen for (a) empty iterator, (b) repeat set to 0?

bbayles (approved):

FinalSh4re:
Thanks for the patience and the help. 

One more thing, I just noticed that there is a massive problem with my implementation. For infinite iterators the method will break, since `*map()` unpacks the resulting iterator before passing it to `chain()`.

Eg. 

    repeat_each(count(), 3)

would crash the program.  

I think a better alternative would be using `chain.from_iterable()`:

    def repeat_each(iterable, n=2):
        return chain.from_iterable(map(repeat, iterable, repeat(n)))

This should (1) fix the problem with infinite iterators and (2) reduce memory usage in general since the map iterator doesn't get unpacked but consumed lazily instead. What do you think?

bbayles:
I'll take a PR with the fix and a test demonstrating the behavior is better. Thanks!

sassbalint:
@FinalSh4re This PR helped me a lot in creating my own first PR. #542 Thanks! :)

PR #534, opened 2021-07-12
Changed repeat_each() implementation to accept infinite iterators as input

FinalSh4re:
### Changes

As mentioned in #533 changed `repeat_each()` implementation to use `chain.from_iterable(map)` instead of `chain(*map)`. Since unpacking (`*map()`) consumes the whole iterator upfront it doesn't work for infinite iterators. Using `chain.from_iterable()` on the other hand consumes the passed iterator lazily, resulting in (1) the possibility of infinite iterators as input as well as (2) less memory usage for large inputs since the map iterator doesn't need to be loaded into memory before being passed to `chain`.

Also mentioned here:
https://stackoverflow.com/questions/39533052/difference-between-chainiterable-vs-chain-from-iterableiterable

### DIFF
--- a/tests/test_more.py
+++ b/tests/test_more.py
@@ -1693,17 +1693,7 @@ class RepeatEachTests(TestCase):
 
     def test_infinite_input(self):
         repeater = mi.repeat_each(cycle('AB'))
-        actual = []
-        counter = 0
-
-        while True:
-            current = next(repeater)
-            actual.append(current)
-            counter += 1
-
-            if counter == 6:
-                break
-
+        actual = mi.take(6, repeater)
         expected = ['A', 'A', 'B', 'B', 'A', 'A']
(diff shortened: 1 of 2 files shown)


tests/test_more.py
```diff
@@ -1690,6 +1691,22 @@ def test_negative_repeat(self):
...
+        counter = 0
+
+        while True:
+            current = next(repeater)
+            actual.append(current)
+            counter += 1
+
+            if counter == 6:
+                break
+
```
bbayles, line 1706: I think this whole block can just be `mi.take(6, repeater)` ?

bbayles (changes requested):
See comment about test

tests/test_more.py
FinalSh4re: Yes, indeed. I have no idea where my head was at.

Issue #535, opened 2021-07-27
`chunked` with chunks as even as possible


valtron:
### Description

Like `chunked`, but the yielded lists all have lengths within 1 of each other.

### References

I've had to implement this before at least twice, but I don't remember the details... probably for some machine learning thing when I had to read data from a large file and I wanted to split it up into some other files, but I didn't want one of them to end up significantly smaller than the other. It's not hard to implement, but it would be nice to have in more-itertools.

### Examples

API could be either a new function or a flag to `chunked`, I'll pick new function arbitrarily:

```python
>>> list(chunked_even([1, 2, 3, 4, 5, 6, 7, 8], n = 4))
[[1, 2, 3, 4], [5, 6, 7, 8]]
>>> list(chunked_even([1, 2, 3, 4, 5, 6, 7, 8, 9], 4))
[[1, 2, 3], [4, 5, 6], [7, 8, 9]]
>>> list(chunked_even([1, 2, 3, 4, 5, 6, 7, 8, 9, 10], 4))
[[1, 2, 3, 4], [5, 6, 7], [8, 9, 10]]
```

It's similar to `divide`, except the second argument gives the desired number of elements per list, not the number of lists.

It can be implemented using `n^2` of temporary storage (since the last `n` lists might be affected when a new element is read).

bbayles:
Thanks for the writeup. Afew questions to help me think about this:
* Is there some obvious proof that this is always possible? Can impossible cases be detected?
* `distribute` does something similar, but doesn't preserve order. Is that key to this?

valtron:
The proof was a fun little puzzle :) The one I came up with is for a method that knows the length, so the lists it would produce might be different than the "online" version, but at least it shows it's always possible to divide it so all lists' lengths are within 1 of each other.

1. Let `N` = total length of the sequence, `n` the desired length of each list.
2. `L = ceil(N / n)`. `n' = N / L`.
3. Form sequence `0 n', 1 n', 2 n', ..., k n', ..., L n'`. The outputted lists are slices `seq[ceil(k n'):ceil((k+1) n')]` for `k` in `{0, ..., L - 1}`.
4. The length of some outputted list is given by `ceil((k+1) n') - ceil(k n')`. Decompose `n'` into integer and fractional parts, `n' = i + f`, and after some rewriting that gives the length to be `i + ceil(k f + f) - ceil(k f)`.
[...]

PR #537, opened 2021-07-31
Add chunked_even

valtron:
### Issue reference

more-itertools/more-itertools#535

### Changes

Adds `chunked_even`, which works similar to `chunked`, but if the length of the list is not evenly divisible into lists of size `n`, the lists that are yielded are all within length 1 of each other. Examples:

```
>>> list(chunked_even([1, 2, 3, 4, 5, 6, 7, 8], n = 4))
[[1, 2, 3, 4], [5, 6, 7, 8]]
>>> list(chunked_even([1, 2, 3, 4, 5, 6, 7, 8, 9], 4))
[[1, 2, 3], [4, 5, 6], [7, 8, 9]]
>>> list(chunked_even([1, 2, 3, 4, 5, 6, 7, 8, 9, 10], 4))
[[1, 2, 3, 4], [5, 6, 7], [8, 9, 10]]
```

Versus `chunked` which would give:
```
>>> list(chunked([1, 2, 3, 4, 5, 6, 7, 8], n = 4))
[[1, 2, 3, 4], [5, 6, 7, 8]]
>>> list(chunked([1, 2, 3, 4, 5, 6, 7, 8, 9], 4))
[[1, 2, 3. 4], [5, 6, 7, 8], [9]]
>>> list(chunked([1, 2, 3, 4, 5, 6, 7, 8, 9, 10], 4))
[[1, 2, 3. 4], [5, 6, 7, 8], [9, 10]]
```


more_itertools/more.py
```diff
@@ -3900,3 +3901,71 @@ def __next__(self):
...
+                break
+    
+    # Yield num_partial lists of partial_size
+    for x in iterator:
+        buffer.append(x)
+        if len(buffer) == partial_size:
+            yield buffer
+            buffer = []
+            num_partial -= 1
+            assert num_partial >= 0
```
bbayles, line 3971: Let's drop the assert here.

```suggestion
```

more_itertools/more.py
```diff
@@ -3900,3 +3901,71 @@ def __next__(self):
...
+            buffer = buffer[n:]
+    yield from _chunked_even_finite(buffer, len(buffer), n)
+
+
+def _chunked_even_finite(iterable, N, n):
+    if N < 1:
+        return
+    
+    # Every list is either size `full_size <= n` or `partial_size = full_size - 1`
+    num_lists = ceil(N / n)
```
bbayles, line 3943: Could you convert this to integer math? I'd rather not import the `math` module.

tests/test_more.py
```diff
@@ -4537,3 +4537,55 @@ def test_basic(self):
...
+        self._test_finite('ABCDEFG', 3, [['A', 'B', 'C'], ['D', 'E'], ['F', 'G']])
+
+    def _test_finite(self, seq, n, expected):
+        # Check with and without `len()`
+        self.assertEqual(list(mi.chunked_even(seq, n)), expected)
+        self.assertEqual(list(mi.chunked_even(iter(seq), n)), expected)
+
+    def test_infinite(self):
+        n = 4
+        def limited_count():
```
bbayles, line 4570: Can this be replaced with `islice(count(), 0, n * k + n * n)`?

bbayles (changes requested):
See human + GitHub Action comments

tests/test_more.py
valtron: `k` varies. This test is kind of like a "timeout", it's supposed to check that, if given an infinite iterable, it can (1) yield things without reading the whole sequence (look-ahead is finite), (2) yield things without reading too far ahead (look-ahead < k^2).

valtron:
Regarding the name: I just used `chunked_even` as a placeholder, but I like your suggestion of `chunked_uniform` more. Maybe we can do that, but also what are your thoughts on adding this is a flag to `chunked`: `chunked(iterable, n, uniform=True)`?

bbayles:
Go ahead and do `chunked_uniform`. I'll pass on making it a flag for `chunked`.

valtron:
I just noticed the new function `interleave_evenly`. Maybe for consistency it's best to stick with `even` then, or maybe `chunked_evenly`? I also see the word "even" is used in the docs for `distribute` and `divide` too.

bbayles (approved):
Thanks for the changes. I think the name is OK as-is. I will likely edit the docstring a bit when I merge this.

Thanks again for the suggestion!

PR #539, opened 2021-08-02
Chunked: Allow chunk size to be of size None.

PhilMacKay:
### Issue reference
more-itertools/more-itertools/issues/538

### Changes
I updated the stub for `chunked` to allow the chunk size parameter `n` to have value `None`.
To prevent confusion when using keyword parameter `strict = True`, I also added a check in the strict code block that raises a ValueError with a specific message. It may not be necessary, but then it would raise a value error when checking the chunk size, and I felt that the error message would be confusing.

I also added two unit tests, with and without setting the strict keyword.


tests/test_more.py
```diff
@@ -128,6 +134,19 @@ def f():
...
+    def test_strict_being_true_with_size_none(self):
+        """Test when ``n`` has value ``None`` and the keyword strict is True
+        (raising an exception).
+
+        """
+
+        def f():
+            return list(mi.chunked('ABCDE', None, strict=True))
+
+        self.assertRaisesRegex(
```
bbayles, line 146: Does this actually do what we expect? Is `f()` called?

PhilMacKay: Oups, thanks for the review! With this mistake, I have revealed that I didn't actually write the test before writing the code, in true TDD development 😜
With my new commit, it's working as it should!

bbayles (approved):

Issue #543, opened 2021-08-22
Add ``zip_with_scalars()`` function


kalekundert:
### Description
Would you be interested in adding a function that mostly works like `zip()`, except that it treats any non-iterable arguments as values that should be included in each output tuple (without affecting how the iterable arguments are zipped)?

### References
I needed a function like this when I was writing a function for reading keys from JSON files.  The most common way to call the function would be with one key and one file, but it was also important to allow multiple keys or files to be specified.  For example, consider these two files:
```
# foo.json
{"a": 1, "b": 2}
```
```
# bar.json
{"a": 3, "c": 4}
```
I needed to be able to do:

    >>> f("foo.json", "a")
    [1]
    >>> f("foo.json", ["a", "b"])
    [1, 2]
    >>> f(["foo.json", "bar.json"], "a")
    [1, 3]
    >>> f(["foo.json", "bar.json"], ["b", "c"])
    [2, 4]

This is pretty hard to write in a general way.  You can't just wrap the scalars in `itertools.repeat`, because you'll get an infinite loop if only scalars are specified.

### Examples
Here are some examples to show how it works:
```
>>> list(zip_with_scalars(1, 2))
[(1, 2)]
>>> list(zip_with_scalars(1, [2]))
[(1, 2)]
>>> list(zip_with_scalars(1, [2, 3]))
[(1, 2), (1, 3)]
>>> list(zip_with_scalars([1, 2], [3, 4]))
[(1, 3), (2, 4)]
```

Some special cases:
```
>>> list(zip_with_scalars("ab"))
[("ab",)]
>>> list(zip_with_scalars("ab", not_iterable=None))
[("a", "b")]
>>> list(zip_with_scalars([1], [2, 3], strict=True))
Traceback...
UnequalIterableError
```

### Implementation
Here's an implementation.  If you're interested in adding this itertool, I can turn this into a PR with tests and docs:
```
def zip_with_scalars(*objs, strict=False, not_iterable=(str, bytes)):
    from builtins import zip
    from operator import itemgetter
    from more_itertools import zip_equal

    def is_scalar(obj):
        if not_iterable and isinstance(obj, not_iterable):
            return True
        try:
            iter(obj)
        except TypeError:
            return True
        else:
            return False

    iterables = []
    formatters = []

    for obj in objs:
        if is_scalar(obj):
            # The double-lambdas are necessary to create a closure.
            formatters.append((lambda x: lambda _: x)(obj))
        else:
[...]

PR #545, opened 2021-08-30
Implement zip_broadcast

bbayles:
Re: https://github.com/more-itertools/more-itertools/issues/543, this PR adds `zip_broadcast`. See the issue for discussion.


bbayles:
Tests for 3.10 are failing due to https://github.com/python/typing/issues/865.

kalekundert:
Thanks for implementing this; it does do what I want.  I have a few comments, though:

1. The function returns the wrong value when `*objects` is empty.  It yields a single empty tuple, when it should yield nothing.
2. I think `zip_fill()` is a better name than `zip_broadcast()`.  The reason is that the ability to support 2- and 3-dimensional inputs is a big part of what `np.broadcast()` is used for, while this is function is specifically for 1-dimensional inputs.
3. Both this function and `always_iterable()` have the same test to see if an object is iterable.  Would it make sense to factor this into it's own publicly-available function?

Here are the tests I wrote for my implementation.  I have some test cases that you don't, so you might want to include them:
```
@pytest.mark.parametrize(
        "objs, kwargs, expected", [(
            # 0 items
            [],
            {},
            [],
        ), (
            [[]],
            {},
            [],
        ), (
            # 1 item
            [1],
            {},
            [(1,)],
        ), (
            [[1]],
            {},
            [(1,)],
        ), (
            [[1], 2],
            {},
            [(1, 2)],
        ), (
            [1, [2]],
            {},
            [(1, 2)],
        ), (
            [[1], [2]],
            {},
            [(1, 2)],
        ), (
            # 2 items
            [[1, 2]],
            {},
            [(1,), (2,)],
        ), (
            [[1, 2], 3],
            {},
            [(1, 3), (2, 3)],
        ), (
            [1, [2, 3]],
            {},
            [(1, 2), (1, 3)],
        ), (
            [[1, 2], [3, 4]],
            {},
            [(1, 3), (2, 4)],
        ), (
            # not iterable
            ['ab'],
            {},
            [('ab',)],
        ), (
            ['ab'],
            dict(not_iterable=None),
            [('a',), ('b',)],
        ), (
            [[1, 2]],
            dict(not_iterable=list),
            [([1, 2],)],
        ), (
            # unequal length
            [[], [1]],
            {},
            [],
        ), (
            [[1], [2, 3]],
            {},
            [(1, 2)],
        ), (
            # infinite
            [1, repeat(2), [3]],
            {},
            [(1, 2, 3)],
        )],
)
def test_zip_with_scalars(objs, kwargs, expected):
    assert list(zip_with_scalars(*objs, **kwargs)) == expected

@pytest.mark.parametrize(
        "objs, kwargs, error", [(
            # 0 items
            [[], [1]],
            dict(strict=True),
            UnequalIterablesError,
        ), (
            [[1], [2, 3]],
            dict(strict=True),
            UnequalIterablesError,
        )],
)
def test_zip_with_scalars_err(objs, kwargs, error):
    with pytest.raises(error):
        list(zip_with_scalars(*objs, **kwargs))
```

kalekundert:
`zip_repeat()` might be another reasonable name.

bbayles:
Thanks for the review. I'm going to go with the `zip_broadcast` name. I considered doing an `_is_iterable` function, but I'll wait for a third consumre of it. Your point about the empty input was right on.

PR #549, opened 2021-09-01
fix local docs build

### LINKED ISSUE #548: local documentation builds using old release
To build the docs locally, I run `cd docs; make clean; make html`. This completes, but sphinx throws an unpleasant error stack because it cannot find any functions which were addd since the new release (autodoc failure). For example, with the current version these functions are: `chunked_even`, `repeat_each`, `interleave_evenly`, `zip_broadcast`, `map_if`. This means that these functions don't appear in the built documentation.

I believe this is because `docs/conf.py` imports `more_itertools` before adding the parent directory to the path. Switching the two around fixes the problem.

MarcinKonowalczyk:
Fixes #548

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -14,12 +14,12 @@ import sys, os
 
 import sphinx_rtd_theme
 
-import more_itertools
-
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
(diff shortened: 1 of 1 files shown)


MarcinKonowalczyk:
Thought: Should the CI pipeline include docs checks? I think this would have been caught then.

MarcinKonowalczyk:
also, is it worth updating the year in `conf.py` on line 47 (the copyright year) ?

bbayles:
Thanks for the PR. I think I'd rather suppress the copyright footer entirely than have to keep updating the year? PR welcome on that.

Issue #541, opened 2021-08-07
Document how `iter_except` can accept multiple exceptions


cool-RR:
This should be both in the docstring and the docs: `iter_except(whatever, (TimeoutError, StopIteration))`.

bbayles:
PR welcome - a test would be good also.

PR #551, opened 2021-09-02
Changed readme links to relative

MarcinKonowalczyk:
While looking at issue #541, I've noticed that locally built documentation has hyperlinks to the online one. While this works when documentation is online, it's not ideal for offline build. Moreover, when browsing legacy documentation online, the links take one to the `stable` documentation and, arguably, they shouldn't. This is a minor issue, so I thought I'll just submit a PR and see what happens.

Sphinx's linkcheck is struggling a bit with the relative links (probably because of the order in which the pages are generated. see [this](https://github.com/sphinx-doc/sphinx/issues/9383) issue), so I've selectivelly silenced them.

I've also taken the liberty to make the .rst table a bit less wide.


MarcinKonowalczyk:
I just realised that this means links on GitHub might not work quite correctly. I'll leave the PR open for review for a time being though.

bbayles:
Yeah, I'd like the GitHub page to keep working. If there's a way to make them both work, I'm all for it. Please file an issue?

bbayles (changes requested):
Sadly this will make the GitHub front page not work.

MarcinKonowalczyk (commented):
How about a preprocessor step in conf.py (see new commits) ?

MarcinKonowalczyk:
git into a bit of a fight with GitHub workflow. sorry, not something I could have fixed locally. turns out that 1) when creating files with python they *need* to be opened `open(..., 'w+')` mode and 2) the entire directory tree must already exist.

docs/conf.py
```diff
@@ -21,6 +21,39 @@
...
+
+import re
+
+with open('../README.rst', 'r+') as source:
+    readme_file = source.readlines()
+
+# Make sure build directory exists. This is required for github workflow.
+build_dir = '_build'
+if not os.path.exists(build_dir):
+    os.makedirs(build_dir)
```
bbayles, line 34: Can this be `os.makedirs(build_dir, exist_ok=True)`?

docs/conf.py
```diff
@@ -21,6 +21,39 @@
...
+
+# Change absolute links in README.rst to relative ones. This way the online
+# documentation copmiles with relative links, while README.rst on GitHub can
+# have absolute links. See issue #551.
+root_path = 'https://more-itertools.readthedocs.io/en/stable/'
+in_table = False
+with open(os.path.join('.', build_dir, 'README.pprst'), 'w+') as target:
```
bbayles, line 41: Why is the `w+` needed here?

docs/conf.py
MarcinKonowalczyk: yup

docs/conf.py
MarcinKonowalczyk: apparently it's not. I've changed it to 'w'

Issue #553, opened 2021-09-14
The sdist doesn't contain type stubs


saaketp:
It seems that package_data declared in `setup.py` is only for bdist and it is ignored when building sdist.
See https://stackoverflow.com/a/14159430.

This is a problem because the conda-forge package is built from the sdist, https://github.com/conda-forge/more-itertools-feedstock/blob/master/recipe/meta.yaml.
So any conda-forge users are not able to benefit from the type stubs.

To include the type stubs in sdist, we need to include in `MANIFEST.in`
```
include more_itertools/py.typed
include more_itertools/*.pyi
```

saaketp:
Actually, after reading the other answers on that stackoverflow post, it seems that `include_package_data=True` is the culprit.
`include_package_data=True` is an alternative to `package_data=...`, it should not be used in addition to it. https://setuptools.readthedocs.io/en/latest/userguide/datafiles.html#data-files-support

Having `include_package_data=True` makes it necessary to include the data files in Manifest.in
So I think it would be better to just remove that one line from `setup.py`.

bbayles:
Thanks for the report. Are you up for making a PR to fix?

saaketp:
Yes, I can make a PR that removes `include_package_data=True` part.

I'm also thinking that maybe the MANIFEST.in can also be removed.
Because it was really required only for sdist because we had that `include_package_data` option in `setup`.
Once that is removed, we can just rely on the manifest that is auto-generated by setuptools. I checked and it includes all the files mentioned in the manually specified manifest.

So should I also remove MANIFEST.in?
Or do you want me to just change that `include_package_data` in setup?

saaketp:
So I created a PR with the `include_package_data` change and also deleted MANIFEST.in there in a separate commit.
I can remove the second commit if we need to keep the current manifest around for some reason.

PR #554, opened 2021-09-14
Remove include_package_data option from setup.py to fix type stubs not included in sdist.

saaketp:
### Issue reference
more-itertools/more-itertools#553

### Changes
Remove include_package_data option from setup.py
The package_data option will be used for both sdist and bdist with this change.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -35,7 +35,6 @@ setup(
     license='MIT',
     packages=['more_itertools'],
     package_data={'more_itertools': ['py.typed', '*.pyi']},
-    include_package_data=True,
     python_requires='>=3.5',
     url='https://github.com/more-itertools/more-itertools',
     classifiers=[


saaketp:
I removed the change in MANIFEST.in, the correct files were not included in the sdist actually if MANIFEST.in is removed.
Earlier when I though the correct files were being included, it was by mistake because of a previous build in the same folder.

So now this is just that one line change.

saaketp:
Thanks @bbayles .
Now have to wait for a new release to get type stubs in the conda-forge build.
Can there be a new release soon?

bbayles:
Here is the release:
https://github.com/more-itertools/more-itertools/releases/tag/v8.10.0

saaketp:
Thanks, it landed in conda-forge too now.
And it works well, no complaints about missing type stubs.

PR #565, opened 2021-10-05
Reimplement zip_broadcast() to fix #561

### LINKED ISSUE #561: `zip_broadcast(strict=True)` fails if the leftmost iterables are exactly 1 item longer than the others
Here's the bug:
```
>>> mi.__version__
'8.10.0'
>>> list(mi.zip_broadcast([1, 2], [3], strict=True))
[(1, 3)]
>>> list(mi.zip_broadcast([1, 2], [3], [4], strict=True))
[(1, 3, 4)]
>>> list(mi.zip_broadcast([1, 2], [3, 4], [5], strict=True))
[(1, 3, 5)]
```
Here are some cases where the expected exception is raised:
```
>>> list(mi.zip_broadcast([1], [2, 3], strict=True))
  ...
more_itertools.more.UnequalIterablesError: Iterables have different lengths
>>> list(mi.zip_broadcast([1], [2, 3], [4], strict=True))
  ...
more_itertools.more.UnequalIterablesError: Iterables have different lengths
>>> list(mi.zip_broadcast([1], [2], [3, 4], strict=True))
  ...
more_itertools.more.UnequalIterablesError: Iterables have different lengths
>>> list(mi.zip_broadcast([1], [2, 3], [4, 5], strict=True))
  ...
more_itertools.more.UnequalIterablesError: Iterables have different lengths
>>> list(mi.zip_broadcast([1, 2], [3], [4, 5], strict=True))
  ...
more_itertools.more.UnequalIterablesError: Iterables have different lengths
```

I think the problem is that `zip()` consumes extra elements from the iterables given to it, so checking which iterables have been exhausted after calling `zip()` is unreliable.  I don't see an easy way to fix this, but the implementation I provided in #543 does handle these cases correctly.

kalekundert:
### Issue reference

Fixes more-itertools/more-itertools#561

### Changes
This implementation delegates the `strict=True` checks to `zip_equal()`, rather than making those checks itself.  This avoids the need to account for how `zip()` consumes items.


kalekundert:
While trying to get the tests to pass, I ran into a small complexity regarding what kind of exception to raise if the iterables don't match lengths.  As mentioned above, the implementation delegates to `zip_equal()`, which raises `UnequalIterablesError`.  However, for python>=3.10, `zip_equal()` also emits a `DeprecationWarning`.  To avoid this warning, the implementation delegates directly to `zip()` if possible.  However, `zip(strict=True)` raises `ValueError` instead of `UnequalIterablesError`.  

I decided to handle this by updating the documentation to state that `zip_broadcast()` raises `ValueError`.  `UnequalIterablesError` is a subclass of `ValueError`, so this is true for all versions of python.  I also note that the specific exception raised in older versions of python is `UnequalIterablesError`.

I think this approach is consistent with the idea of phasing out `zip_equal()` for python >=3.10, but it also breaks backwards compatibility very slightly.  Let me know if you'd rather just raise `UnequalIterablesError` for all versions of python.

bbayles:
Thanks for the PR. I merged a combination of your version and mine that passes tests. I also worked around the deprecation warning such that 3.10 can raise `UnequalIterablesError` without version detection.

Issue #512, opened 2021-03-17
Feature suggestion: find_duplicates


OrBin:
The `all_unique` itertool is useful, but sometimes you need not only to find whether all the elements of iterable are unique, but also to find which element are not.

For this, I suggest adding a `find_duplicates` itertool, that gets an iterable and an optional key (just as `all_unique`, `unique_justseen` and `unique_everseen`), and returns an iterable of the duplicate values.
I thought of something like the following implementation (it's an initial suggestion though, but if you like the suggestion I will go on and try to improve it):

```python
def find_duplicates(iterable, key=None):
    seenset = set()
    seenset_add = seenset.add
    seenlist = []
    seenlist_add = seenlist.append
    dupset = set()
    dupset_add = dupset.add
    duplist = []
    duplist_add = duplist.append

    for element in map(key, iterable) if key else iterable:
        try:
            if element not in seenset:
                seenset_add(element)
            elif element not in dupset:
                dupset_add(element)
                yield element
        except TypeError:
            if element not in seenlist:
                seenlist_add(element)
            elif element not in duplist:
                duplist_add(element)
                yield element
```

The suggested implementation relies much on the existing implementation of `all_unique` and `unique_everseen`.

What do you think?

bbayles:
Thanks for the suggestion.

My go-to for this would be `collections.Counter`:

```python
from collections import Counter
iterable = ... # whatever
C = Counter(iterable)
[k for k, v in C.items() if v > 1]
```

Can you think of examples where this wouldn't work? I think in Python 3.6+ even the order would be preserved.

OrBin:
I think the main disadvantage of using a `Counter` is that it consumes the iterable and forces you to iterate over the counter. In case of finding duplicates, it frequently means that you iterate almost twice over the iterable (once for counting and once for finding which keys have `v > 1`).
My suggestion returns an iterable, so I believe it should be more efficient, and is also usable in case you want to find the first duplicate, for example (In some cases it may be enough to tell the user that their input is invalid and just show them the first duplicate value).

bbayles:
OK, I'll accept a PR that takes an iterable and emits the values that have appeared earlier.

OrBin:
Cool. Regarding the returned iterable, do you think it should emit the duplicate keys or the instead of values, in case the caller supplied a `key` parameter?

bbayles:
Definitely the original values.

PR #570, opened 2021-10-13
Add duplicates_everseen and duplicates_justseen

DavidPratt512:
### Issue reference

Resolves more-itertools/more-itertools#512

### Changes

Implements two functions analogous to the existing recipes `unique_everseen()` and `unique_justseen()`: `duplicates_everseen()` and `duplicates_justseen()`.

`duplicates_everseen()` is what one would typically think of if they were wanting to search for duplicates in an iterable. It yields duplicate elements after their first appearance in the iterable. For example
```python
>>> list(duplicates_everseen('ABBA'))
['B', 'A']
```

`duplicates_justseen()` is the analog to `unique_justseen()` in that it will only yield duplicate elements if they are grouped together. For example 
```python
>>> list(duplicates_justseen('ABBA'))
['B']
```

I was grappling whether to just implement a single `duplicates()` generator that would take the role of `duplicates_everseen()`, but I realized that I would have never thought to implement `unique_justseen()` if I was given the task of implementing a generator for unique values. So perhaps it's better if both duplicate-type generators are implemented. But I'm open to changing my mind, of course :)


bbayles:
Nice work - many thanks.

Issue #571, opened 2021-10-15
strictly_n should be lazy - return an iterator with a length hint


SnirBroshi:
The new `more_itertools.strictly_n` was suggested in #552 and implemented in #558 which is awesome, but it always consumes the `n` elements right-away and returns a list.

Instead, I think it should be lazy - return an iterator (optimally with a length hint, #510), yielding elements as usual, but if the given iterator ends prematurely raise `too_short`, and if we got to the `n`th element and there are more, raise `too_long`.

### Example
```py
>>> items = 'abcdefg'
>>> for item in strictly_n(items, 5):
...     print('Item', item)
Item a
Item b
Item c
Item d
Item e
ValueError: too many items in iterable (got at least 6)
```

I'm not sure whether the error should be thrown before or after the nth element is yielded.

The current behavior can then be achieved by calling `list()` on the returned iterator.

### Real world use case

I used a REST API to query a long list of items. This API used pagination - each request gave me a few items and a token to pass in the next request to get the next few items, and so on. The API also told me the total number of items.

The pagination was pretty weird, it sometimes repeated items, so I wanted to make sure I got all the items correctly, but I didn't want to wait to get all the items to start processing them, as it would take a long time and also each item was pretty big so I didn't want them all in memory at the same time.

I wrote the following:

```py
class ExplicitLengthIterable:
    def __init__(self, iterable, length):
        self.iterable = iterable
        self.length = length
    def __len__(self):
        return self.length
    def __iter__(self):
        iterator = iter(self.iterable)
        for i, value in zip(itertools.count(0), iterator):
            if i >= self.length:
                raise Exception('Iterable was longer than promised')
            yield value
        if i + 1 != self.length:
            raise Exception('Iterable was shorter than promised')
```

(I needed an iterable with `__len__` rather than an iterator with `__length_hint__`, but iterator is more generic and can be wrapped later)

---

Tagging @hwalinga since they might be interested.

NotWearingPants:
@bbayles What is the benefit of collecting the items to pass to these, why not just the amount of items?

So will a kwarg called `keep_items_for_callbacks=True` do it?

Note that both `more_itertools.one()` and `more_itertools.only()` do not pass any information to the `too_short` and `too_long` callbacks.

PR #573, opened 2021-10-20
Make strictly_n lazy

bbayles:
Re: #571, this PR makes `strictly_n` lazy.

Specifically:
* Items are yielded as they are encountered
* We don't raise (or call `too_short`) until the iterable is exhausted
* We don't raise (or call `too_long`) until we've seen `n` + 1 items
* `too_short` and `too_long` no longer get called with any arguments

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -621,7 +621,7 @@ def strictly_n(iterable, n, too_short=None, too_long=None):
 
     You can instead supply functions that do something else.
     *too_short* will be called with the number of items in *iterable*.
-    *too_long will be called with ``n + 1``.
+    *too_long* will be called with `n + 1`.
 
         >>> def too_short(item_count):
         ...     raise RuntimeError


bbayles:
> Why not pass the amount of elements encountered in the iterator to the too_long/too_short functions? That might be useful.

My thought was that in the `too_short` case, the user has already had the items `yield`ed to them, so they know how many they got. And in the `too_long` case, the number is always going to be `n` + 1.

> Docs might have some line about the fact that too_long/too_short functions are called when the complete iterator is iterated through and not upon construction of the generator. That's something that might be confusing.

Good suggestion.

> Why have too_long=None and check for if too_long is None: too_long = . Why not use default arguments right in the function definition?

Because I didn't want to put `raise_` in the docs.

> Docs might highlight the fact that more_itertools.raise_ exists to make the creation of the functions that raise some exception a little bit easier.

It's a nice function, but I'd rather not have people importing it from this library.

NotWearingPants:
I also vote for passing the number of elements to the error functions (and maybe the additional item in too_long), since it costs nothing, but currently to get the count you have to do something like this:

```py
def too_short():
    print(f'iterable had {n+1} items, expected 5')
for n, item in enumerate(more_itertools.strictly_n(iterable, 5, too_short=too_short)):
    process_item(item)
```
which looks pretty weird and confusing to me

Issue #586, opened 2021-12-05
ichunked caches items even when it doesn't need to


hjtran0:
ichunked makes tees for every chunk, which lets us iterate on chunks out of order, but also _always_ creates a cache within the chunk, even if we iterate through all the chunks in order.

PR #587, opened 2021-12-05
Fix inefficient use of `tee`s in `ichunked

hjtran0:
### Issue reference

more-itertools/more-itertools#586

### Changes
ichunked makes tees for every chunk, which lets us iterate on chunks out of order, but also always creates a cache within the chunk, even if we iterate through all the chunks in order.

Example:
```python
def big_string_generator():
    while True:
        yield 'X'*5000
ichunk = next(ichunked(big_string_generator(), 1000))
consume(ichunk) # You'll see a very large memory spike here
```

I've fixed by creating an auxiliary `_IChunk` class which holds a cache that we only manually fill after a new chunk is requested. If the previous chunk has been consumed by the time a new chunk is requested, then the cache will get filled with nothing (saving us memory)

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -3327,17 +3327,9 @@ def only(iterable, default=None, too_long=None):
     return first_value
 
 class _IChunk:
-    def __init__(self, iterable, n, prev_chunk):
-        if prev_chunk:
-            def it_slice():
-                prev_chunk.fillCache()
-                yield from islice(iterable, n)
-            it_slice = it_slice()
-        else:
-            it_slice = islice(iterable, n)
-        self._it = it_slice
+    def __init__(self, iterable, n):
+        self._it = islice(iterable, n)
         self._cache = deque()
-        self._prev_chunk = prev_chunk
 
     def fillCache(self):
         self._cache.extend(self._it)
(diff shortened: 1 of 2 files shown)


hjtran0:
I ran tests with python39 which is what I have installed. Tried running `tox` but I don't have 36/37/38 installed. Not super familiar with how to get them all on my computer so it'd take me some time to figure it out, but let me know if it's necessary.

Also feel free to note any stylistic issues. First contribution here :-)

bbayles:
Great work. Could you fix the linter errors, and change the `fillCache` method's name to `fill_cache`?

hjtran0:
Looks like there's a `pypy` test failure now. `pypy` might not implement `tracemalloc` which the unittest depends on. Is it okay with you if we just `try:... except ModuleNotFoundError: return` in that test to account for pypy?

hjtran0:
How do you programmatically check whether we’re running with pypy?

bbayles:
Check out the platform module; I think it should work on the versions of
Python we need.

hjtran0:
Somehow took 6 commits but it's successfully skipped now.

bbayles (approved):

seanmacavaney:
Awesome fix! Thanks!

PR #594, opened 2022-01-26
Switch to flit build system

### LINKED ISSUE #593: Please consider switching to flit_core
I'm the @Gentoo maintainer for core Python packages. We're currently considering options for bootstrapping setuptools while unvendoring its dependencies in the PEP 517 world. One of the interesting options is to use [flit](https://flit.readthedocs.io/en/latest/) as it has much fewer dependencies than setuptools and we've got its bootstrap trivially covered.

Using flit as more-itertools' build system would mean that we can cleanly install it before building setuptools. Flit can also generate a backwards-compatible `setup.py` (or be used alongside `setup.py`, though for this to work for us `pyproject.toml` would have to specify flit_core as the `build-backend`). FWICS it is becoming increasingly popular these days.

Would you be interested in this? If yes, then I can prepare a PR either converting the project to flit entirely, or adding it alongside existing `setup.py`, as you prefer.

mgorny:
### Issue reference
Fixes #593

### Changes
Converted to the flit build system, added link to change history in `README.md`, updated project files for flit.


mgorny:
The usual process is that `flit build --setup-py` generates it on the fly while creating the sdist tarball. However, I can add it directly to the repo if you prefer.

bbayles:
> However, I can add it directly to the repo if you prefer.

Yes, thanks!

mgorny:
Added.

bbayles:
Mildly disappointing that the generated one misses some of the things defined in the TOML file.

Let's just revert the last commit and I'll live without `setup.py`? This will fix the `black` and `flake8` issues.

mgorny:
> Mildly disappointing that the generated one misses some of the things defined in the TOML file.

Well, IIUC its purpose is to make the package installable with old tools, and they're not that concerned about getting it perfect.

> Let's just revert the last commit and I'll live without `setup.py`? This will fix the `black` and `flake8` issues.

Done. Do you prefer sdist archives including this generated `setup.py` or not?

bbayles (approved):
This is fine for what we need. Thanks!

PR #610, opened 2022-05-05
Replace list(islice()) by take() in docstrings

psarka:
### Issue reference
Trivial change, no issue

### Changes
`take` is awesome, let's use it in the docstrings :)


bbayles:
Thanks for the PR! I'm going to pass on this one - `take` is great, but `list` has the benefit of being universally recognized.

PR #623, opened 2022-07-10
Lazily import ThreadPoolExecutor

bbayles:
Re: https://github.com/pypa/setuptools/pull/3091 - I'm late in noticing this, but had I seen this issue in `setuptools` before I would have made this upstream change.

I'll ping @jaraco for comment. Feel free to request changes like this if something in `more_itertools` interferes with smooth operation of `setuptools`. I think it's awesome (and a little terrifying) to have this project included in such a critical package, and as long as that's the case want to make things work well.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -3636,6 +3636,7 @@ class callback_iter:
         self._aborted = False
         self._future = None
         self._wait_seconds = wait_seconds
+        # Lazily import concurrent.future
         self._executor = __import__(
             'concurrent.futures'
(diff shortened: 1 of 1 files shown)


jaraco:
Thanks for proposing the change. I’d recommend to add a comment or test to capture the motivation, as otherwise this change could be optimized away.

I’m honestly unsure this project should own this downstream concern. I can see other ways, such as a demand import system that would also address this issue.

Regardless, I appreciate the consideration.

PR #626, opened 2022-07-26
Use suppress

Masynchin:
### Changes

Use `contextlib.suppress` instead of `try: ... except: pass`. Tries with additional `else: ...` block haven't been changed.


bbayles:
Thanks for the PR. I'll pass on this; I'd rather not take on the additional import.

Issue #510, opened 2021-03-10
Tail and some others can work better if they'll use a length hints (pep 424)


code-of-kpp:
I understand, that recipes were ported directly from the docs, but things like tail do not perform great in some situations.

Note, that when `deque` is constructed it consumes almost everything from the iterator, so creation of `tail` iterator is not lazy.

A better possible implementation is:
```python
def tail(n, iterable):
    if isinstance(iterable, collections.abc.Sequence):
        return iter(iterable[-n:])
    hint: int = operator.length_hint(iterable)
    if hint != 0 and n >= hint:
        return itertools.islice(iterable, hint - n, hint)
    return iter(collections.deque(iterable, maxlen=n))
```

This will make things like

```python
>>> tail(100, range(1 << 30))
<range_iterator at 0x...>
>>> next(tail(100, range(1 << 30)))
1073741724
```
really fast, things like
```python
>>> tail(100, reversed(range(1 << 30))) # lazy
<itertools.islice at 0x...>
>>> next(tail(100, reversed(range(1 << 30))))
99
```
somewhat faster (but at least lazy).

And hopefully will "automagically" improve in the future for some other cases like `enumerate` and `map`.

podshumok:
Please, note, that this is not a PR, as `length_hint` is not guaranteed to be accurate (but it looks lite it is in practice), so actual implementation should consider other options of the length estimation, like `isinstance(obj, Sized)` and/or trusting length hints for types, that are known to have accurate hint.

bbayles:
Thanks for the suggestion. I'm open to the "if it's a Sequence, slice it" optimization - you can make a PR for that? But because the "length hint" isn't necessarily true, I'd rather not rely on it.

podshumok:
How about using `abc.Sized` to cover `set`s and `dict`s, for example and wrapping the call to `deque` in a `yeild` to force laziness?

podshumok:
Another approach is to wrap deque call in `map` and `chain` or something like that. This way the return type of the function can be a little more transparent,  e.g. `range_iterator` or `islice` in some cases.

Anyway this can be a breaking change as previously data were effectively copied and original iterator was exhausted before the next item picked from the `tail()`, forcing `StopIteration` and exiting loops with original iterator-parameter.

nanouasyn:
```python
if isinstance(iterable, collections.abc.Sequence):
    return iter(iterable[-n:])
```
[...]

PR #630, opened 2022-08-07
Optimize memory allocation for `tail` function

nanouasyn:
### Issue reference

more-itertools/more-itertools#510

### Changes

The `tail` function no longer allocates memory for elements if we never request elements from the tail. If it is possible to determine the size of an iterable without iterating over its elements, we will no longer load all the elements of the tail into memory.

### DIFF
--- a/more_itertools/recipes.py
+++ b/more_itertools/recipes.py
@@ -107,6 +107,10 @@ def tail(n, iterable):
     ['E', 'F', 'G']
 
     """
+    # If the given iterable has a length, then we can use islice to get its
+    # final elements. Note that if the iterable is not actually Iterable,
+    # either islice or deque will throw a TypeError. This is why we don't
+    # check if it is Iterable.
     if isinstance(iterable, Sized):
         yield from islice(iterable, max(0, len(iterable) - n), None)
     else:
--- a/more_itertools/recipes.py
+++ b/more_itertools/recipes.py
@@ -11,7 +11,7 @@ import operator
 import warnings
 
 from collections import deque
-from collections.abc import Sized
+from collections.abc import Collection
 from itertools import (
     chain,
     combinations,
@@ -107,7 +107,7 @@ def tail(n, iterable):
     ['E', 'F', 'G']
 
     """
-    if isinstance(iterable, Sized):
+    if isinstance(iterable, Collection):
         yield from islice(iterable, max(0, len(iterable) - n), None)
     else:
         yield from iter(deque(iterable, maxlen=n))
--- a/more_itertools/recipes.py
+++ b/more_itertools/recipes.py
@@ -11,7 +11,7 @@ import operator
 import warnings
 
 from collections import deque
-from collections.abc import Collection
+from collections.abc import Sized
 from itertools import (
     chain,
     combinations,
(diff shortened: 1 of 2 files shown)


bbayles:
Does `Sized` imply slice-able with `islice`? I was thinking we would need to use `Sequence`.

nanouasyn:
No, it doesn't. In the linked issue, I described the problem that occurs when getting a slice: we are losing laziness. But if you agree to allow this, I will add optimization with slices.

https://github.com/more-itertools/more-itertools/issues/510#issuecomment-1207219902

bbayles:
We lose laziness with `islice`? I'm suggesting changing `Sized` to `Sequence`, not changing to `iterable[:end]`

nanouasyn:
I'm sorry, I misunderstood you a little. `Sized` implies an iterable whose length is known. This is enough to lazily iterate through the sequence in all cases. `Sequence` also implies getting elements by indexes. There are not many cases in the standard library where a sequence is `Sized`, but is not a `Sequence`. We can take a slice of `set`, but it doesn't make sense because the elements inside have no order. However, `dict` preserves the insertion order, so it has sense. We can also assume the existence of linked lists that will be `Sized`, but not `Sequence`, because getting an element by index is very expensive, unlike storing the length.

bbayles:
> This is enough to lazily iterate through the sequence in all cases

Is this true? `Sized` is merely:

>     ABC for classes that provide the __len__() method.

Perhaps you're saying "if the user can use their object with this function at all, we know it provides `__iter__`, so if it also provides `__len__` then it can be used with `islice`?"

nanouasyn:
Oh, really. I was wrong. I always thought that `Sized` is a subclass of `Iterable`. Then, indeed, we cannot rely on the fact that every `Sized` will be `Iterable`. Сould we use `Collection`? This class imposes fewer restrictions than `Sequence`, although still more than I would like. We can also check the object for being `Sized` *and* `Iterable` instead.

bbayles:
After reading the [collections.abc source](https://github.com/python/cpython/blob/3.10/Lib/_collections_abc.py), I think we ought to either:
* Go back to your original approach (check for `Sized`) and add a comment that says something like: "if the given object can be used with this function , it must be iterable. If it also has a length, then we can use `islice` to get its final elements"
* Check for `isinstance(Iterable) and isinstance(Sized)`. This will make it clear what we're looking for without an additional comment, and I can't think of something that would pass both this check and work with `deque` but not `islice`.

nanouasyn:
You're right. Prior to optimization, if a user passed a non-`Iterable` object to the `tail` function, `deque` would throw TypeError, regardless of whether this object is `Sized`. After optimization, if a user passes a `Sized` but not `Iterable` object, and we pass it to `islice`, `islice` will throw the same `TypeError`. In other words, even if a user finds a `Sized` but not `Iterable` object somewhere (which will not be easy), we can safely pass it to `islice`, and this will not change the behavior of the `tail` function relative to the original implementation. It doesn't matter if `deque` or `islice` gets a non-`Iterable` object and throws a `TypeError`. So I'm returning an implementation with `Sized`.

bbayles:
Agree with the logic, and if you'll add a comment to this effect, I'll approve and merge

bbayles (approved):

PR #631, opened 2022-08-07
Fix an indentation level in docstrings according to black

nanouasyn:
### Changes

Removed extra indentation level in the docstrings.


nanouasyn:
I don't understand what the sphinx didn't like. Could you explain?

PR #632, opened 2022-08-08
Small fix of `islice_extended` test

nanouasyn:
### Changes

Previously, the test checked the same `step` 4 twice, although the author obviously meant -4. I fixed it. Also, all `start` and `stop` values from -4 to 9 were checked, although it is enough to check the usual values, values at the borders, and values beyond the borders. I changed that too and this test was performed twice as fast. It ain't much but it's honest work.

### DIFF
--- a/tests/test_more.py
+++ b/tests/test_more.py
@@ -3218,7 +3218,7 @@ class StripFunctionTests(TestCase):
 class IsliceExtendedTests(TestCase):
     def test_all(self):
         iterable = ['0', '1', '2', '3', '4', '5']
-        indexes = [-10, -6, -5, -4, -2, -1, 0, 1, 2, 4, 5, 6, 10, None]
+        indexes = [*range(-4, 10), None]
         steps = [1, 2, 3, 4, -1, -2, -3, -4]
         for slice_args in product(indexes, indexes, steps):
             with self.subTest(slice_args=slice_args):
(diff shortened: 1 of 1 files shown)


tests/test_more.py
```diff
@@ -3218,8 +3218,8 @@ def test_math(self):
 class IsliceExtendedTests(TestCase):
     def test_all(self):
         iterable = ['0', '1', '2', '3', '4', '5']
-        indexes = list(range(-4, len(iterable) + 4)) + [None]
-        steps = [1, 2, 3, 4, -1, -2, -3, 4]
+        indexes = [-10, -6, -5, -4, -2, -1, 0, 1, 2, 4, 5, 6, 10, None]
```
bbayles, line 3221: Why are we not testing between `-3` and `3` now?

nanouasyn: It seems to me that `3` and `-3` are unremarkable. We are testing `2` and `-2`, also the usual values. `0` is the edge case, `-1` and `1` are close to `0`. `5` and `-5` are the edge cases. `4`, `-4`, `5` and `-5` are close to these cases. `10` and `-10` are far beyond the borders.

bbayles: I'll approve and merge this if we put those back; I meant for that whole range to be tested. Missing `-4` was definitely a mistake, though!

bbayles (approved):

PR #633, opened 2022-08-09
Python 3.10 trove

chayim:
I noticed that CI runs against 3.10, but that there's no trove available in PyPi. Given that we wanted to use this library, and poetry dies, I thought this PR prudent.

For reference, we will use it here: https://github.com/redis/redis-om-python/pull/305


bbayles:
Version [8.14.0](https://pypi.org/project/more-itertools/8.14.0/) is now on PyPI. Thanks!

chayim:
Amazing, thanks @bbayles

Issue #635, opened 2022-08-09
Optimization of `islice_extended` for iterables with fixed length


nanouasyn:
The same optimization for `islice_extended` as in this issue for `tail`:

https://github.com/more-itertools/more-itertools/issues/510

If start or stop is negative, but the length of the iterable is known, we can adjust start and stop to be positive and use `islice`.

PR #636, opened 2022-08-09
Optimize `islice_extended` for `Sized` iterables

nanouasyn:
### Issue reference

more-itertools/more-itertools#635

### Changes

The `islice_extended` with positive `step` no longer allocates memory for elements if `start` or `stop` is negative, but the iterable is `Sized`. I raised the usual case check to the top, because an attempt to apply optimization should occur before the non-optimized processing of negative start/stop is applied, but I would like the `isinstance` check not to be performed if the case is normal.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -2437,11 +2437,10 @@ class islice_extended:
     """
 
     def __init__(self, iterable, *args):
-        it = iter(iterable)
         if args:
-            self._iterable = _islice_helper(it, slice(*args))
+            self._iterable = _islice_helper(iterable, slice(*args))
         else:
-            self._iterable = it
+            self._iterable = iter(iterable)
 
     def __iter__(self):
         return self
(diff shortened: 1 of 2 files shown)


nanouasyn:
I made a mistake, my change does not affect anything. I will redo it.

nanouasyn:
Phew, it was hard for me, but I coped. Optimization works now. Moreover, I have added optimization for negative steps. I also expanded the test cases because I noticed that I made a mistake that the test did not notice.

more_itertools/more.py
```diff
@@ -2437,11 +2437,10 @@ class islice_extended:
     """
 
     def __init__(self, iterable, *args):
-        it = iter(iterable)
         if args:
-            self._iterable = _islice_helper(it, slice(*args))
+            self._iterable = _islice_helper(iterable, slice(*args))
```
bbayles, line 2441: Why do we need to change variable names here (and below)? I'd rather not have this in the diff.

more_itertools/more.py
```diff
@@ -2456,19 +2455,31 @@ def __getitem__(self, key):
         raise TypeError('islice_extended.__getitem__ argument must be a slice')
 
 
-def _islice_helper(it, s):
-    start = s.start
-    stop = s.stop
-    if s.step == 0:
+def _islice_helper(iterable, s):
+    start, stop, step = s.start, s.stop, s.step
```
bbayles, line 2459: Same question here - why switch to tuple unpacking here? It's not really related to the change being proposed.

more_itertools/more.py
```diff
@@ -2456,19 +2455,31 @@ def __getitem__(self, key):
...
+    if step == 0:
         raise ValueError('step argument must be a non-zero integer or None.')
-    step = s.step or 1
+    step = step or 1
 
     if step > 0:
-        start = 0 if (start is None) else start
```
bbayles, line 2467: Ditto here; let's please not change things just to change them. The style enforcement for this repo is done by `black` and `flake8`.

more_itertools/more.py
nanouasyn: Previously, an iterator was extracted from `iterable` to the `it` variable, and then this `it` was passed to `_islice_helper`. Now, `iterable` itself is passed to `_islice_helper`. This is important because if we pass an iterator, we won't be able to determine that it belongs to a `Sized` iterable, and we will not be able to apply optimization. For this reason, I changed the name of the argument from `it` to `iterable` so that it could be seen that an iterable is passed, not an iterator. But some conditional branches need an iterator of iterable. Therefore, in these branches I get an iterator and put it in the variable `iterator'. If we had left `it', it might have been confusing. Should I rename `iterator` to `it`?

more_itertools/more.py
nanouasyn: Sorry, I have changed this and what you mentioned above just because I edited the code nearby :) OK, I'll roll back these minor changes.

nanouasyn:
Perhaps you have some other comments about the code? Shouldn't I do a rebase and force push so that all the changes are in one commit and the history is less confusing?

bbayles:
It took a long time to get the logic for this right (and I shipped a subtly wrong version at one point), so I'm wary of changes and it will take me a while to review things fully. No need to rebase.

bbayles:
Sorry, I'm going to pass on this additional feature. I don't think there's anything wrong with the PR, but the additional complexity is something I'm not inclined to take on.

I will incorporate your additional tests and credit you for those in the next release. Thanks, and apologies again!

Issue #616, opened 2022-05-23
set_partitions ignoring duplicates


boomboomjassi:
### Description
Configure set_partitions to ignore duplicates whenever there is a repeat element

### References
`set_partitions` works fine when all the element in the source list/iterator are distinct. But whenever there is a repeat it generates redundant sets.

### Examples
```
lst = [2, 2,  3, 5]
pp.pprint(list(mit.set_partitions(lst, 2)))
[[[2], [2, 3, 5]],
 [[2, 2], [3, 5]],
 [[2], [2, 3, 5]],
 [[2, 2, 3], [5]],
 [[2, 3], [2, 5]],
 [[2, 3], [2, 5]],
 [[3], [2, 2, 5]]]
```
As we can see row 0 and row 2(zero indexed) are identical  ie `[[2], [2, 3, 5]],`
`set_partitions` needs an argument to ignore duplicates. Don't you think so?

bbayles:
This is in line with what `itertools.combinations` and `itertools.permutations` do. Is it a problem for some application?

boomboomjassi:
I don't quite get how can I achieve the same using `itertools.combinations` and `itertools.permutations`.
Maybe if you can help me out with a code snippet using `itertols.combinations` and `itertools.permutations` it would be great.

Just to reiterate my problem again to avoid any confusion, basically all I need is 
```
[[[2], [2, 3, 5]],
 [[2, 2], [3, 5]],
 [[2, 2, 3], [5]],
 [[2, 3], [2, 5]],
 [[3], [2, 2, 5]]]
```
instead of 
```
[[[2], [2, 3, 5]],
 [[2, 2], [3, 5]],
 [[2], [2, 3, 5]],
 [[2, 2, 3], [5]],
 [[2, 3], [2, 5]],
 [[2, 3], [2, 5]],
 [[3], [2, 2, 5]]]
```
My original requirement was simple **Partition a given list of numbers into two or more unordered sets**. Which essentially means neither the order of sets matter nor is the order of elements within any set. Here is the original [stackoverflow question](https://stackoverflow.com/questions/72334282/partition-a-given-list-into-two-or-more-unordered-sets/72348893#72348893) where I do for all K=1 to len(lst).

bbayles:
I meant that `itertools.permutations` and `itertools.combinations` treat duplicate elements the same way; they are represented in the output the same way as they are in `set_partitions`.

[...]

PR #655, opened 2022-10-26
Add unique option for set_partitions function

TakenBrandi:
### Issue reference
Addresses #616 

### Changes
* Adds an optional unique argument to the `set_partitions` function


bbayles:
Hmm, I think what this PR is doing is passing the existing results through `unique_everseen`. But that isn't stopping the non-unique items from being generated, is it? I think that if this is to be a valuable feature addition, we would need to have an algorithm that efficiently generated only the unique items.

To illustrate the difference, try `list(more_itertools.unique_everseen(itertools.permutations('mississippi')))` vs. `list(more_itertools.distinct_permutations('mississippi'))`.

TakenBrandi:
You're right. It's just a filter. I'll dig into the algorithm instead.

ktbarrett:
I'm not sure I'm sold on this as a flag. You can either use `set_partitions(set(values))` if you need to uniquify the values in the input, or (based on what's in the examples) what you are looking for is a multiset partition. That would be a completely different algorithm that's aware of the values and number of each.

ktbarrett:
Algorithm M in Volume 4A of The Art Of Computer Programming is apparently a multiset partition algorithm. I have it at home, so I'll have a look in a couple of days.

PR #656, opened 2022-10-29
Use postponed evaluation of annotations (PEP 563).

Isira-Seneviratne:
### Changes

Make use of postponed annotation evaluation ([PEP 563](https://peps.python.org/pep-0563/)), which allows for use of the type hinting syntax introduced in Python 3.9 and 3.10 ([PEP 585](https://peps.python.org/pep-0585/) and [604](https://peps.python.org/pep-0604/)) on older Python versions.


bbayles:
I don't think I can accept this while I still support versions prior to 3.9, right?

Isira-Seneviratne:
> I don't think I can accept this while I still support versions prior to 3.9, right?

Actually, the idea is to support the newer type hints on Python versions prior to 3.9/3.10, sorry for the confusion. I'll update the description.

bbayles:
Ahh, the `__future__` statement goes back to 3.7.

PR #665, opened 2023-01-07
Add gray_product

haukex:
I would like to propose the addition of the iterator `gray_product`, which produces the same output as `product`, but in a different order: like the binary [Gray code](https://en.wikipedia.org/wiki/Gray_code), only one item in the output changes from one iteration to the next. This can be very useful in testing, as changing only one input at a time can help reveal certain errors.

```
>>> from more_itertools import gray_product
>>> list(gray_product("AB", (1, 2, 3)))
[('A', 1), ('B', 1), ('B', 2), ('A', 2), ('A', 3), ('B', 3)]
```


bbayles (approved):

bbayles:
Great idea! Many thanks; I'll merge this shortly.

haukex:
That's great, thank you very much!

PR #667, opened 2023-01-10
Make sliced return `Iterator[T]` instead of `Iterator[Sequence[T]]`

### LINKED ISSUE #666: `sliced` should preferably return `Iterator[T]` over `Iterator[Sequence[T]]`
https://github.com/more-itertools/more-itertools/blob/401a2c11a3cdbf88b06e8ccda89f78f61505fe94/more_itertools/more.pyi#L150-L152

ad-chaos:
We make sure that `T` can be sliced instead for it be a Sequence

### Issue reference
Fixes #666 

### Changes
Adds a `SupportsSlicing` protocol to check if the type can be sliced instead

Tested against the following code:
```py
from typing import assert_type, Iterator
from array import array
from more_itertools import sliced

assert_type(sliced(str(1), 1), Iterator[str])
assert_type(sliced(list(range(1)), 1), Iterator[list[int]])
assert_type(sliced(tuple(range(1)), 1), Iterator[tuple[int, ...]])
assert_type(sliced(range(1), 1), Iterator[range])
assert_type(sliced((1).to_bytes(), 1), Iterator[bytes])
assert_type(sliced(array("l", [1, 2, 3]), 1), Iterator[array[int]])
sliced({1: 1}, 1)  # error dict[int, int] cannot be assigned to SupportsSlicing[T]
sliced({1}, 1)  # error set[int] cannot be assigned to SupportsSlicing[T]
```


more_itertools/more.pyi
```diff
@@ -148,8 +152,8 @@ def side_effect(
     after: Callable[[], object] | None = ...,
 ) -> Iterator[_T]: ...
 def sliced(
-    seq: Sequence[_T], n: int, strict: bool = ...
-) -> Iterator[Sequence[_T]]: ...
+    seq: SupportsSlicing[_T], n: int, strict: bool = ...
+) -> Iterator[_T]: ...
```
bbayles, line 156: I'm confused as to why this is returning `Iterator[_T]` and not an iterator over things that themselves support slicing?

ad-chaos: That is precisely what this is doing. `sliced` is a generic function that accepts any type that supports slicing and returns an iterator over that type.

bbayles:
I'm not sure why GitHub Actions is not running tests for this PR. However, when I run the checks myself, I get:

```
% stubtest more_itertools.more more_itertools.recipes
git oerror: more_itertools.more.SupportsSlicing is not present at runtime
Stub: at line 42
<TypeInfo more_itertools.more.SupportsSlicing>
Runtime:
MISSING

Found 1 error (checked 2 modules)
```

Could you determine what's causing this and correct it? Thanks.

ad-chaos:
Done. I also noticed that the `_SizedReversible` protocol is not used anywhere in the codebase.

PR #674, opened 2023-02-18
Delay computation of numeric_range len until needed

### LINKED ISSUE #673: Optimization: delay computing len of numeric_range until needed
Instead of always computing `_len` in `__init__`, we can delay the computation until it is used. For simple iteration over the range, we can avoid computing it entirely. For other uses it will add negligible overhead.

eltoder:
Fixes #673

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -2,7 +2,7 @@ import warnings
 
 from collections import Counter, defaultdict, deque, abc
 from collections.abc import Sequence
-from functools import partial, reduce, wraps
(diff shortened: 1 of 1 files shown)


bbayles (changes requested):
Per the [issue discussion](https://github.com/more-itertools/more-itertools/issues/673#issuecomment-1437102925), the changes requested are:
* Remove the `cached_property` implementation
* Wait until 3.7 goes EOL

Many thanks

eltoder:
Will do.

eltoder:
@bbayles Did my part.

PR #680, opened 2023-02-25
Introduce starstarmap

### LINKED ISSUE #679: Consider introducing starstarmap
### Description
`starstarmap` is the lost cousin of `itertools.starmap`, so whereas the relation between `map` and `starmap`  parallels the relation between `function(a,b)` and `function(*c)`, the relationship between `map` and `starstarmap` parallels the relation between `function(a,b)` and `function(**d)`.

And similar to how `starmap` can be implemented as:
```
def starmap(function, iterable):
    for args in iterable:
        yield function(*args)
```
`starstarmap` could be implemented as:
```
def starstarmap(function, iterable):
    for args in iterable:
        yield function(**args)
```

### References

The argument for inclusion would be the fact that it simply seems missing from `itertools`, and its usefulness follows from the same pattern as `starmap`.

### Examples
A REST interface returns a list of JSON objects:
```
[
    {"identifier": 1, "title": "The Cathedral and the Bazaar"},
    {"identifier": 2, "title": "Homesteading the Noosphere"},
   ...
]
```
Each object needs to be processed and transformed via a processing function:
```
def func(identifier, title):
    ...
    return Book(...)
```
Now to convert from the JSON-parsed list of dictionaries to processed books, we can simply do:
```
books = starstarmap(func, dictionaries)
```

Skeen:
### Issue reference
Closes #679

### Changes
Introduces `starstarmap`


DamianB-BitFlipper:
Just to give my 2 cents as a complete outsider, before I read up on the issue, I though that this function mapped and applied the arguments 2 layers deep.

Something akin to iterating `[(1, 2), (3, 4)], [(5, 6, 7), (8,)]` as `fn(1, 2, 3, 4), fn(5, 6, 7, 8)`, if that makes sense.

After reading the issue, it makes total sense.

It seems that Python nomenclature calls the `**` a "double star". I do not know if it will make the code more clear if it were `doublestarmap`. I am still iffy on this naming even.

Skeen:
I'm very open to renaming it `doublestarmap`. I have no strong attachment to the `starstarmap` name.

@bbayles What do you think?

more_itertools/more.py
```diff
@@ -4390,3 +4391,23 @@ def gray_product(*iterables):
             o[j] = -o[j]
             f[j] = f[j + 1]
             f[j + 1] = j + 1
+
+
+def starstarmap(function, iterable):
```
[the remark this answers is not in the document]
Skeen: Done!

tests/test_more.py
```diff
@@ -5230,6 +5230,51 @@ def test_vs_product(self):
...
+        with self.assertRaises(TypeError):
+            list(mi.doublestarmap(lambda x: x, iterable))
+
+    def test_mismatch_function_larger(self):
+        iterable = [{'a': 1}, {'a': 2}]
+        with self.assertRaises(TypeError):
+            list(mi.doublestarmap(lambda a, b: a+b, iterable))
+
+    def test_no_mapping(self):
+        iterable = [1,2,3,4]
```
ergoithz, line 5268: pep8 violation: add space after `,`

more_itertools/more.py
```diff
@@ -4392,6 +4393,24 @@ def gray_product(*iterables):
...
+    >>> iterable = [{'a': 1, 'b': 2}, {'a': 3, 'b': 4}]
+    >>> list(doublestarmap(lambda a, b: a+b, iterable))
+    [3, 7]
+
+    Raises TypeError if *function*'s signature does not match the mapping
+    contained in the iterable.
+    Raises TypeError if *iterable* does not contain mappings.
+    """
+    for item in iterable:
+        yield function(**item)
```
ergoithz, line 4414: pep8 violation: use two empty lines between module-level functions

ergoithz (changes requested):
In case you didn't noticed, CI is complaining about your PR containing PEP8 violations, you'll have a higher chance of getting this PR approved after solving them. I annotated where those issues are coming from, hoping it helps.

Issue #684, opened 2023-02-26
[Feature Request] nth_combination_with_replacement


lukemarris:
### Description

Similar to nth_combination and nth_permutation, which are already implemented in the library.

It would have signature: `nth_combination_with_replacement(iterable, r, index)`
and would return `list(combinations_with_replacement(iterable, r))[index]`
without enumerating all combinations before.

lukemarris:
Here is my attempt:

```python
def n_choose_r(n, r):
  c = 1
  k = min(r, n - r)
  for i in range(1, k + 1):
    c = c * (n - k + i) // i
  return c


def nth_combination_with_replacement(iterable, r, index):
  """Equivalent to ``list(combinations_with_replacement(iterable, r))[index]``.
  """
  pool = tuple(iterable)
  n = len(pool)
  if (r < 0) or (r > n):
    raise ValueError

  c = n_choose_r(n + r - 1, r)
  if index < 0:
    index += c

  if (index < 0) or (index >= c):
    raise IndexError

  result = []
  i = 0
  while r:
    r -= 1
    while n >= 0:
      num_combs = n_choose_r(n + r - 1, r)
      if index < num_combs:
        break
      n -= 1
      i += 1
      index -= num_combs
    result.append(pool[i])

  return tuple(result)
```

bbayles:
Yes, thanks - I'll accept a PR for this.

lukemarris:
I will not have the bandwidth to make a pull request for this for the foreseeable future.

I am happy for someone else to cannibalize my code above into a full PR.

PR #689, opened 2023-03-02
Indexing of combinations with replacement

### LINKED ISSUE #685: [Feature Request] inverse_nth_*
### Description

The inverse of `nth_combination`, `nth_permutation`, etc ... functions.

For example, it would have a signature `inverse_nth_combination(iterable, r, comb)`
and would return `list(combinations(iterable, r)).index(sorted(comb))`
without enumerating all combinations.

Schoyen:
### Issue reference
Related to #684 and #685, and closes #685.

### Changes
This pull request implements the function `combination_with_replacement_index` as an analog to `combination_index`, but for `itertools.combinations_with_replacement` instead of `itertools.combinations`. The pull request includes a test-class similar to the test-class for `combination_index`.

The algorithm used (and the non-standard naming of the `occupation`-list in the function) is from equation 20 in [https://link.aps.org/doi/10.1103/PhysRevA.81.022124](https://link.aps.org/doi/10.1103/PhysRevA.81.022124). I am not sure if this is the most efficient algorithm, but it works.

I have tried to adhere to the style used in this project, but let me know if I missed my mark.


more_itertools/more.py
```diff
@@ -3956,6 +3957,60 @@ def combination_index(element, iterable):
...
+    computes the index of the first *element*, without computing the previous
+    combinations with replacement.
+
+        >>> combination_with_replacement_index('adf', 'abcdefg')
+        20
+
+    ``ValueError`` will be raised if the given *element* isn't one of the
+    combinations with replacement of *iterable*.
+    """
+    l = ilen(element)
```
bbayles, line 3974: Could you explain why you're using `ilen` here? If the element is an iterator, then this will exhaust it. Would `len` make more sense?

more_itertools/more.py
```diff
@@ -3956,6 +3957,60 @@ def combination_index(element, iterable):
...
+    element = enumerate(element)
+
+    k, y = next(element, (None, None))
+    if k is None:
+        return 0
```
bbayles, line 3979: Is the idea here that if `element` is empty, then the index of the combination is 0? If that's not the right interpretation, could you add a corrected one as a comment?

more_itertools/more.py
```diff
@@ -3956,6 +3957,60 @@ def combination_index(element, iterable):
...
+            else:
+                k = tmp
+        if y is None:
+            break
+    else:
+        raise ValueError(
+            'element is not a combination with replacment of iterable'
+        )
+
+    n = ilen(iterable)
```
bbayles, line 3998: Same question as above: we've already looped over `iterable`, so is this length expected to be valid?

bbayles (changes requested):
Changes requested: There are some comments with questions, and GitHub Actions has some automated checks that are failing.

more_itertools/more.py
Schoyen: This was a copy of the function `combination_index` as seen here: https://github.com/more-itertools/more-itertools/blob/89d58c2ec089f3df54ed6cc9279949b8e2bbcd29/more_itertools/more.py#L3930-L3932
I did not put much thought as to why this is done, but I think an argument might be that both `combinations` and `combinations_with_replacement` from itertools return an empty tuple when the length of the subsequence is zero. Hence the indexing functions should return zero as that is the only element in the combinations. An example of this is then:
```python
assert list(combinations("abcd", 0)) == [()]
assert list(combinations_with_replacement("abcd", 0)) == [()]
assert list(combinations("abcd", 0))[combination_index("", "abcd")] == ()
assert list(combinations_with_replacement("abcd", 0))[combination_with_replacement_index("", "abcd")] == ()
```

more_itertools/more.py
Schoyen: This should have been a normal `len`. I have changed it by converting `element` to a `tuple` and then fetching the length using `len`.

Schoyen:
Thanks for the review! I hope I have addressed all the changes you requested.

bbayles (approved):

PR #690, opened 2023-03-02
Update recipes.iter_index to match CPython PR 102360

bbayles:
This PR updates `more_itertools.recipes.iter_index` to match https://github.com/python/cpython/pull/102360.

We still support Python 3.7, so we're not using the Walrus operator.

### DIFF
--- a/more_itertools/recipes.py
+++ b/more_itertools/recipes.py
@@ -830,9 +830,13 @@ def iter_index(iterable, value, start=0):
     except AttributeError:
         # Slow path for general iterables
         it = islice(iterable, start, None)
-        for i, element in enumerate(it, start):
-            if element is value or element == value:
+        i = start - 1
+        try:
+            while True:
+                i = i + operator.indexOf(it, value) + 1
                 yield i
+        except ValueError:
+            pass
     else:
         # Fast path for sequences
         i = start - 1


vbrozik:
The function `iter_index(iterable, value, start=0)` will silently suppress `ValueError` exceptions raised by a generator passed to the argument `iterable`.

bbayles:
Is this also a problem with the version in the CPython docs?

vbrozik:
Yes, thanks, it is the same. Just to be sure I prepared a demonstration of the bug here:

https://github.com/vbrozik/python-ntb/blob/main/problems_from_forums/2023-07-24_iter_index.ipynb

So I should probably create an issue for the CPython docs, right?

bbayles:
Yes, probably a good idea!

rhettinger:
@bbayles It's going to take a while for the CPython docs issue to resolve because the best solution involves adding a new exception type to distinguish between the two sources of `ValueError`. Without that change, `operator.indexOf` cannot be reliably used with iterators or exotic sequences.

For `more-itertools`, the safest play is to revert this PR which was only an optimization. 

Alternatively, the optimization could be preserved by inspecting the exception to distinguish between the two cases:

```
try:
    operator.indexOf('abc', 'd')
except ValueError as e:
    _index_of_value_error_repr = repr(e)
    
def iter_index(iterable, value, start=0):
    "Return indices where a value occurs in a sequence or iterable."
    # iter_index('AABCADEAF', 'A') --> 0 1 4 7
    try:
        seq_index = iterable.index
    except AttributeError:
        # Slow path for general iterables
        it = islice(iterable, start, None)
        i = start - 1
        try:
            while True:
                yield (i := i + operator.indexOf(it, value) + 1)
        except ValueError as e:
            if repr(e) != _index_of_value_error_repr:  # <==== new
                raise                                  # <==== new
    else:
        # Fast path for sequences
        i = start - 1
        try:
            while True:
                yield (i := seq_index(value, i+1))
        except ValueError:
            pass
```

vbrozik:
@bbayles you are right. In the fast path the problem was already there although it is less common to define own sequence class raising a `ValueError`.

Here Stefan @pochmann replicated the problem: https://github.com/python/cpython/issues/107208#issuecomment-1648708461

bbayles:
Thanks - I subscribed to that thread.

rhettinger:
I'm not concerned about ValueError capture for the sequence path.  The technique of looping over `seq.index()` calls and catching `ValueError` is as old as the hills and is the only reason that `index` was ever given a `start` argument.  The technique doesn't seem to have ever been a problem in practice and is sure not worth hobbling the recipe.  That would just make everyone pay for a problem that no one actually has.

PR #691, opened 2023-03-04
add zip_longest_fill_last

GeorgeFischhof:
### Changes
adds a function, zip_longest_fill_last that implements the itertool proposed in the initial issue in the conversation (smaller result set). It accepts an arbitrary number of iterables and outputs a stream of tuples containing the first item from each iterable. Each subsequent iterable changes every item by exhausting the input iterables in order.
When an iterable is exhausted, the last item will be used, or a value passed to the function for subsequent tuples

The number of tuples in result set is the length of longest iterable


bbayles:
Thanks for the PR... I've got to think about whether I want to do this _and_ `partial_product`, since I think it captures the essence of "trying to get something from each iterable when it's too expensive to do everything." I'm fine with one or the other, but both seems a little much.

I'll do some looking into my own code and see which approach makes more sense.

GeorgeFischhof:
@bbayles I think both make sense; in testing there are several circumstances, several constraints, here the two constraints are the expensive tests and whether several values can be changed in one time.
Also I think, the more iterators more-itertools have, the more chance that users can find the appropriate one for their case

bbayles:
I decided that between this and `partial_product`, `partial_product` is the one I want to include. Thanks for the contribution, though - I appreciate it!

Issue #698, opened 2023-03-17
Optimization: _chunked_even_finite()


elliotwutingfeng:
**_chunked_even_finite()** currently builds up each chunk from an `iterable` by filling up a `buffer` using `append()` and yielding the buffer when it reaches a `full_size` or `partial_size`.

A faster implementation is to yield the chunk as a slice of the `iterable`, without using a for-loop to build up the chunk.

PR #699, opened 2023-03-17
Optimize _chunked_even_finite()

elliotwutingfeng:
### Issue reference

#698

### Changes

- Optimize **_chunked_even_finite** by yielding the chunk as a slice of the `iterable`, without using a for-loop to build up the chunk.

### Implementation Details

`yield list(islice(...))` was used to ensure that chunks for non-list iterables such as strings are yielded as lists, and to handle non-sliceable iterables like sets and dicts.

I used the following in IPython as a rudimentary benchmark.

**OS:** Linux x64 **CPU:** AMD Ryzen 7 5800X

```python
%timeit _ = [list(chunked_even(range(x), y)) for x in range(500) for y in range(1,500)]
# Before: 2.9 s ± 24.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
# After: 1.39 s ± 12.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%timeit _ = [list(chunked_even([z for z in range (x)], y)) for x in range(500) for y in range(1,500)]
# Before: 3.61 s ± 11.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
# After: 1.91 s ± 19.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%timeit _ = [list(chunked_even({z:z for z in range(x)}, y)) for x in range(500) for y in range(1,500)]
# Before: 4.33 s ± 34.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
# After: 2.84 s ± 24.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%timeit _ = [list(chunked_even(set(z for z in range(x)), y)) for x in range(500) for y in range(1,500)]
# Before: 4.31 s ± 26.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
# After: 2.72 s ± 7.32 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -4430,4 +4430,3 @@ def partial_product(*args):
         previous.append(current)
         current = future[0] if len(future) > 0 else None
         future = future[1:]
-
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -4069,7 +4069,7 @@ def _chunked_even_finite(iterable, N, n):
     partial_start_idx = num_full * full_size
     if full_size > 0:
         for i in range(0, partial_start_idx, full_size):
-            yield list(iterable[i : i + full_size])
+            yield list(islice(iterable,i , i + full_size))
 
     # Yield num_partial lists of partial_size
     if partial_size > 0:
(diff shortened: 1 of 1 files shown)


bbayles:
This looks good - thanks.

PR #702, opened 2023-03-23
Micro-op for interleave_evenly

### LINKED ISSUE #701: Micro-op for interleave_evenly
**interleave_evenly()** uses a `list` + `zip` comprehension for applying `operator.sub` to 2 iterables. A small speed gain can be achieved by using `map` instead. The same strategy is already being used in **recipes.py**.

elliotwutingfeng:
### Issue reference

Closes #701

### Changes

- Micro-optimization for **interleave_evenly**

### Implementation Details

Use `map` instead of `zip` as `map` is slightly faster for applying `operator.sub`.

Benchmarked as follows

**OS:** Linux x64 **CPU:** AMD Ryzen 7 5800X

```python
from more_itertools import interleave_evenly
%timeit list(interleave_evenly(([x for x in range(26)] * 100, [chr(ord('a') + x) for x in range(26)] * 100)))

Before: 1.1 ms ± 12 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
After : 895 µs ± 13.5 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -1133,7 +1133,7 @@ def interleave_evenly(iterables, lengths=None):
         yield next(iter_primary)
         to_yield -= 1
         # update errors for each secondary iterable
-        errors = [e - delta for e, delta in zip(errors, deltas_secondary)]
+        errors = list(map(sub, errors, deltas_secondary))
 
         # those iterables for which the error is negative are yielded
         # ("diagonal step" in Bresenham)


bbayles:
Usually this sort of thing involves a performance hit on PyPy. Did you test that too?

elliotwutingfeng:
Did a couple of benchmarks. It's an improvement on PyPy. On a Raspberry Pi it was a little slower.

**OS:** Linux x64 **CPU:** AMD Ryzen 7 5800X **Python Version:** PyPy 3.9.16

```python
Before: 362 µs ± 3.37 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
After : 317 µs ± 1.82 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

**OS:** Raspberry Pi OS 5.15.61 **CPU:** Raspberry Pi 3 Model B **Python Version:** Python 3.9.2
 
```python
Before: 49.7 ms ± 2.82 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
After : 51 ms ± 134 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

bbayles:
I'll probably pass on this one, given that the effect is ambiguous. Thanks for the PR, though!

Issue #677, opened 2023-02-20
Improve `partition`


pochmann:
I propose a new implementation. Benchmark results with the documentation's example predicate `is_odd` on the example's `range(10)` and on `range(10000)`:
```
range(0, 10)
  4.98 μs ± 0.01 μs  official Python recipe
  5.24 μs ± 0.01 μs  new proposal
  6.54 μs ± 0.04 μs  current more-itertools

range(0, 10000)
  1.89 ms ± 0.01 ms  new proposal
  2.41 ms ± 0.01 ms  official Python recipe
  3.17 ms ± 0.01 ms  current more-itertools

Python version:
3.10.8 (main, Oct 11 2022, 11:35:05) [GCC 11.3.0]
```
<details><summary>Benchmark script</summary>

```python
from more_itertools import partition, consume
import operator
from itertools import compress, tee, filterfalse
from timeit import timeit
from statistics import mean, stdev
import sys

def partition_new(pred, iterable):
    if pred is None:
        pred = bool
    t1, t2, p = tee(iterable, 3)
    p1, p2 = tee(map(pred, p))
    return (
        compress(t1, map(operator.not_, p1)),
        compress(t2, p2)
    )

def partition_recipe(pred, iterable):
    "Use a predicate to partition entries into false entries and true entries"
    # partition(is_odd, range(10)) --> 0 2 4 6 8   and  1 3 5 7 9
    t1, t2 = tee(iterable)
    return filterfalse(pred, t1), filter(pred, t2)

funcs = partition, partition_new, partition_recipe

funcs = {
    partition: 'current more-itertools',
    partition_recipe: 'official Python recipe',
    partition_new: 'new proposal',
}

def test(args, number, unit, scale):
    print(args[1])

    # Correctness
    expect = None
    for f in funcs:
        result = list(map(list, f(*args)))
        expect = expect or result
        assert result == expect

    # Speed
    times = {f: [] for f in funcs}
    def stats(f):
        ts = [t*scale for t in sorted(times[f])[:5]]
        return f'{mean(ts):6.2f} {unit} ± {stdev(ts):4.2f} {unit} '
    for _ in range(25):
        for f in funcs:
            t = timeit(lambda: consume(map(consume, f(*args))), number=number) / number
            times[f].append(t)
    for f in sorted(funcs, key=stats):
        print(stats(f), funcs[f])
    print()

is_odd = lambda x: x % 2 != 0
iterable = range(10)
[...]

PR #710, opened 2023-05-15
Issue #677: Improve `partition`

pochmann:
### Issue reference

more-itertools/more-itertools#677

### Changes

Improve `partition`, using `compress` instead of generator expressions.


bbayles:
Let's let `black` do its thing. I don't agree with its every decision, but it's a defensible standard.

PR #720, opened 2023-05-21
Fix unique_in_window to match described behavior

### LINKED ISSUE #719: `unique_in_window` is unclear or wrong
It's documented as:

> Yield the items from iterable that haven’t been seen recently. n is the size of the lookback window.

The example is:

```pycon
>>> iterable = [0, 1, 0, 2, 3, 0]
>>> n = 3
>>> list(unique_in_window(iterable, n))
[0, 1, 2, 3, 0]
```

It's unclear what "seen recently" means. Given the words "window" and "seen", one might think of it as a window of the last n seen input elements:

```
[0, 1, 0, 2, 3, 0]

[0]  => yield 0
[0, 1]  => yield 1
[0, 1, 0]  => don't yield 0, as another is in the window
   [1, 0, 2]  => yield 2
      [0, 2, 3]  => yield 3
         [2, 3, 0]  => yield 0
```

But that's not the case. It's really rather like this:

```
[0, 1, 0, 2, 3, 0]

[0]  => yield 0
[0, 1]  => yield 1
[0, 1, 0]  => don't yield 0, as another is in the window
[0, 1,    2]  => yield 2
   [1,    2, 3]  => yield 3
         [2, 3, 0]  => yield 0
```

Another example:

```pycon
>>> iterable = [0, 1, 1, 1, 1, 1, 1, 1, 1, 0, 2, 3, 4, 2]
>>> n = 2
>>> list(unique_in_window(iterable, n))
[0, 1, 2, 3, 4, 2]
```
The second 0 wasn't yielded, even though 0 hadn't been seen in the last eight elements. But the second 2 was yielded even though the first one was seen just three elements earlier.

If that's intentional, then maybe replace

> Yield the items from iterable that haven’t been seen recently.

with

> Yield the items from iterable that haven’t been yielded recently.

and replace the example with mine (since that rules out the other interpretation, unlike the current example).

elliotwutingfeng:
### Issue reference

Closes #719

### Changes

- Fix **unique_in_window()** to match behavior as described in its documentation.

### Additional

If window size is 1, unique_in_window is effectively a simple iterator unless the the **key** function is not deterministic.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -4273,21 +4273,21 @@ def unique_in_window(iterable, n, key=None):
         raise ValueError('n must be greater than 0')
 
     window = deque(maxlen=n)
-    uniques = defaultdict(int)
+    counts = defaultdict(int)
     use_key = key is not None
 
     for item in iterable:
         if len(window) == n:
             to_discard = window[0]
-            if uniques[to_discard] == 1:
-                del uniques[to_discard]
+            if counts[to_discard] == 1:
+                del counts[to_discard]
             else:
-                uniques[to_discard] -= 1
+                counts[to_discard] -= 1
 
         k = key(item) if use_key else item
-        if k not in uniques:
+        if k not in counts:
             yield item
-        uniques[k] += 1
+        counts[k] += 1
         window.append(k)
 
 
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -4278,7 +4278,7 @@ def unique_in_window(iterable, n, key=None):
 
     for item in iterable:
         if len(window) == n:
-            to_discard = window[0][1]
+            to_discard = window[0]
             if uniques[to_discard] == 1:
                 del uniques[to_discard]
             else:
@@ -4288,7 +4288,7 @@ def unique_in_window(iterable, n, key=None):
         if k not in uniques:
             yield item
         uniques[k] += 1
-        window.append((item, k))
+        window.append(k)
 
 
 def duplicates_everseen(iterable, key=None):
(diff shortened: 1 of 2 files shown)


more_itertools/more.py
```diff
@@ -4273,21 +4273,22 @@ def unique_in_window(iterable, n, key=None):
...
+                del uniques[to_discard]
+            else:
+                uniques[to_discard] -= 1
 
-        yield item
+        k = key(item) if use_key else item
+        if k not in uniques:
+            yield item
+        uniques[k] += 1
+        window.append((item, k))
```
pochmann, line 4292: You store `(item, k)` pairs but never use `item` again, so better only store `k`.

more_itertools/more.py
```diff
@@ -4273,21 +4273,22 @@ def unique_in_window(iterable, n, key=None):
         raise ValueError('n must be greater than 0')
 
     window = deque(maxlen=n)
-    uniques = set()
+    uniques = defaultdict(int)
```
pochmann, line 4276: I'd rename `uniques` to `counts`.

more_itertools/more.py
elliotwutingfeng: Thanks for spotting!

pochmann:
Btw thumbs up for this. I had written a few ways and wasn't happy with any of them. I was going to submit one pretty much the same as this (yours is probably a bit faster). It feels like a lot of code, but at least it's straightforward and fixes the issue.

bbayles (approved):

bbayles:
Many thanks for this PR

Issue #721, opened 2023-05-28
Extension of `iterate`: any arity


nanouasyn:
Let the `iterate` function use `func` of any arity, not only unary `func`.

If a unary `func`, using the initial element `x`, generates a sequence (`x`, `func(x)`, `func(func(x))`, ...), then a binary `func`, using two initial elements `x` and `y`, should generate a sequence (`x1`, `x2`,` func(x1, x2)`, `func(x2, func(x1, x2))`, `func(func(x1, x2)`, `func(x2, func(x1, x2)))`, ...). If we generalize it, an `N`-ary function, using the initial elements `x1`, `x2`, ..., `xN`, should generate a sequence (`x1`, ..., `xN`, `func(x1, ..., xN)`, `func(x2, ..., xN, func(x1, ..., xN))`, `func(x3, ..., xN, func(x1, ..., xN), func(x2, ..., xN, func(x1, ..., xN)))` ...).

So, `iterate` function should accept a variable number of initial elements. Their number will determine how many previous elements of the sequence the `func` function should receive in order to generate the next element of the sequence. To apply an `N`-ary `func`, you only need to pass this function and the first `N` elements of sequence. So, `iterate(func, x)` will use unary `func`, `iterate(func, x1, x2)` will use binary `func`, and so on. Similarly, the number of arguments affects the arity in the `map` function.

If we can use binary functions, we can, for example, express the Fibonacci numbers in one line:

```python
>>> take(10, iterate(lambda a, b: a + b, 0, 1))
[0, 1, 1, 2, 3, 5, 8, 13, 21, 34]
```

Or triangular numbers, if we use a ternary function and three initial elements:

```python
>>> take(10, iterate(lambda a, b, c: 3*c-3*b+a, 0, 1, 3))
[0, 1, 3, 6, 10, 15, 21, 28, 36, 45]
```

A large number of useful sequences are defined in a similar recursive way, when the next element of the sequence is determined by some fixed number of previous ones. Therefore, such a change will significantly expand the use cases of `iterate`.

My suggestion is something like this:

```python
def iterate(func, *initials):
	yield from initials
    queue = deque(initials, maxlen=len(initials))
    while True:
        queue.append(func(*queue))
        yield queue[-1]
```

I think we should also handle 0-ary functions for consistency. `iterate(func)` will be an equivalent of `repeatfunc(func)`.

[...]

PR #722, opened 2023-05-28
Generalize `iterate` for arbitrary arity (#721)

nanouasyn:
### Issue reference
#721

### Changes

`iterate` function now accepts arbitrary number of initial elements and a function of the corresponding arity


bbayles:
Declined per discussion in https://github.com/more-itertools/more-itertools/issues/721

Issue #733, opened 2023-07-16
Have matmul() rely on private versions of batched() and transpose()


rhettinger:
The more-itertools version of `batched()` will start emitting deprecation warnings as soon as Python 3.12 goes final, so the `matmul()` recipe should use its own private version instead.  For 3.12, it can use the real `itertools.batched()` which is much faster, has improved error checking, and has fixed the API to yield tuples instead of lists (like `enumerate`, `zip`, etc.).

Since the `matmul()` recipe is somewhat new, it is not too late to have it return `Iterator[tuple]` instead of `Iterator[list]`.  Better to that switch now than to wait for 3.9 to hit end-of-life by which time `matmul()` will already have an established user base.  If lists are returned, it is almost certain that people will rely on it and mutate their matrices in place.

Also, in Python versions where `zip()` supports `strict`, the `matmul()` recipe should use it.   Making sure the matrix dimensions are compatible is something we really want — silently dropping data is hazardous.

Here is some suggested code:

```
try:

    from itertools import batched as _batched

except ImportError:
    
    def _batched(iterable, n):
        if n < 1:
            raise ValueError('n must be at least one')
        it = iter(iterable)
        while True:
            batch = tuple(islice(it, n))
            if not batch:
                break
            yield batch


try:

    zip(strict=True)

except TypeError:

    _transpose = transpose

else:

    def _transpose(it):
        "Swap the rows and columns of the input."
        # transpose([(1, 2, 3), (11, 22, 33)]) --> (1, 11) (2, 22) (3, 33)
        return zip(*it, strict=True)


def matmul(m1, m2):
    """Multiply two matrices.
    >>> list(matmul([(7, 5), (3, 5)], [(2, 5), (7, 9)]))
    [(49, 80), (41, 60)]

    The caller should ensure that the dimensions of the input
    matrices are compatible with each other.  In Python 3.9
    and later, will raise ValueError for incompatible sizes.
    """
    n = len(m2[0])
    return _batched(starmap(_sumprod, product(m1, _transpose(m2))), n)
```

bbayles:
Thanks! I'll implement this as you suggest.

bbayles:
In PR #734, I'm making some associated changes...

The `DeprecationWarning`s have always come back to annoy me; I've changed course and made `more_itertools.recipes.batched` an alias for `itertools.batched` for newer versions of Python. I did something similar for `pairwise` and evidently forgot about it.

[...]

Issue #732, opened 2023-07-16
Improve math.isqrt() emulation


rhettinger:
Currently, `math.isqrt()` is emulated with `lambda x: int(math.sqrt(x))`.

First, the `int()` should be `floor()` to match the intended semantics.

Second, the revised approach only works for up to 53 bits and then it gives wrong answers:

```
>>> n = 9007199326062755
>>> isqrt(n) == floor(sqrt(n))
False
```

Third, it should raise a `TypeError` for a floating point input:

```
>>> n = 3.14159
>>> floor(sqrt(n))
1
>>> isqrt(n)
Traceback (most recent call last):
...
TypeError: 'float' object cannot be interpreted as an integer
```

Fourth, the `sqrt()` overflows if the input is larger than the range of floats:

```
>>> n = 2 << 1030
>>> r = isqrt(n)
>>> floor(sqrt(n))
Traceback (most recent call last):
...
OverflowError: int too large to convert to float
```

Here is some suggested code:

```
# math.isqrt is available for Python 3.8+
try:

    _isqrt = math.isqrt

except AttributeError:

    def _isqrt(n):
        root = n >> (n.bit_length() // 2)
        if n < 0:
            raise ValueError('isqrt() argument must be nonnegative')
        if n < 0x20000000000000:
            return math.floor(math.sqrt(n))
        while not (root * root <= n < (root + 1) * (root + 1)):
            quotient = n // root
            root = (quotient + root) // 2
        return root

```

Note, the large hex constant `0x20000000000000` is just `sys.float_info.radix ** sys.float_info.mant_dig`.

bbayles:
Thanks for the suggestions; I'll incorporate these.

bbayles:
Since `math.isqrt` is available from Python 3.8 ([ref](https://docs.python.org/3/library/math.html#math.isqrt)), and Python 3.7 is EOL, I decided to drop Python 3.7 support and use `math.isqrt` directly.

PR #734, opened 2023-07-17
Changes for version 10.0.0

bbayles:
This PR updates the library for version 10.0.0.

What's new in this PR:
* Python 3.7 is deprecated
* Issue #732 is fixed by dropping support for Python 3.7, which doesn't have `math.isqrt`
* Issue #733 is fixed by un-deprecating `batched` and making it and `matmul` yield tuples instead of lists

There are several other feature updates; see the changelog for details.


rhettinger:
These edits look like nice improvements.  Thank you.

Also consider using the older version of the `convolve()` recipe.  It is faster than the currently published version which has more educational aims (explaining the algorithm for those who don't know it already and demonstrating how `sliding_window()` can be used) rather than speed.  Besides inlining `sliding_window()` the older version avoids an unnecessary cast to `tuple()` and runs `sumprod()` directly on the deque instance.

```
def convolve(signal, kernel):
    # See:  https://betterexplained.com/articles/intuitive-convolution/
    # convolve(data, [0.25, 0.25, 0.25, 0.25]) --> Moving average (blur)
    # convolve(data, [1, -1]) --> 1st finite difference (1st derivative)
    # convolve(data, [1, -2, 1]) --> 2nd finite difference (2nd derivative)
    kernel = tuple(kernel)[::-1]
    n = len(kernel)
    window = collections.deque([0], maxlen=n) * n
    for x in chain(signal, repeat(0, n-1)):
        window.append(x)
        yield sumprod(kernel, window)
```

bbayles:
Thanks - I've update the `convolve` recipe.

I'll publish version 10.0.0 next week. Given how widely deployed this package is, I want to make sure pushing a major version doesn't cause problems for people with less-than-ideal CI system configurations just before the weekend.

PR #737, opened 2023-07-23
Speed up `zip_broadcast`

Masynchin:
### Issue reference

No Issues 😎 

### Changes

Change logic inside `zip_broadcast`. Now it converts scalars to iterables, so that it can just zip them all together.

Simple benchmark:

- Before:

Result: 37.9 ms

- After:

Result: 3.21 ms


Masynchin:
It is almost exact same as the original implementation (https://github.com/more-itertools/more-itertools/commit/0926631dee33163aa39ff2ce30882ea38be48a1a), but passes new test cases introduced after #561 and #565.

kalekundert:
I don't think this implementation works in the following case:
```
>>> from more_itertools import zip_broadcast
>>> list(zip_broadcast(1, [1,2], strict=True))
[(1, 1), (1, 2)]
```
The problem is that `repeat(1)` will not be considered same size as `[1,2]`, so the `strict=True` check will fail.  It should succeed though, because the scalars are meant to be treated as iterables of the same length as their peers.  See #543 for more discussion of this issue.

Unfortunately, it seems like this case is not covered by the tests.  I feel bad about that; this is definitely a case that I should've included in #565.  I realize now that there are no tests for cases in which `zip_broadcast(..., strict=True)` does anything except raise an exception, so we should definitely add some.

Maybe a way to improve the performance of `zip_broadcast()` would be to make custom wrappers for the scalars and the iterables.  The iterable wrappers would detect when the iterables become exhausted, then notify the scalar wrappers.  The scalar wrappers would behave like `repeat()` initially, but would stop upon receiving the aforementioned notification.  Not totally sure if this would avoid the issues from #561, but it might be worth thinking about.

bbayles:
I'll take a PR with more tests, please!

Masynchin:
@kalekundert I see your optimization in #740. I was thinking about different approach, and came up with this idea (but not any drafts or PoC).

`zip_broadcast` uses `_zip_equal` if strict:

https://github.com/more-itertools/more-itertools/blob/bed081105c314721576ce3cd319860b5d0664181/more_itertools/more.py#L4242

Which uses `_zip_equal_generator`, in case there any "scalar" iterables passed (in this PR version), which in fact `repeat(scalar)`.

https://github.com/more-itertools/more-itertools/blob/bed081105c314721576ce3cd319860b5d0664181/more_itertools/recipes.py#L357-L360

So the issue is that `repeat` never stops and when any "real" iterable is exhausted, `zip_longest` yields combo of scalar and `_marker`. Because of it an `UnequalIterablesError` raised, but we don't want it.

https://github.com/more-itertools/more-itertools/blob/bed081105c314721576ce3cd319860b5d0664181/more_itertools/recipes.py#L335-L340

What if we write another version of `_zip_equal` and `_zip_equal_generator` specifically for `zip_broadcast`? I think about something like this pseudocode:

~~~python
def _zip_equal_generator(n, *iterables):
    for combo in zip_longest(*iterables, fillvalue=_marker):
        match combo.count(_marker):
            case 0: yield combo
            case n: break
            case _: raise UnequalIterablesError()
~~~

Where `n` is a count of "real" iterables. I have added the check for markers count, if it equals to `n`, it would mean that all iterables have stopped at the same length, so it would handle the case you suggested in first reply.

What do you think? (Sorry for bad english)

kalekundert:
That seems like a good idea to me.  I can't think of any reason why it wouldn't work, and it should be just as fast as your original implementation.

Also, just to be clear, I didn't mean for #740 to preclude more substantial optimizations like this.  I just noticed a small inefficiency, and it was easy to fix.

Masynchin:
> I didn't mean for #740 to preclude more substantial optimizations like this. I just noticed a small inefficiency, and it was easy to fix.

Your optimization is absolutely good, didn't mean to put any offense on it. Hope any other ideas would work too!

bbayles:
For both branches, could you please make sure that you've merged in `master` and gotten the new tests?

Masynchin:
> For both branches, could you please make sure that you've merged in `master` and gotten the new tests?

Yep, done!

Masynchin:
@kalekundert I have implemented optimization above, seems that it passes both old tests and new ones you added in #739. I am looking for your help reviewing it.

kalekundert:
I only have a few very minor comments.  Most of them pertain more to `_zip_equal()` than to your PR, but I included them anyways just because they were things I noticed while reading the code.  Feel free to ignore:

- I think it's probably worth avoiding the duplication of `_zip_equal()`, e.g. by giving the existing `_zip_equal()` a keyword-only `zip_equal_generator` argument that defaults to the standard behavior, but can be changed in cases like this.
- Alternatively, it might make sense to not use `_zip_equal()` at all.  Most of the time there will be at least one scalar argument, so `_zip_equal()` will just end up calling `_zip_equal_generator()` anyways.  The extra complexity/code duplication doesn't really seem worth it.  This code path also involves raising and catching an exception, which is a relatively expensive operation, but I don't really think the difference would be noticeable here.
- I'd be tempted to define `_marker` as a local variable, both in your version of `_zip_equal_generator()` and the existing version.  Not only would this completely eliminate the possibility of this value appearing in a user-provided iterable (which is admittedly a paranoid concern), it would also be ≈5% faster for iterables with more than ≈100 elements due to local variable lookups being more efficient than global ones (see below).
- The `for val in combo:` loop in `_zip_equal_generator()` could be replaced with `if _marker in combo:`.  That would be easier to understand, and ≈10% faster (see below).
- The `for`/`else` logic in `_zip_equal()` seems unnecessarily complex.  Why not just raise the exception immediately, instead of breaking and raising it outside the loop?  Of course, this has nothing to do with this PR.

Other than that, everything looks good to me.  I assume this version still runs ≈10x faster than the current implementation?

---

Speed tests for minor optimizations mentioned above:
```py
from itertools import zip_longest
from more_itertools import consume

_marker = object()

args = [
        range(100_000),
        range(100_000),
        range(100_000),
]

def z1(iterables):
    for combo in zip_longest(*iterables, fillvalue=_marker):
        for val in combo:
            if val is _marker:
                raise UnequalIterablesError()
        yield combo

def z2(iterables):
    _marker = object()
    for combo in zip_longest(*iterables, fillvalue=_marker):
        for val in combo:
            if val is _marker:
                raise UnequalIterablesError()
        yield combo

def z3(iterables):
    for combo in zip_longest(*iterables, fillvalue=_marker):
        if _marker in combo:
            raise UnequalIterablesError()
        yield combo

def z4(iterables):
    _marker = object()
    for combo in zip_longest(*iterables, fillvalue=_marker):
        if _marker in combo:
            raise UnequalIterablesError()
        yield combo

import timeit    

for f in [z1, z2, z3, z4]:
    print(
            timeit.timeit(
                stmt=f'consume(f(args))',
                globals=dict(f=f) | globals(),
                number=100,
            )
    )
```
Output:
```
2.199860891967546
2.0907241030363366
2.0204972539795563
1.9232789399684407
```

Masynchin:
> * I think it's probably worth avoiding the duplication of `_zip_equal()`, e.g. by giving the existing `_zip_equal()` a keyword-only `zip_equal_generator` argument that defaults to the standard behavior, but can be changed in cases like this.

I can do something like this:

~~~diff
def zip_broadcast(*objects, scalar_types=(str, bytes), strict=False):
    ...
    if strict:
+        yield from _zip_equal(
+            *iterables, gen=partial(
+                _zip_equal_generator_scalars, iterables_count
+            )
+        )
    else:
        yield from zip(*iterables)


+def _zip_equal(*iterables, gen=_zip_equal_generator):
    ...
    except TypeError:
+        return gen(iterables)


def _zip_equal_generator_scalars(n, iterables):
    ...
~~~

It reverts `_zip_equal` to its previous positional arguments (without needing to provide `n`), with new keyword. Do `partial` part looks good? Would it be better with lambda, or just we should rewrite/choose different approach?

> * This code path also involves raising and catching an exception, which is a relatively expensive operation, but I don't really think the difference would be noticeable here.

It is also raised when regular iterable of infinite/undefined length is occured, so how can this check be eliminated?

kalekundert:
- Yeah, that's exactly what I had in mind.  And I think `partial()` is the right way to go.
- What I meant was something like this:
  ```py
  def zip_broadcast(*objects, scalar_types=(str, bytes), strict=False):
      ...
      if strict:
  +        # It would also make sense to just put the for loop directly here, with no extra function call.
  +        yield from _zip_equal_generator_scalars(iterables, iterables_count)
      else:
          yield from zip(*iterables)
  ```
  Basically, just don't call `_zip_equal()` at all.  The only downside is that you end up using `zip_longest()` instead of `zip()` in the case where all the iterables have lengths that are the same.  But this will not often be the case.

Don't take my suggestions too seriously; I'm not sure they're all good ideas.

Masynchin:
> * I'd be tempted to define `_marker` as a local variable, both in your version of `_zip_equal_generator()` and the existing version.  Not only would this completely eliminate the possibility of this value appearing in a user-provided iterable (which is admittedly a paranoid concern), it would also be ≈5% faster for iterables with more than ≈100 elements due to local variable lookups being more efficient than global ones (see below).
> * The `for val in combo:` loop in `_zip_equal_generator()` could be replaced with `if _marker in combo:`.  That would be easier to understand, and ≈10% faster (see below).
> * The `for`/`else` logic in `_zip_equal()` seems unnecessarily complex.  Why not just raise the exception immediately, instead of breaking and raising it outside the loop?  Of course, this has nothing to do with this PR.

Good spots! Maybe I should't optimize `_zip_equal_generator ` in this PR, so that you can add this optimization and all the others in optimization-specific PR.

Masynchin:
>   Basically, just don't call `_zip_equal()` at all.  The only downside is that you end up using `zip_longest()` instead of `zip()` in the case where all the iterables have lengths that are the same.  But this will not often be the case.
> 
> I'm not sure they're all good ideas.

Feel free to try, I invited you to collaborate on my fork, so that you can push your commits in this PR. I would agree with any of your decision to add/not add any change, such as above.

Masynchin:
@bbayles I have resolved conflicts, is it planned to be merged?

pochmann:
> I think about something like this pseudocode:
> 
> ```python
> def _zip_equal_generator(n, *iterables):
>     for combo in zip_longest(*iterables, fillvalue=_marker):
>         match combo.count(_marker):
>             case 0: yield combo
>             case n: break
>             case _: raise UnequalIterablesError()
> ```

That's not safe. Miscounts if there's a non-marker object that equals the marker. Demo:

```python
from unittest.mock import ANY

_marker = object()
for combo in zip('foo', [1, ANY, 3]):
    print(combo.count(_marker))
```

Output ([Attempt This Online!](https://ato.pxeger.com/run?1=m72soLIkIz9vwYKlpSVpuhY3y9OK8nMVSvMyS0pSi0v0cvOTsxUycwvyi0oUHP0iubjicxOLslOLFGwV8pOyUpNLNDS50vKLFJLzc5PyFTLzFKoyCzTU0_Lz1XUUog11QHp0FIxjNa24FICgoCgzr0QDrFYvOb8UyIYap6kJsR7qCphrAA)):

```
0
1
0
```

Masynchin:
@pochmann if it treats `ANY` as `_marker`, then it also should fail in the current implementation? Can't test this right now

Masynchin:
> @pochmann if it treats `ANY` as `_marker`, then it also should fail in the current implementation? Can't test this right now

Oh, is it because current one uses `is` comparison and this PR uses `count` which uses `__equals__` under the hood? If so, can it be fixed with something like `sum(1 for o in combo if o is _marker)`?

pochmann:
The current implementation doesn't use the marker. Yes, counting with `is` is correct.

Btw I think I also optimized this, but can't find it right now... Maybe I dismissed it because it was more complicated than the current implementation (especially after prefilling the scalars, which I had also done). But it might be simpler than the new suggestion...

pochmann:
> ```
> 2.199860891967546
> 2.0907241030363366
> 2.0204972539795563
> 1.9232789399684407
> ```

What Python version did you use? I think globals got faster in the last few versions. And are those results stable? (I.e., you ran it multiple times and always got very similar results?)

pochmann:
@Masynchin About your current proposal: I'd rather do it like this, without the extra functions:
```python
def zip_broadcast(*objects, scalar_types=(str, bytes), strict=False):
    ...
    iterables = [repeat(obj) if is_scalar(obj) else obj for obj in objects]

    if not strict:
        yield from zip(*iterables)

    if lengths of the non-scalars are all the same:
        yield from zip(*iterables)

    for combo in zip_longest(*iterables, fillvalue=_marker):
        ...
```

Advantages:
- Less code.
- Higher chance of the are-all-lengths-equal check to succeed, as this doesn't include the repeated scalars in that check.
- Faster, as it avoids one generator layer. (One more could also be avoided for all but the last case, by making `zip_broadcast` not be a generator.)

Masynchin:
> That's not safe. Miscounts if there's a non-marker object that equals the marker

Fixed and added as test

Masynchin:
Also, what should I do about flake8 in CI? Same as https://github.com/more-itertools/more-itertools/pull/742#issuecomment-1657233524

bbayles:
Merge in the `master` branch for the `flake8` issue, if you could please.

Masynchin:
I found that in this PR `UnequalIterableError` may be raised with incorrect iterable index. Let me fix that before the merge

kalekundert:
> > ```
> > 2.199860891967546
> > 2.0907241030363366
> > 2.0204972539795563
> > 1.9232789399684407
> > ```
> 
> What Python version did you use? I think globals got faster in the last few versions. And are those results stable? (I.e., you ran it multiple times and always got very similar results?)

I used version 3.10.0, and I ran it a bunch of times.  IIRC, it's not a stable effect when there are only ≈10 items per iterable, but it's very stable by the time you get to ≈1000.  That said, this is a really micro optimization, and I just pointed it out because I noticed it.

Masynchin:
> I found that in this PR `UnequalIterableError` may be raised with incorrect iterable index. Let me fix that before the merge

Fixed

> I'd rather do it like this, without the extra functions

Done

pochmann:
> I found that in this PR `UnequalIterableError` may be raised with incorrect iterable index

Can you tell what that was, give an example? I don't see it, and the [fix commit](https://github.com/more-itertools/more-itertools/pull/737/commits/8836a0ee5eb3984f63fa65baf1ecb4dc13f9e2b1) changes a lot of code.

more_itertools/more.py
```diff
@@ -4228,14 +4229,14 @@ def is_scalar(obj):
     if not objects:
         return
 
-    iterables_count = sum(1 for obj in objects if not is_scalar(obj))
+    iterables = list(filterfalse(is_scalar, objects))
+    iterables_count = ilen(iterables)
```
pochmann, line 4233: It's a list, just use `len`. And I don't think you need the count in the variable. You only use it once or twice, the first time you can just check the list instead, and the second time you can just call `len` there.

Masynchin: > It's a list, just use `len`.

My bad, I had `filterfalse` in mind while adding `ilen`.

> ...and the second time you can just call len there.

~~~python
elif markers == iterables_count:
~~~

Should I use `len` here? It would be more performant if I check the length once and not every step of for-loop

pochmann: > Should I use `len` here?

Yes.

> It would be more performant if I check the length once and not every step of for-loop

No. You get there at most once.

pochmann:
> Before the fix it would be `index 1 has length 1; index 2 has length 2`.

Ah, ok. Not sure which is better, indexes referring to all arguments or just to iterables.

Wasn't it `index 0 has length 1; index 2 has length 2`, though? (The `index 0` is [hardcoded](https://github.com/more-itertools/more-itertools/blob/ff2a90fe44cfe69f6d00663b6f68ef48e909c913/more_itertools/recipes.py#L328).) That would really be wrong.

more_itertools/more.py
Masynchin: > You get there at most once.

Am I missing something?

Left side: code with debug print where length eval would happen, right side: run of proposed code with 10 debug prints

Masynchin:
> Wasn't it `index 0 has length 1; index 2 has length 2`, though?

I just rerun `consume(zip_broadcast(0, [1], [1, 2], strict=True))` on hash before fix and it raises `UnequalIterablesError: Iterables have different lengths` without any length indexes 😬

more_itertools/more.py
pochmann: Your debug message is not true. You only get to the `len` call if the `if` condition is false. Prepend the print to the `elif` condition instead: `elif print(...) or condition:`.

Masynchin: You are right. If `if` statement failed, then for-loop cancels in both `elif` or `else` branches. Thanks for notice!

Masynchin:
I run this benchmark on Python 3.11.3:

~~~python
import timeit
from more_itertools import consume, zip_broadcast

N = 1_000
G = globals()

t1 = timeit.timeit("consume(zip_broadcast(1, 2, [1] * 100_000))", number=N, globals=G)
t2 = timeit.timeit("consume(zip_broadcast(1, 2, [1] * 200_000, [2, 3] * 100_000))", number=N, globals=G)
t3 = timeit.timeit("consume(zip_broadcast(1, 2, [1] * 100_000, strict=True))", number=N, globals=G)

print(t1, t2, t3, sep="\n")
~~~

On master (266ebdcf9027b7bb6ab72f8cd4585804c1e1547e) and this PR (2dd6fe2d03ec66e281facde14a70c7887eeac056), here are the results:

- Master

~~~plain
17.083783166483045
40.894872582517564
21.31543183233589
~~~

- PR

~~~plain
2.304440625011921
5.877900958992541
2.28413129132241
~~~

Almost 10x speed up. There is no unresolved questions, can we merge it?

pochmann:
I think none of those test your slow case, can you test that as well?

bbayles:
Is there still a test to add here?

Masynchin:
> Is there still a test to add here?

After #739 been merged and I tweaked this PR to pass this tests, no other problems was found. The only thing requested is regression benchmarks, but I am currently too busy with other stuff. My thought is that it can only regress if caller provides only one iterable and N scalars. I would be happy if anyone could verify this

bbayles:
I think nobody is eager to finish this off, so closing. Thanks for the contribution nonetheless!

Issue #749, opened 2023-08-05
Other more_itertools functions use Iterable[object] in type annotations


obaltian:
Couple of days ago [we fixed an incorrect type annotation for consume() function](https://github.com/more-itertools/more-itertools/issues/745), but later I noticed that such issue is present also in the following functions:

1. all_equal()
2. dotproduct()
3. iter_index()
4. ilen()
5. bucket()
6. locate()
7. iequals()
8. constrained_batches()

Also, the `_GenFn` [type annotation](https://github.com/more-itertools/more-itertools/blob/266ebdcf9027b7bb6ab72f8cd4585804c1e1547e/more_itertools/more.pyi#L32) should be fixed to make mypy able to understand, that the same types are used in `iterator` and `too_short`/`too_long` functions, for example.

I use more_itertools a lot, and I would be happy to get rid of many cast()/type ignore stuff currently present in my projects. I will attach PR to this issue.

Small script showing the problem:
```python
# bug.py
# more-itertools==10.0.0, mypy==1.4.1 (strict=True)
import itertools

import more_itertools

more_itertools.all_equal(
    itertools.takewhile(
        lambda label: label.startswith("VB"),
        ["VBZ", "VBG", "VBN", "NP", "NN"],
    ),
)

more_itertools.dotproduct(
    itertools.takewhile(
        lambda label: label.startswith("VB"),
        ["VBZ", "VBG", "VBN", "NP", "NN"],
    ),
    [1,2,3,4],
)

more_itertools.iter_index(
    itertools.takewhile(
        lambda label: label.startswith("VB"),
        ["VBZ", "VBG", "VBN", "NP", "NN"],
    ),
    True,
)

more_itertools.ilen(
    itertools.takewhile(
        lambda label: label.startswith("VB"),
        ["VBZ", "VBG", "VBN", "NP", "NN"],
    )
)

bucket: more_itertools.bucket[str, str] = more_itertools.bucket(
    ["VBZ", "VBG", "VBN", "NP", "NN"],
    key=lambda x: x[0],
    validator=lambda x: x.startswith("V"),
)

more_itertools.locate(
    itertools.takewhile(
        lambda label: label.startswith("VB"),
        ["VBZ", "VBG", "VBN", "NP", "NN"],
    ),
    lambda label: label.startswith("VB"),
)

more_itertools.iequals(
    itertools.takewhile(
        lambda label: label.startswith("VB"),
[...]

PR #750, opened 2023-08-05
Fix issue 749

obaltian:
### Issue reference
https://github.com/more-itertools/more-itertools/issues/749

### Changes
Replaces some very strict `Iterable[object]` and related type annotations to fix bugs, described in the linked issue.


bbayles:
[This comment](https://github.com/more-itertools/more-itertools/pull/309#discussion_r331772970) describes the original intention. What do you say to this?

obaltian:
@bbayles could you specify which changes do you ask about? I did not fully understood what do you mean by linking this comment – the original issue is still present. Now, you will get errors with basic itertools functions, passed to more-itertools.

obaltian:
I think it's better to discuss specific changes – it would be better if you could leave comments next to the changed code.

bbayles:
When these type stubs were being written, one commenter asked about making the change you suggest.

The original author said:

image

I'd like to get your take on whether the original author was wrong, and if so, how you would have responded to their explanation.

obaltian:
I mean these [object] annotations may work inside library, but they completely break mypy runs in projects using more-itertools. You should not specify that you accept only `object` if you literally accept any type (because this library is intended to work with iterables of any type).

bbayles:
So do you think that this example is not relevant?

```
a: Callable[[], Any]
b: Callable[[], object]

if a():
    ...
if b():
    ...

a() + 5
b() + 5  #  error: Unsupported operand types for + ("object" and "int")  [operator]
```

Would you be OK with `a() + 5`?

obaltian:
I think this comment was written with intention to cover more-itertools itself. Yes, it's better to limit usage of arguments inside more-itertools, but at the same this limitation applies to the code, which use more-itertools. 

I don't think we should make typing inside library more strict by bringing many false-positives in user code. Example in linked issue speaks for me.

obaltian:
> So do you think that this example is not relevant?
> 
> ```
> a: Callable[[], Any]
> b: Callable[[], object]
> 
> if a():
>     ...
> if b():
>     ...
> 
> a() + 5
> b() + 5  #  error: Unsupported operand types for + ("object" and "int")  [operator]
> ```
> 
> Would you be OK with `a() + 5`?

Without context and in ideal world, no. But it's hard to answer without example – where this code is supposed to be, inside repository or in user code? And at what cost?

Can you propose an alternative solution for the linked issue?

bbayles:
@ilai-deutel, forgive me for tagging you in 4 years later... do you have an objection to this set of changes? I can see both ideas, but I'm not sure which one to endorse.

ilai-deutel:
Thanks for looping me in @bbayles. This PR looks good to me, looks like using TypeVar instead of object or Any is the right call, making the API less strict but still enforcing internal soundness of the library.

obaltian:
@ilai-deutel thanks. This PR still adds usage of Any in few places, and I would like to comment them to prevent any misunderstanding:

1. **iequals** – if we would use `Iterable[_T]` here, it means we expect every iterable to be of single type. It's not completely true – this function may be used to check few iterables with different types inside.

2. dotproduct - return type Any. It's that special case when we can possible get int from multiplying floats and etc. We can replace Any with _U here if it would be better from your point of view, but personally I see no difference since we don't know resulting type here.

3. Change in `GenFn` – here I am not so sure, I just tried to remove the usage of `[object]`, but I remember that adding typevar there introduced false-positive mypy error in the linked example script.

@ilai-deutel, could you kindly point out if these `Any` in this PR should be replaced by something else? Thank you in advance.

ilai-deutel:
1. **itequals**: I think using a TypeVar here would make sense. "We expect every iterable to be of a single type", but that type can be anything, including `object`, which covers all types. mypy seems to be ok with iterables of different inner types:

  ```python
  from typing import Iterable, TypeVar, Generic
  
  T = TypeVar("T")
  
  class A(Generic[T]):
    def __init__(*iterables: Iterable[T]):
      return

  reveal_type(A())  #  Revealed type is "d.A[<nothing>]" 
  reveal_type(A([1]))  #  Revealed type is "d.A[builtins.int]"
  reveal_type(A([1], [1.2]))  # Revealed type is "d.A[builtins.float]"
  reveal_type(A([1], [1.2], 'abc'))  # Revealed type is "d.A[builtins.object]"
   ```

2. For **dotproduct** I agree that `Any` is fine

3. For **_GenFn** I'm not sure if there is any difference between `Any` and `object` in this bound. I think either would be fine

```python
from typing import TypeVar, Callable, Iterator, Any

_GenFn1 = TypeVar('_GenFn1', bound=Callable[..., Iterator[object]])
_GenFn2 = TypeVar('_GenFn2', bound=Callable[..., Iterator[Any]])

def f1(f: _GenFn1) -> _GenFn1:
  return f 

def f2(f: _GenFn2) -> _GenFn2:
  return f 

def g() -> Iterator[int]:
  yield 1

def h() -> int:
 return 1


reveal_type(f1(g))  # Revealed type is "def () -> typing.Iterator[builtins.int]"
reveal_type(f2(g))  # Revealed type is "def () -> typing.Iterator[builtins.int]"
reveal_type(f1(h()))  # mypy error
reveal_type(f2(h())) # mypy error

```

ilai-deutel:
Oh, interesting. Looks like mypy infers `itertools.takewhile(...)` to be of type `itertools.takewhile[object]` in that case.

- Replacing `itertools.takewhile(...)` with `itertools.takewhile[str](...)` seems to work, but it's extra work on the caller side
- I guess using `Any` is ok in that case if we want to avoid the caller to explicitly specialize or  cast

obaltian:
@bbayles seems like we agreed these changes, what do you think of this?

bbayles:
Thanks for working through the cases. If there are no more revisions, I will merge and put this in the next release.

PR #753, opened 2023-09-03
Test latest Python 3.12 prerelease

hugovk:
### Issue reference

more-itertools/more-itertools#XXXX

Does this need an issue?

### Changes

Instead of pinning to 3.12.0 beta 4, test against the latest available. Right now that's RC1, with RC2 due tomorrow or soon after. And when the full release comes out next month, it'll test that. Also add 3.12 Trove classifier and remove some redundant conditionals.


bbayles (approved):
💯

Issue #754, opened 2023-09-04
Optimize/simplify `first`


pochmann:
Current implementation:
```python
def first(iterable, default=_marker):
    try:
        return next(iter(iterable))
    except StopIteration as e:
        (empty case handling here)
```

Proposal:
```python
def first(iterable, default=_marker):
    for item in iterable:
        return item
    (empty case handling here)
```

That's faster, especially for empty iterables. Here's a benchmark with various arguments (columns: time of current implementation in ns, time of proposal, time difference, arguments):

```python
137 118 -19 (0,),
140 121 -19 [0],
150 122 -28 "0",
200 174 -26 {0},
142 123 -19 {0: 0},
115 92 -23 iter((0,) * 10000),
114 92 -22 iter([0] * 10000),
113 92 -21 repeat(0),
140 118 -22 (x for x in repeat(0)),
220 196 -24 Infinite(),
458 124 -334 (), None
457 126 -331 [], None
454 124 -330 "", None
463 132 -331 set(), None
455 125 -330 {}, None
425 98 -327 iter(()), None
426 98 -328 iter([]), None
422 98 -324 repeat(None, 0), None
429 98 -331 (x for x in ()), None
711 506 -205 Empty(), None
```

<details><summary>Benchmark script</summary>

[...]

PR #755, opened 2023-09-05
Optimize/simplify `first`

pochmann:
### Issue reference

more-itertools/more-itertools#754

### Changes

Optimize/simplify it, see the issue for benchmarks.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -198,15 +198,14 @@ def first(iterable, default=_marker):
     ``next(iter(iterable), default)``.
 
     """
-    try:
-        return next(iter(iterable))
-    except StopIteration as e:
-        if default is _marker:
-            raise ValueError(
-                'first() was called on an empty iterable, and no '
-                'default value was provided.'
-            ) from e
-        return default
+    for item in iterable:
+        return item
+    if default is _marker:
+        raise ValueError(
+            'first() was called on an empty iterable, and no '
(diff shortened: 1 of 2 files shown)


pochmann:
Oh. Hadn't noticed that `from e` before. It was explicitly [added](https://github.com/more-itertools/more-itertools/pull/436#issue-637280353), but only to replace an inaccurate *"During handling of the above exception, another exception occurred:"*. My new implementation doesn't have that issue to begin with, so I'm removing the `from e`.

For the record, `first([])` currently results in:

```
  File "/home/runner/first/.pythonlibs/lib/python3.10/site-packages/more_itertools/more.py", line 202, in first
    return next(iter(iterable))
StopIteration

The above exception was the direct cause of the following exception:

  File "main.py", line 34, in <module>
    first([])
  File "/home/runner/first/.pythonlibs/lib/python3.10/site-packages/more_itertools/more.py", line 205, in first
    raise ValueError(
ValueError: first() was called on an empty iterable, and no default value was provided.
```

In my opinion, that's just distracting undesirable noise about an implementation detail of the function. I doubt the users want that. With my implementation it will simply be:

```
  File "main.py", line 34, in <module>
    first([])
  File "main.py", line 28, in first
    raise ValueError(
ValueError: first() was called on an empty iterable, and no default value was provided.
```

bbayles (approved):

Issue #735, opened 2023-07-18
Trap exceptions in an existing iterable


jaraco:
### Description

Consider the case where one has an iterable but wishes to handle an Exception that occurs during iteration, but to declare that trapping when the iterator is created, such as when invoking `pathlib.Path.iterdir()`. A common example is where a user wishes to trap the FileNotFoundError while listing a directory (taken from https://github.com/python/cpython/issues/78722):

```python
try:
    children = path_that_doesnt_exist.iterdir()
except FileNotFoundError:
    print("directory doesn't exist")
for path in children:
    print(path.stat().st_size)
```

I'd like to be able to write a wrapper for `iterdir()` that will suppress a FileNotFound error and in that condition return an empty iterable.

`iter_except` almost works:

```
>>> list(iter_except(pathlib.Path('does-not-exist').iterdir().__next__, FileNotFoundError))
[]
```

Unless the directory does exist, then the iterator fails when it reaches the end:

```
>>> list(iter_except(pathlib.Path().iterdir().__next__, FileNotFoundError))
Traceback (most recent call last):
  File "/opt/homebrew/Cellar/python@3.11/3.11.4_1/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/more_itertools/recipes.py", line 543, in iter_except
    yield func()
          ^^^^^^
StopIteration

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
RuntimeError: generator raised StopIteration
```

It is possible to additionally trap the `StopIteration`:

```
>>> list(iter_except(pathlib.Path().iterdir().__next__, (FileNotFoundError, StopIteration)))
[...]
```

But now the user is forced to entangle their primary concern (`FileNotFoundError`) with an implementation detail of the exception handler (`StopIteration`). It's also unnatural that the caller must pass `__next__`.

What if instead, `iter_except` would implicitly handle both conditions, such that one could simply call:

```
iter_except(pathlib.Path().iterdir(), FileNotFoundError)
```

Either that or have a new itertool with that interface.

Such a tool would allow users to catch arbitrary exceptions encountered when processing an iterable (or iterator) without having to eagerly consume and store the contents of the iterable (the current best recommendation for callers of iterdir).

jaraco:
An alternate approach could be to supply a tool called `check()`. It would act like `peek`, except it would return just the iterable. Then someone could write:

```
try:
[...]

PR #756, opened 2023-09-05
Issue 735: iter_suppress

bbayles:
Re: https://github.com/more-itertools/more-itertools/issues/735, this PR adds `iter_suppress`.

Per [the discussion](https://github.com/more-itertools/more-itertools/issues/735#issuecomment-1703986485), the new function converts a user-specified exception into something equivalent to `StopIteration`.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -83,6 +83,7 @@ __all__ = [
     'is_sorted',
     'islice_extended',
     'iterate',
+    'iter_suppress',
     'last',
     'locate',
     'longest_common_prefix',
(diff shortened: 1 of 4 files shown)


pochmann:
Why not just this?
```python
def iter_suppress(iterable, *exceptions):
    try:
        yield from iterable
    except exceptions:
        pass
```
Or this?
```python
from contextlib import suppress

def iter_suppress(iterable, *exceptions):
    with suppress(*exceptions):
        yield from iterable
```

Both pass your tests.

more_itertools/more.py
```diff
@@ -4567,3 +4568,37 @@ def outer_product(func, xs, ys, *args, **kwargs):
         starmap(lambda x, y: func(x, y, *args, **kwargs), product(xs, ys)),
         n=len(ys),
     )
+
+
+def iter_suppress(iterable, *exceptions):
```
rhettinger, line 4573: Would this simplification suffice?

```
def iter_suppress(iterable, *exceptions):
    try:
        yield from iterable
    except exceptions:
        return

```

bbayles: 👍

PR #764, opened 2023-10-12
Update versions.rst to clarify 9.0.0 dropping support for previous ve…

howeaj:
This confused me, because it sounded like v9.0.0 only adds support for new versions; it was unclear that it is actually dropping support for older versions.


bbayles:
Hmm, that's what I meant by the `+` sign. How did you interpret that?

m9810223:
Python docs also use `+` sign

PR #767, opened 2023-10-20
Fix formatting of matmul and factor examples in recipes.py

peterbygrave:
### Changes

Fix formatting of examples for matmul and factor in recipes.py

Current rendering:
Screenshot 2023-10-20 at 14 32 30

Thanks for the awesome library!


bbayles (approved):

bbayles:
Many thanks!

PR #769, opened 2023-10-29
docs: add sphinx_rtd_theme to docs/requirements.txt

### LINKED ISSUE #768: Documentation build error: `ModuleNotFoundError: No module named 'sphinx_rtd_theme'`
Hi, it seems that the current docs build is failing because of the missing theme package. Probably this is due to the change in the theme package.

Screenshot 2023-10-29 at 14 44 37

build log: https://readthedocs.org/projects/more-itertools/builds/22295471/

shuuji3:
### Issue reference
fix #768

### Changes
This PR adds explicit dependency to `sphinx_rtd_theme` which is the missing one during the build. Also, I specified `requirements.txt` in `.readthedocs.yaml` following the documentation of Read the Docs[^1] (maybe it's be better to pin the package version as well). That should ensure installing dependencies.

[^1]: [How to create reproducible builds — Read the Docs user documentation](https://docs.readthedocs.io/en/stable/guides/reproducible-builds.html#id6)

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -7,3 +7,7 @@ build:
 
 sphinx:
    configuration: docs/conf.py
+
+python:
+  install:
+    - requirements: docs/requirements.txt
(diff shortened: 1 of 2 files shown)


bbayles:
Many thanks!

PR #776, opened 2023-12-11
Add is_unique_everseen

haukex:
Backstory: I do a lot of checking of data for duplicates. I needed a variation of the `unique_everseen` iterator, so I wrote my own, but I realized that the implementations of `unique_everseen`, `duplicates_everseen`, and my own iterator were practically identical. So I realized that these three could be reduced down to a single implementation, `is_unique_everseen`, which can then be used as a filter.

I propose two things:

1. The addition of the `is_unique_everseen` iterator: "For each element in the input iterable, return either `True` if this element is being seen for the first time, or `False` if this element has been seen in the input before."

       >>> list(is_unique_everseen('mississippi'))
       [True, True, True, False, False, False, False, False, True, False, False]

   An iterable can be easily checked for duplicates via `all(is_unique_everseen(...))`, and filters like [this one](https://github.com/haukex/igbpyutils/blob/58ba2a92/igbpyutils/iter.py#L118) can be easily implemented based on `is_unique_everseen`.

   This is implemented by the first commit in this pull request.

2. Optional: Implement `unique_everseen` and `duplicates_everseen` via `is_unique_everseen`; implemented by the second thru fourth commit in this pull request. Implementing this reduces code duplication, but may cause a slight slowdown of the former two implementations.

   (The second commit in this pull request simply moves `unique_everseen` from `recipes` to `more`, so that all three implementations are right next to each other in the source. If you prefer to not change the `unique_everseen` and `duplicates_everseen` implementations, I might suggest to instead just incorporate the first two commits.)


bbayles:
Thanks for the suggestion. I'm open to adding a new function that yields `True` or `False` based on uniqueness of elements, but I'd rather not change the implementations of `unique_everseen` since it  is used quite a bit and has stood the test of time.

If you would add something new like `classify_unique` that yields `(element, is_unique_justseen, is_unique_everseen)`, then I think that would be useful.

PR #777, opened 2023-12-13
Add classify_unique

haukex:
@bbayles As you suggested over in #776, here's an implementation of `classify_unique` that yields `(element, is_unique_justseen, is_unique_everseen)`. Feel free to close #776 in favor of this one.


bbayles (approved):

bbayles:
Thanks for this - nice work.

PR #781, opened 2023-12-21
Suggestion: Add a local-tests.sh

haukex:
I noticed in several pull requests (including my own) that people sometimes miss some of the checks in the GitHub Actions workflow. So this commit adds `local-tests.sh` which sets up a `.venv` and does the same steps as the "build" action in `.github/workflows/python-app.yml` locally, so people can check their pull requests before submission.


bbayles:
Thanks for the suggestion! I went with [an alternative](https://github.com/more-itertools/more-itertools/pull/785), but your PR here made me see the need.

Issue #774, opened 2023-11-22
Use math.comb and math.perm instead of math.factorial?


pochmann:
There are multiple uses of factorial-combinations that I think could be replaced by comb/perm, which were introduced in Python 3.8.

elliotwutingfeng:
Some rudimentary benchmarks comparing **math.perm** and **math.factorial** on Python 3.11.6, Linux x64, AMD 5800X.

```python
(factorial(10) // factorial(3)) == perm(10, 7)
# True
%timeit perm(10, 7)
# 31.3 ns ± 1.76 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
%timeit factorial(10) // factorial(3)
# 39.4 ns ± 0.0919 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

(factorial(1000) // factorial(42)) == perm(1000, 958)
# True
%timeit perm(1000, 958)
# 24.8 µs ± 142 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
%timeit factorial(1000) // factorial(42)
# 20.6 µs ± 45.9 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

pochmann:
The latter is surprising. I'll have a look.

How about `comb`? That would replace *three* factorials and might be optimized better (I don't remember whether `perm` was optimized similarly).

pochmann:
`perm` and `comb` with those large numbers (and alternatives with `f`=`factorial`):

```
n, r = 1000, 958
nr = n - r

 66.2 ms  perm(n, r)
 58.0 ms  f(n) // f(nr)

  2.3 ms  comb(n, r)
110.3 ms  f(n) // (f(r) * f(nr))
104.9 ms  f(n) // f(r) // f(nr)
109.9 ms  f(n) // f(nr) // f(r)
```
With `r` on the other end:
```
n, r = 1000, 42
nr = n - r

  0.9 ms  perm(n, r)
103.2 ms  f(n) // f(nr)

  1.6 ms  comb(n, r)
103.5 ms  f(n) // (f(r) * f(nr))
110.9 ms  f(n) // f(r) // f(nr)
102.1 ms  f(n) // f(nr) // f(r)
```

elliotwutingfeng:
```python
from math import perm, comb, factorial as f
n, r = 1000, 42
nr = n - r

  0.3 ms  perm(n, r)
[...]

PR #782, opened 2023-12-21
Issue #774: Use math.comb() and math.perm() where possible

rhettinger:
The `math.perm()` and `math.comb()` became available in Python 3.8.   I think more-itertools has dropped support for 3.7 so we don't need a pure python equivalent.


bbayles:
Many thanks!

PR #790, opened 2024-01-07
Skip TimeLimitedTests.test_zero_limit on Windows

haukex:
In `time_limited.__next__`, the condition checked is `monotonic() - self._start_time > self.limit_seconds`, where `self._start_time = monotonic()`. But on Windows, `monotonic()` is too inaccurate, so `monotonic() - self._start_time == 0` and the test fails.

Note that this commit allows all the checks from the `Makefile` to pass on Windows.

### DIFF
--- a/tests/test_more.py
+++ b/tests/test_more.py
@@ -1,4 +1,5 @@
 import warnings
+import sys
 
 from collections import Counter, abc
 from collections.abc import Set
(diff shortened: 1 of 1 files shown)


haukex:
However, thinking about this a bit more, since the [intended behavior](https://github.com/more-itertools/more-itertools/blob/91a38218a99e633768c5cfc3010ef0d246017f2e/tests/test_more.py#L3904) appears to be that `time_limited(0, ...)` always returns nothing, I think the correct fix might actually be to special-case the [behavior](https://github.com/more-itertools/more-itertools/blob/91a38218a99e633768c5cfc3010ef0d246017f2e/more_itertools/more.py#L3233) for `limit_seconds==0`.

PR #792, opened 2024-01-07
100% Code Coverage

haukex:
This PR increases the code coverage to 100% (and ups the required code coverage accordingly), tested on Python 3.8 to 3.12 on Linux. Note this includes a few actual fixes:

- 6d60ae231b9d3eb9a65f808bf59c9b6a7916dcfa: Corrects `NthPermutationTests` (some tests accidentally referred to `nth_combination` instead)
- a1c7e6611a6f75700b7e9f524d994f97ecf3a47c: Removes unreachable code in `nth_permutation`

I chose to simply suppress coverage on a few bits of version-dependent code. As a side note, there is an alternative that I [use](https://github.com/haukex/igbpyutils/blob/58ba2a92a612a06223f045d5dc9d26fdf3686e54/prove.sh#L11) in my `igbpyutils` project: [Generated](https://igbpyutils.readthedocs.io/en/stable/dev.html#igbpyutils.dev.generate_coveragerc) `.coveragerc` files with tags such as `# cover-req-ge3.10` and `# cover-req-lt3.10` for each Python version. This allows me to run Python version-specific coverage checks in that project.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,8 @@ pip-log.txt
 .coverage
 .tox
 .noseids
+htmlcov/
+coverage.xml
 
 # Docs by Sphinx
 _build
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ format:
 .PHONY: coverage
 coverage:
 	coverage run --include="more_itertools/*.py" -m unittest
-	coverage report --show-missing --fail-under=99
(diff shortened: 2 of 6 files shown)


bbayles:
I took the additional test and `.gitignore` changes, but left the `pragma` statements and `fail-under` change. Thanks for this!

PR #793, opened 2024-01-09
Speed up `ichunked`

james-wasson:
### Issue reference

https://github.com/more-itertools/more-itertools/issues/797

### Changes
I was trying to create my own `ichunked` method and found this library and thought I might contribute. This new method implementation runs about 2x faster than the current implementation due to it not using any user created iterators (eg. classes with the `next()` method).

There should be no changes visible to the end user other than the speed difference unless the user is relying on the private class `_IChunk`.

This implementation uses a scoped variables within a wrapping method scope to modify the state of the generator on the fly.

I have tested and timed this method extensively using both calls to `next()` and just looping. I have also tried modifying the state part way through iteration and I have not run across any issues.

A simple timing test:

- Before
image
- After
image

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -3352,13 +3352,16 @@ def ichunked(iterable, n):
         raise ValueError('n must be a positive integer')
     
     while True:
+        # Create new chunk
         (chunk, materialize_next) = _ichunk(iterable, n)
 
+        # Check to see whether we're at the end of the source iterable
         if not materialize_next():
             return
 
         yield chunk
 
+        # Fill previous chunk's cache
         materialize_next(None)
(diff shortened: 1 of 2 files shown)


james-wasson:
All checks passed locally.

Had to modify a test to work around the non exposure of the `_cache` attribute. I believe all the same functionality is tested.

james-wasson:
For timing reference:
```
from more_itertools import ichunked, consume
%timeit consume(consume(c) for c in ichunked([0] * 10000, 107))
```

bbayles:
I took a few shots at making a simpler version of this, but yours is undeniably faster. Thanks!

PR #795, opened 2024-01-09
Minor improvement for `padded`

james-wasson:
### Issue reference

https://github.com/more-itertools/more-itertools/issues/798

### Changes

A small speed improvement for the `padded` method by using `yield from` rather than a loop.

Before:
image

After:
image

### Checks and tests
Not sure how to do this but will do as soon as I can figure it out.


bbayles:
Thanks for the PR. The performance difference here is small enough that I'd rather not change things.

Issue #794, opened 2024-01-09
`make_size` - A tools that will make an iterable exactly size n


james-wasson:
### Description

I propose a new itertool class `make_size` that will make an iterable exactly size n.

It is really more of a convince method as the implementation is almost trivial but is a bit long and can be hard to digest.

### References

I have created this method to solve the problem of no guaranteed chunk sizes in `more_itertools.ichunked`.

I'm sure there are many other use cases such as truncation of a string while padding with spaces at the end.

### Examples

```
>>> list(make_size([1, 2, 3], 5))
[1, 2, 3, None, None]

>>> list(make_size([1, 2, 3], 5, fillvalue=0))
[1, 2, 3, 0, 0]
```

### Proposed Implementation

```python
def make_size(iterable, n, fillvalue=None):
    """Yield the elements from *iterable*, followed by *fillvalue*, such that
    exactly *n* items are emitted.

        >>> list(make_size([1, 2, 3], 5))
        [1, 2, 3, None, None]

        >>> list(make_size([1, 2, 3], 5, fillvalue=0))
        [1, 2, 3, 0, 0]

    *n* must be >= 0
    """

    iterable = iter(iterable)
    if n < 0:
        raise ValueError('n must be >= 0')

    return islice(chain(iterable, repeat(fillvalue)), n)
```

james-wasson:
PR: https://github.com/more-itertools/more-itertools/pull/806

bbayles:
Doesn't `padded` with `next_multiple` solve this?

james-wasson:
@bbayles
`padded([1, 2, 3, 4, 5], n=3, next_multiple=True)` gives `[1, 2, 3, 4, 5, None]` when we would want the size limited to exactly 3. The correct result would be `[1, 2, 3]`.

If there is a way to incorporate this extended functionality into padded I think it would fit nicely in that method but I wasn't able to think of a good solution to do that.

bbayles:
I think I'd rather let people componse `padded` and `islice` themselves - I'll accept a PR that gives an example in the documentation for this, if you'll make one.

james-wasson:
Added exactly size of n doc to padded

james-wasson:
I'm reopening this because I thought maybe we could add a `fillvalue` option to the `take` method to implement this feature.

@bbayles Let me know your thoughts here.

PR #806, opened 2024-01-20
Add new tool `make_size`

james-wasson:
### Issue reference
more-itertools/more-itertools#794

### Changes
Yield the elements from an iterable followed by a fillvalue, such that exactly n items are emitted.


james-wasson:
@bbayles says

> I think I'd rather let people componse padded and islice themselves - I'll accept a PR that gives an example in the documentation for this, if you'll make one.

Closing this PR

PR #807, opened 2024-01-31
Add exactly size of n doc to padded

james-wasson:
### Issue reference
more-itertools/more-itertools#794

### Changes
I have used to native method I discussed in the issue to add to the documentation. If this is not acceptable and we want to only compose using the `padded` method I don't think any docstring PR should be merged related to this issue.

Using the padded method
`islice(padded(iterable, '?', n), n)`
is less performant and less clear than using
`islice(chain(iterable, repeat('?')), n)`
in my opinion.

Thoughts on this are welcome.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -1536,7 +1536,7 @@ def padded(iterable, fillvalue=None, n=None, next_multiple=False):
     If *n* is ``None``, *fillvalue* will be emitted indefinitely.
 
     To create an *iterable* of exactly *n* size you can compose native
-    itertool methods:
+    itertools methods:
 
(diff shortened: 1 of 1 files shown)


james-wasson:
Not sure why this build is failing. Locally there is no issue with `make check-all`. Any help would be appreciated.

elliotwutingfeng:
I am getting the same "black" reformat warnings from a clean slate + removed erroneous quotes from the workflow file (https://github.com/elliotwutingfeng/more-itertools/actions/runs/7739500311).

bbayles:
Resolution: update to the latest versions of `flake8` and `black`.

Issue #809, opened 2024-02-16
Optimize windowed


james-wasson:
## Intro

I have been working towards increasing the performance of the `windowed` method and have made some good headway. Main performance gains come from using built in itertools methods and using islice to "fast forward" the windows.

At low list sizes the performance does take a small hit but I did try to minimize that as much as possible.
As the list size, `n`, or `step` get larger the benefits become clearer gaining approximately a 25-50% performance gain.

### Testing

**Test:**
```python
from more_itertools import windowed, consume
%timeit -n 100000 consume(windowed([1]*10, 3, step=1))
%timeit -n 100000 consume(windowed([1]*10, 3, step=2))
%timeit -n 100000 consume(windowed([1]*10, 3, step=4))
%timeit -n 100000 consume(windowed([1]*10, 3, step=6))
print()
%timeit -n 100000 consume(windowed([1]*20, 5, step=1))
%timeit -n 100000 consume(windowed([1]*20, 5, step=2))
%timeit -n 100000 consume(windowed([1]*20, 5, step=4))
%timeit -n 100000 consume(windowed([1]*20, 5, step=8))
%timeit -n 100000 consume(windowed([1]*20, 5, step=12))
print()
%timeit -n 100000 consume(windowed([1]*40, 10, step=1))
%timeit -n 100000 consume(windowed([1]*40, 10, step=4))
%timeit -n 100000 consume(windowed([1]*40, 10, step=7))
%timeit -n 100000 consume(windowed([1]*40, 10, step=9))
%timeit -n 100000 consume(windowed([1]*40, 10, step=15))
%timeit -n 100000 consume(windowed([1]*40, 10, step=25))
print()
%timeit -n 100000 consume(windowed([1]*80, 20, step=1))
%timeit -n 100000 consume(windowed([1]*80, 20, step=6))
%timeit -n 100000 consume(windowed([1]*80, 20, step=12))
%timeit -n 100000 consume(windowed([1]*80, 20, step=19))
%timeit -n 100000 consume(windowed([1]*80, 20, step=25))
%timeit -n 100000 consume(windowed([1]*80, 20, step=35))
print()
%timeit -n 10000 consume(windowed([1]*1000, 30, step=1))
%timeit -n 10000 consume(windowed([1]*1000, 30, step=2))
%timeit -n 10000 consume(windowed([1]*1000, 30, step=3))
%timeit -n 10000 consume(windowed([1]*1000, 30, step=4))
[...]

PR #810, opened 2024-02-16
Optimize windowed itertool

james-wasson:
### Issue reference
more-itertools/more-itertools#809

### Changes
I have been working towards increasing the performance of the windowed method and have made some good headway. Main performance gains come from using built in itertools methods and using `islice` to "fast forward" the windows.


bbayles:
I'm -0 on this one - the better performance for larger size inputs is good, but I'm not sure if the extra complexity is worth it.

james-wasson:
I can test out the performance of the single non-overlapping branch but I really do think non initializing a deque if possible is the way to go here. I will see if I can optimize further.

james-wasson:
Newest changes improve every metric minorly anywhere from 0-40% speed up over base code. Not huge but considering this is one of the most widely used methods and is used all over this code base as well it should provide a nice bonus.

bbayles:
Thanks for the analysis. I'll take this PR if you can revert the change to the argument names?

james-wasson:
@bbayles Done, I will avoid changing argument names in the future. Thanks!

bbayles (approved):

Issue #811, opened 2024-02-17
Optimize adjacent itertool


james-wasson:
## Intro

Increases the performance of the `adjacent` itertool by using a counting method to check for adjacency.

### Testing

**Test:**
```python
from more_itertools import adjacent, consume
%timeit -n 10000 consume(adjacent(lambda x: x == 2, [1, 2, 3, 4, 5, 6]))
%timeit -n 10000 consume(adjacent(lambda x: x == 2, [1, 2, 3, 4, 5, 6], distance=2))
print()
%timeit -n 10000 consume(adjacent(lambda x: x == 2, [1, 2, 3, 4, 5, 6]*3))
%timeit -n 10000 consume(adjacent(lambda x: x == 2, [1, 2, 3, 4, 5, 6]*3, distance=2))
print()
%timeit -n 10000 consume(adjacent(lambda x: x == 2, [1, 2, 3, 4, 5, 6]*10))
%timeit -n 10000 consume(adjacent(lambda x: x == 2, [1, 2, 3, 4, 5, 6]*10, distance=2))
%timeit -n 1000 consume(adjacent(lambda x: x == 2, [1, 2, 3, 4, 5, 6]*10, distance=5))
print()
%timeit -n 1000 consume(adjacent(lambda x: x == 2, [1, 2, 3, 4, 5, 6]*50))
%timeit -n 1000 consume(adjacent(lambda x: x == 2, [1, 2, 3, 4, 5, 6]*50, distance=2))
%timeit -n 1000 consume(adjacent(lambda x: x == 2, [1, 2, 3, 4, 5, 6]*50, distance=10))
%timeit -n 1000 consume(adjacent(lambda x: x == 2, [1, 2, 3, 4, 5, 6]*50, distance=20))
print()
%timeit -n 1000 consume(adjacent(lambda x: x == 2, [1, 2, 3, 4, 5, 6]*500))
%timeit -n 1000 consume(adjacent(lambda x: x == 2, [1, 2, 3, 4, 5, 6]*500, distance=2))
%timeit -n 1000 consume(adjacent(lambda x: x == 2, [1, 2, 3, 4, 5, 6]*500, distance=10))
%timeit -n 1000 consume(adjacent(lambda x: x == 2, [1, 2, 3, 4, 5, 6]*500, distance=50))
%timeit -n 1000 consume(adjacent(lambda x: x == 2, [1, 2, 3, 4, 5, 6]*500, distance=100))
```

**Before:**
```
[...]

PR #812, opened 2024-02-17
Optimize adjacent itertool

james-wasson:
### Issue reference
more-itertools/more-itertools#811

### Changes
Increases the performance of the adjacent itertool by using a counting method to check for adjacency.

Check the issue for performance overview.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -1955,7 +1955,7 @@ def adjacent(predicate, iterable, distance=1):
     drawn from *iterable* and the `bool` indicates whether
     that item satisfies the *predicate* or is adjacent to an item that does.
 
-    For example, to find whether items are adjacent to a ``3``::
+    For example, to find whether items are adjacent to a ``3``:
 
         >>> list(adjacent(lambda x: x == 3, range(6)))
         [(False, 0), (False, 1), (True, 2), (True, 3), (True, 4), (False, 5)]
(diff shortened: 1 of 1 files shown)


bbayles:
I think I'm -1 on this: the performance does indeed look better, but the code is simpler before.

I would be open to changing `windowed` to `sliding_window`, which I think is a simpler/faster implementation; `adjacent` doesn't need the additional features.

james-wasson:
@bbayles I would assume the counting approach would always be a better option but maybe this is a sub optimal way of implementing it. I will try to improve it and if I can't I will change the PR to use `sliding_window` instead.

james-wasson:
@bbayles I tried just adding the `sliding_window` replacement and for my tests didn't come up with any major results. After trying around a little more I was able to remove the `accumulate`  dependency, increase readability, and add a slight performance increase.

I think this is as far as I can take this function so I would love to hear any feedback.

Thanks!

james-wasson:
@bbayles I have compacted this further and increased readability; I don't think this can go much smaller or faster.
If you're still at a -1 on this PR feel free to close it, no worries :)

james-wasson:
If needed in the future my recommendation is to go with @pochmann approach here: https://github.com/more-itertools/more-itertools/issues/811#issuecomment-2002209995

Closing this PR now. See the issue for resolution.

Issue #813, opened 2024-02-18
Optimize replace itertool


james-wasson:
## Intro

Increases the performance of the `replace` itertool by removing the reliance on `windowed` as well as other minor adjustments.

I have also added the `default_arg` option to specify what default arguments are passed to the `pred` when reaching the end of the iterable. I'm not sure using `_marker` was ever good practice but I have left the default as `_marker` for backwards compatibility. We should probably think about changing this in the future however.

### Testing

**Test:**
```python
from more_itertools import replace, consume

iterable = [1, 1, 0, 1, 1, 0, 1, 1]
pred = lambda x: x == 0
substitutes = (2, 3)

%timeit -n 10000 consume(replace(iterable, pred, substitutes))
%timeit -n 10000 consume(replace(iterable*5, pred, substitutes))
%timeit -n 10000 consume(replace(iterable*10, pred, substitutes))
%timeit -n 1000 consume(replace(iterable*25, pred, substitutes))
%timeit -n 1000 consume(replace(iterable*100, pred, substitutes))
print()
%timeit -n 10000 consume(replace(iterable, pred, substitutes, count=2))
%timeit -n 10000 consume(replace(iterable*5, pred, substitutes, count=2))
%timeit -n 10000 consume(replace(iterable*10, pred, substitutes, count=2))
%timeit -n 1000 consume(replace(iterable*25, pred, substitutes, count=2))
%timeit -n 1000 consume(replace(iterable*100, pred, substitutes, count=2))
print()

iterable = [0, 1, 2, 5, 0, 1, 2, 5]
window_size = 3
pred = lambda *args: args == (0, 1, 2)
substitutes = (2, 3)

%timeit -n 10000 consume(replace(iterable, pred, substitutes, window_size=window_size))
%timeit -n 10000 consume(replace(iterable*5, pred, substitutes, window_size=window_size))
%timeit -n 10000 consume(replace(iterable*10, pred, substitutes, window_size=window_size))
%timeit -n 1000 consume(replace(iterable*25, pred, substitutes, window_size=window_size))
%timeit -n 1000 consume(replace(iterable*100, pred, substitutes, window_size=window_size))
```

**Before:**
```
3.98 µs ± 35.5 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
14.4 µs ± 337 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
[...]

PR #814, opened 2024-02-18
Optimize replace itertool

james-wasson:
### Issue reference

more-itertools/more-itertools#813

### Changes

Increases the performance of the `replace` itertool by removing the reliance on `windowed` as well as other minor adjustments.

Check the issue for performance overview.

I have also added the `default_arg` option to specify what default arguments are passed to the `pred` when reaching the end of the iterable. I'm not sure using `_marker` was ever good practice but I have left the default as `_marker` for backwards compatibility. We should probably think about changing this in the future however.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -3093,7 +3093,9 @@ def rlocate(iterable, pred=bool, window_size=None):
     return reversed(list(locate(iterable, pred, window_size)))
 
 
-def replace(iterable, pred, substitutes, count=None, window_size=1):
+def replace(
+    iterable, pred, substitutes, count=None, window_size=1, default_arg=_marker
+):
     """Yield the items from *iterable*, replacing the items for which *pred*
     returns ``True`` with the items from the iterable *substitutes*.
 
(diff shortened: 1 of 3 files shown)


bbayles (changes requested):
If you'll put the parameters back to the way they were, I'll accept this PR. I'm not sold on the `default_arg` thing, but the rest looks good. Thanks!

more_itertools/more.py
```diff
@@ -3121,37 +3123,48 @@ def replace(iterable, pred, substitutes, count=None, window_size=1):
...
-                continue
-
-        # If there was no match (or we've reached the replacement limit),
-        # yield the first item from the window.
-        if w and (w[0] is not _marker):
-            yield w[0]
+    for _ in filler:
+        # Produce substitues and advance the window if
+        # the predicate is true and we haven't reached our count
+        if pred(*window) and (count is None or n < count):
```
MattCatz, line 3161:
```suggestion
        if (count is None or n < count) and pred(*window):
```
Swapping the evaluation order would speed up cases where `count` replacements have been happened.

more_itertools/more.py
```diff
@@ -3121,37 +3123,48 @@ def replace(iterable, pred, substitutes, count=None, window_size=1):
...
     # Save the substitutes iterable, since it's used more than once
     substitutes = tuple(substitutes)
 
-    # Add padding such that the number of windows matches the length of the
-    # iterable
-    it = chain(iterable, [_marker] * (window_size - 1))
-    windows = windowed(it, window_size)
+    # Create the iterable with padding so we don't run out of elements
+    # to pass to the predicate
+    iterable = chain(iter(iterable), (default_arg,) * (window_size - 1))
```
rhettinger: Try: `repeat(default_arg, (window_size - 1))`.

Issue #815, opened 2024-02-19
Optimize chunked_even itertool


james-wasson:
## Intro

Increases the performance of the `chunked_even` itertool by removing looped slicing and combining online and finite methods.

As `n` gets large running `islice` in a loop becomes unsustainable as it has to chew through more and more elements to get to the appropriate slice.

We could check if the list is finite and copy to a list to perform slicing, and it is slightly faster but if you run something like `chunked_even(range(1, 1000000000), 100)` you will get an out of memory exception which can be avoided by using an iterator. So I decided to forgo that method as I couldn't find a good way to always avoid that out of memory exception when copying.

### Testing

**Test:**
```python
# test online
from more_itertools import chunked_even, consume

print('online n=3')
%timeit -n 10000 consume(chunked_even(iter(range(1, 8)), 3))
%timeit -n 10000 consume(chunked_even(iter(range(1, 20)), 3))
%timeit -n 10000 consume(chunked_even(iter(range(1, 60)), 3))
%timeit -n 10000 consume(chunked_even(iter(range(1, 120)), 3))
print('online')
%timeit -n 10000 consume(chunked_even(iter(range(1, 20)), 10))
%timeit -n 10000 consume(chunked_even(iter(range(1, 60)), 30))
%timeit -n 10000 consume(chunked_even(iter(range(1, 120)), 50))
%timeit -n 10000 consume(chunked_even(iter(range(1, 500)), 50))
%timeit -n 10000 consume(chunked_even(iter(range(1, 1000)), 100))
%timeit -n 1000 consume(chunked_even(iter(range(1, 10000)), 100))
%timeit -n 100 consume(chunked_even(iter(range(1, 100000)), 100))
print()

# test finite list
from more_itertools import chunked_even, consume

print('finite list n=3')
%timeit -n 10000 consume(chunked_even(list(range(1, 8)), 3))
%timeit -n 10000 consume(chunked_even(list(range(1, 20)), 3))
%timeit -n 10000 consume(chunked_even(list(range(1, 60)), 3))
%timeit -n 10000 consume(chunked_even(list(range(1, 120)), 3))
print('finite list')
%timeit -n 10000 consume(chunked_even(list(range(1, 20)), 10))
%timeit -n 10000 consume(chunked_even(list(range(1, 60)), 30))
%timeit -n 10000 consume(chunked_even(list(range(1, 120)), 50))
%timeit -n 10000 consume(chunked_even(list(range(1, 500)), 50))
[...]

PR #816, opened 2024-02-19
Optimize `chunked_even` itertool

james-wasson:
### Issue reference
more-itertools/more-itertools#815

### Changes
Increases the performance of the chunked_even itertool by removing looped slicing and combining online and finite methods.

See the issue for performance overview.


bbayles:
This one is great - even simpler code.

bbayles (approved):

Issue #817, opened 2024-02-19
Optimize unique_in_window itertool


james-wasson:
## Intro

Optimize the performance of the `unique_in_window` itertool by using `collections.OrderedDict` to keep track of the last items we've seen. This increases performance by ~30%.

### Testing

**Test:**
```python
from more_itertools import unique_in_window, consume
%timeit -n 10000 consume(unique_in_window([0, 1, 0, 2, 3, 0], 3))
%timeit -n 10000 consume(unique_in_window([0, 1, 0, 2, 3, 0]*5, 3))
%timeit -n 10000 consume(unique_in_window([0, 1, 0, 2, 3, 0]*10, 3))
%timeit -n 10000 consume(unique_in_window([0, 1, 0, 2, 3, 0]*20, 3))
%timeit -n 10000 consume(unique_in_window([0, 1, 0, 2, 3, 0]*50, 3))
print()
%timeit -n 10000 consume(unique_in_window([0, 1, 0, 2, 3, 0], 1))
%timeit -n 10000 consume(unique_in_window([0, 1, 0, 2, 3, 0]*5, 5))
%timeit -n 10000 consume(unique_in_window([0, 1, 0, 2, 3, 0]*10, 10))
%timeit -n 10000 consume(unique_in_window([0, 1, 0, 2, 3, 0]*20, 20))
%timeit -n 10000 consume(unique_in_window([0, 1, 0, 2, 3, 0]*50, 50))
print()
```

**Before:**
```
2.44 µs ± 41 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
9.34 µs ± 126 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
17.9 µs ± 258 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
35 µs ± 173 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
87.8 µs ± 856 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

2.62 µs ± 14.7 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
8.79 µs ± 20.5 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
13.9 µs ± 335 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
26.2 µs ± 480 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
[...]

PR #818, opened 2024-02-19
Optimize unique_in_window itertool

james-wasson:
### Issue reference
more-itertools/more-itertools#817

### Changes
Optimize the performance of the unique_in_window itertool by using collections.OrderedDict to keep track of the last items we've seen. This increases performance by ~30%.

See the linked issue for performance overview.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -1,6 +1,6 @@
 import warnings
 
-from collections import Counter, defaultdict, deque, abc
+from collections import Counter, OrderedDict, defaultdict, deque, abc
 from collections.abc import Sequence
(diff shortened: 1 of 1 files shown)


bbayles:
Thanks for the PR. I'm -0 on this one; it's more code and an extra import for something not super frequently used.

james-wasson:
I've updated the PR to match the number of lines of code (not comments) or do better than the previous solution.

I think the extra import is unremovable from this solution.

james-wasson:
After considering it I agree with @bbayles. The performance is already excellent as is and adding such a major change could definitely cause issues. More than that I think this solution is less readable/maintainable.

Closing this PR

Issue #832, opened 2024-04-27
distinct_permutations with incomparable items


hgustafsson:
Because `distinct_permutations` uses `sorted` and `<` it is not possible to input incomparable items.

This is unexpected since the documentation claims that `distinct_permutations` is equivalent to `set(permutations(iterable)`.

Example:
```
>>> set(itertools.permutations(["+","+",0,1]))

{('+', '+', 0, 1),
 ('+', '+', 1, 0),
 ('+', 0, '+', 1),
 ('+', 0, 1, '+'),
 ('+', 1, '+', 0),
 ('+', 1, 0, '+'),
 (0, '+', '+', 1),
 (0, '+', 1, '+'),
 (0, 1, '+', '+'),
 (1, '+', '+', 0),
 (1, '+', 0, '+'),
 (1, 0, '+', '+')}

>>> more_itertools.distinct_permutations(["+","+",0,1])

... more_itertools/more.py:742, in distinct_permutations(iterable, r)
--> 742 items = sorted(iterable)

TypeError: '<' not supported between instances of 'int' and 'str'
```

JamesParrott:
The ordering is heavily used by the implementation and perhaps the docs should mention that.

Is the order of the permutations themselves important to anyone?:

```
>>> list(more_itertools.distinct_permutations([-1,-1,0]))
[(-1, -1, 0), (-1, 0, -1), (0, -1, -1)]
>>> list(more_itertools.distinct_permutations([2,2,0]))
[(0, 2, 2), (2, 0, 2), (2, 2, 0)]
```

Otherwise a list of unique unsortable symbols could be formed, and distinct_permutations called on the index in that list, of each symbol in the list containing duplicates.

bbayles:
I'll accept a patch that makes it work for non-comparable items. We can add a note about the order not being guaranteed. Thanks!

JamesParrott:
Awesome.  How does this look to you?

https://github.com/more-itertools/more-itertools/pull/834/files

JamesParrott:
I've added some tests.  There are a few issues so please bear with me while I sort them.

JamesParrott:
The tests are all passing now :)

pochmann3:
`distinct_permutations([1, True, 'love'])` yields `(1, 1, 'love')`, `(1, 'love', 1)` and `('love', 1, 1)`. Doesn't look good. I think all elements should appear.

JamesParrott:
[...]

PR #834, opened 2024-05-02
Distinct permutations of incomparable items

JamesParrott:
### Issue reference
#832

### Changes
Extends distinct_permutations to support iterables whose items cannot form a partial ordering.

### Checks and tests
```
\.local\pipx\venvs\tox\Lib\site-packages\pyproject_api\_backend.py True flit_core.buildapi
  py38: SKIP (1.48 seconds)
  py39: OK (53.61=setup[49.55]+cmd[4.06] seconds)
  py310: SKIP (0.05 seconds)
  py311: OK (17.26=setup[14.01]+cmd[3.25] seconds)
  py312: OK (40.24=setup[36.33]+cmd[3.91] seconds)
  congratulations :) (113.05 seconds)
```

```
\more-itertools>pyenv exec python -m unittest
.......................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................s................................................................................................................................................................................................................................s...................................................
----------------------------------------------------------------------
Ran 780 tests in 3.137s

OK (skipped=2)

\more-itertools>pyenv exec python --version
Python 3.8.10

\more-itertools>pyenv local 3.10.11

\more-itertools>pyenv exec python -m unittest
........................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................\more-itertools\more_itertools\more.py:1708: DeprecationWarning: zip_equal will be removed in a future version of more-itertools. Use the builtin zip function with strict=True instead.
  warnings.warn(
...............................................................................................................................................................................................................................s.....................................................
----------------------------------------------------------------------
Ran 781 tests in 2.802s

OK (skipped=1)

\more-itertools>pyenv exec python --version
Python 3.10.11
```


more_itertools/more.py
```diff
@@ -741,14 +741,37 @@ def _partial(A, r):
             i += 1
             head[i:], tail[:] = tail[: r - i], tail[r - i :]
 
-    items = sorted(iterable)
+    try:
+        items = sorted(iterable)
+        sortable = True
+    except TypeError:
+        items = list(iterable)
```
pochmann3, line 748: This doesn't work with an iterator, does it? The `sorted` consumes it and then `list` wont find anything. Better read it into a list first and then try `.sort()`.

more_itertools/more.py
```diff
@@ -741,14 +741,37 @@ def _partial(A, r):
...
+        sortable = False
+
+        # Sets are unordered.  Could use list(dict.fromkeys(items))
+        # if every item was Hashable.
+        # This is suboptimal, but both functions above are already O(n^2)
```
pochmann3, line 753: What's O(n^2)? The permuting functions? Certainly not, as there are usually more than quadratically many permutations.

more_itertools/more.py
JamesParrott: Spot on.  Indeed it doesn't.  I've fixed this now.  Shoud've checked this message first - thanks though.

more_itertools/more.py
JamesParrott: _full and _partial.  I've not looked at them in detail, I just noticed they both contained 2 or 3 for loops (in sequence) within a while loop.

I can change it to "at best O(n^2)" if you like?  The point of that comment still stands:  using O(n) sequential look up instead of an O(1) hash table for the n items, does not make things subtantially slower than it would be already

pochmann3: I'd say "are already worse than O(n^2)".

JamesParrott: Done.

bbayles:
Checking in on this... the TODO in the test cases, is that still pending?

JamesParrott:
@bbayles Well spotted.  Yep updating the documentation is still to do.  I'll take a look later.

PR #845, opened 2024-05-29
Add doublestarmap (closes #679)

### LINKED ISSUE #679: Consider introducing starstarmap

monk-time:
A continuation of work done by @Skeen in PR #680 on adding a doublestarmap function (see issue #679 for details). Hopefully this attempt will push it over the finish line. I fixed linting violations and improved the wording of the documentation.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -293,6 +293,7 @@ Others
 .. autoclass:: SequenceView
 .. autofunction:: time_limited
 .. autofunction:: map_if(iterable, pred, func, func_else=lambda x: x)
+.. autofunction:: doublestarmap
 
 ----
 
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -67,6 +67,7 @@ __all__ = [
     'distinct_permutations',
     'distribute',
     'divide',
+    'doublestarmap',
     'duplicates_everseen',
     'duplicates_justseen',
     'classify_unique',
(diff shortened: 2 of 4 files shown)


bbayles:
Thank you!

PR #846, opened 2024-05-29
Clarify perfomance cost phrasing in `divide`/`distribute` docs (closes #601)

### LINKED ISSUE #601: Confusing phrasing in documentation
The [description for `distribute`](https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.distribute) says:

> This function uses `itertools.tee()` and may require significant storage. If you need the order items in the smaller iterables to match the original iterable, see [`divide()`](https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.divide).

(The phrasing is somewhat ambiguous, and a possible interpretation of this is that `divide` does not require significant storage.)

In the [description for `divide`](https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.divide), we have:

> This function will exhaust the iterable before returning and may require significant storage. If order is not important, see [`distribute()`](https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.distribute), which does not first pull the iterable into memory.

This seems to contradict the former statement that `distribute` may require significant storage.
(Also, instead of merely saying that it _exhausts_ the iterable and _may_ require significant memory, why not simply say it converts the entire iterable to a tuple up-front? That would be both simpler and more precise. But anyway, that is not the complaint I'm making here.)

I don't know much about the internals of `itertools.tee`, but I think -- correct me if I'm wrong -- that the current implementation of `distribute` can in the worst case consume at least as much space as `divide`. Namely, if you exhaust the last of the returned iterables completely and leave others untouched, then no possible optimization could avoid having all elements in memory at once in that case.

So, my suggestion is to make it clear in the docstring of each of the two mentioned functions, that the other function is _not_ significantly more memory-efficient. The current phrasing seems to imply that each is more (or also less) efficient than the other.

monk-time:
I tried to incorporate all @plazer1's remarks from issue #601 about the slightly confusing phrasing of perfomance cost of `divide`/`distribute`. 

`divide` no longer implies that `distribute` requires strictly less memory and clarifies the difference in behavior and that in the worst case the latter is no better than the former. And `distribute` is brought closer to the structure of its counterpart.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -1624,8 +1624,10 @@ def distribute(n, iterable):
         >>> [list(c) for c in children]
         [[1], [2], [3], [], []]
 
-    This function uses :func:`itertools.tee` and may require significant
-    storage. If you need the order items in the smaller iterables to match the
+    This function uses :func:`itertools.tee` and in the worse case may require
+    significant storage.
+
+    If you need the order of items in the smaller iterables to match the
     original iterable, see :func:`divide`.
 
     """
@@ -1873,9 +1875,12 @@ def divide(n, iterable):
         >>> [list(c) for c in children]
         [[1], [2], [3], [], []]
 
-    This function will exhaust the iterable before returning and may require
-    significant storage. If order is not important, see :func:`distribute`,
-    which does not first pull the iterable into memory.
+    This function will convert the iterable to a tuple upfront and thus may
+    require significant storage.
+
+    If order is not important, see :func:`distribute`, which only consumes
+    the iterable on demand but in the worst case could still keep the entire
+    iterable in memory.
 
     """
     if n < 1:


more_itertools/more.py
```diff
@@ -1873,9 +1875,12 @@ def divide(n, iterable):
         >>> [list(c) for c in children]
         [[1], [2], [3], [], []]
 
-    This function will exhaust the iterable before returning and may require
-    significant storage. If order is not important, see :func:`distribute`,
-    which does not first pull the iterable into memory.
+    This function will convert the iterable to a tuple upfront and thus may
```
pochmann3, line 1878: That's not true. It only does that if the slicing attempt fails.

monk-time: Fair. Would it be better to revert the phrasing here back to "exhaust the iterable" or would inserting "(if needed)" suffice?

pochmann3: Not sure. It's tricky. "Exhausting" a list for example even sounds *wrong*. And the later slices `seq[start:stop]` can *also* use significant storage.

bbayles:
Thanks for this attempt. These additions feel a little bit cumbersome to me, however... what do you think about the change I made [here](https://github.com/more-itertools/more-itertools/commit/d5cd09a16218bee9f12a9247e821fc67ea77ebef)?

The things I want to convey are:
* If order isn't important, you can use `distribute`
* If order is important, you can use `divide`
* If you're inclined to think that these are all lazily-evaluated iterators, please be careful because they can still involve storing a bunch of stuff in memory.

monk-time:
@bbayles I feel like your changes are a reasonable compromise that solves the original issue. I'm OK with closing this PR.

Issue #843, opened 2024-05-27
Proposal: Discrete Fourier Transform


rhettinger:
```
from math import e, tau, sumprod
from itertools import chain
from typing import Iterator, Sequence

def _sumprod(v1: Sequence[complex], v2: Sequence[complex]) -> complex:
    "High precision sumprod() for complex numbers."
    r1 = chain((p.real for p in v1), (-p.imag for p in v1))
    r2 = chain((q.real for q in v2), (q.imag for q in v2))
    i1 = chain((p.real for p in v1), (p.imag for p in v1))
    i2 = chain((q.imag for q in v2), (q.real for q in v2))
    return complex(sumprod(r1, r2), sumprod(i1, i2))

def dft(xarr: Sequence[complex]) -> Iterator[complex]:
    "Discrete Fourier Transform"
    N = len(xarr)
    roots_of_unity = [e ** (n / N * tau * -1j) for n in range(N)]
    for k in range(N):
        coeffs = [roots_of_unity[k * n % N] for n in range(N)]
        yield _sumprod(xarr, coeffs)

def idft(Xarr: Sequence[complex]) -> Iterator[complex]:
    "Inverse Discrete Fourier Transform"
    N = len(Xarr)
    roots_of_unity = [e ** (n / N * tau * 1j) for n in range(N)]
    for k in range(N):
        coeffs = [roots_of_unity[k * n % N] for n in range(N)]
        yield _sumprod(Xarr, coeffs) / N


if __name__ == '__main__':

    from random import randrange, random
    import cmath

    # https://en.wikipedia.org/wiki/Discrete_Fourier_transform#Example
    xarr = [1, 2-1j, -1j, -1+2j]
    Xarr = [2, -2-2j, -2j, 4+4j]
    assert all(map(cmath.isclose, dft(xarr), Xarr))
    assert all(map(cmath.isclose, idft(Xarr), xarr))

    # Round-trip for random inputs
    for _ in range(1_000):
        N = randrange(35)
        xarr = [complex(random(), random()) for i in range(N)]
        Xarr = list(dft(xarr))
        assert all(map(cmath.isclose, idft(Xarr), xarr))
```

P.S. It would be nice if the main docs page grew a category for mathematical tools:  matmul, convolve, polynomial manipulation etc.

bbayles:
[Here](https://github.com/more-itertools/more-itertools/pull/847) is an implementation. Thanks for this contribution!

pochmann3:
[...]

PR #847, opened 2024-05-30
Add dft and idft

bbayles:
This PR adds `dft` and `idft`, contributed by @rhettinger in #843.

Additionally, the docs now have a separate section for the various math-related functions.


more_itertools/more.py
```diff
@@ -4726,3 +4729,53 @@ def join_mappings(**field_to_map):
             ret[key][field_name] = value
 
     return dict(ret)
+
+
+def _complex_sumprod(v1, v2):
```
bbayles, line 4734: Since `_sumprod` is Python 3.12+, I'm using the backup implementation from `recipes` for older versions.

rhettinger: So that the high-precision calculation doesn't get wasted, I recommend an alternate `_sumprod` implemented as `math.fsum(map(operator.mul, v1, v2))`

tests/test_more.py
```diff
@@ -5803,3 +5804,18 @@ def test_basic(self):
 
     def test_empty(self):
         self.assertEqual(dict(mi.join_mappings()), {})
+
+
+class DiscreteFourierTransformTests(TestCase):
+    def test_basic(self):
+        xarr = [1, 2 - 1j, -1j, -1 + 2j]
```
rhettinger, line 5811: Suggest adding a comment that links back to the wikipedia example which is worked out step-by-step: `# https://en.wikipedia.org/wiki/Discrete_Fourier_transform#Example`.

tests/test_more.py
```diff
@@ -5803,3 +5804,18 @@ def test_basic(self):
...
+        Xarr = [2, -2 - 2j, -2j, 4 + 4j]
+        self.assertTrue(all(map(isclose, mi.dft(xarr), Xarr)))
+        self.assertTrue(all(map(isclose, mi.idft(Xarr), xarr)))
+
+    def test_roundtrip(self):
+        for _ in range(1_000):
+            N = randrange(35)
+            xarr = [complex(random(), random()) for i in range(N)]
+            Xarr = list(mi.dft(xarr))
+            assert all(map(isclose, mi.idft(Xarr), xarr))
```
rhettinger, line 5821: I would spell this out as `cmath.isclose` rather than just `isclose` because the semantics are so different.

more_itertools/more.py
```diff
@@ -4726,3 +4729,53 @@ def join_mappings(**field_to_map):
...
+    >>> from cmath import isclose
+    >>> xarr = [1, 2-1j, -1j, -1+2j]
+    >>> Xarr = [2, -2-2j, -2j, 4+4j]
+    >>> all(map(isclose, dft(xarr), Xarr))
+    True
+
+    See :func:`idft` for the inverse Discrete Fourier Transform.
+    """
+    N = len(xarr)
+    roots_of_unity = [e ** (n / N * tau * -1j) for n in range(N)]
```
rhettinger, line 4758: I think replacing `e ** (` with `cmath.exp(` would be fine. Even though the `cmath.exp` form is less attractive, it is a bit faster on my machine.  Also in theory, the specialized `cmath.exp` form should be a least as good as the general purpose `e **` form.  In practice, both ways give identical results on my machine.

more_itertools/more.py
bbayles: Added in acdbb80fe181406eab4466db8087019d862a0f46

more_itertools/more.py
bbayles: I'll leave the `e ** (` notation now, because that's how I learned it in EE243.

pochmann3:
Title has a typo, says `dft` twice. And did you add `doublestarmap` here intentionally?

bbayles:
Thanks, I fixed the title.

The `doublestarmap` PR conflicted with this one.

rhettinger:
LGTM :-)

PR #855, opened 2024-06-11
Issue 854: sample improvements

### LINKED ISSUE #854: Issues with sample()
I've only just started looking at this and have a few thoughts.

The comparison to the standard library should mention:

* The `counts` parameter in the `random.sample()` is not comparable to `weights` in `more_itertools.sample()`
The former allows an element to be selected multiple times to model classic "urn problems".  The latter just affects the likelihood of an element being selected at all.

* The standard library `random.sample()` raises an exception if *k* is larger than the population.  However `more_itertools.sample()` just returns the population unchanged and in order:

```
>>> import more_itertools
>>> more_itertools.sample(range(10), k=15)
[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]

>>> import random
>>> random.sample(range(10), k=15)
   ...
ValueError: Sample larger than population or is negative
```
* The standard library `random.sample()` guarantees the results to be fully shuffled.  In contrast, the `more_itertools.sample()` initial reservoir fill is not shuffled at all:

```
>>> import more_itertools
>>> import random

>>> random.seed(8675309)
>>> more_itertools.sample(range(2000), k=1990)[:15]
[0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14]

>>> random.seed(8675309)
>>> random.sample(range(2000), k=1990)[:15]
[823, 438, 1044, 1923, 1359, 1032, 1762, 1838, 575, 1764, 1313, 1674, 1325, 1386, 465]
```

It would be great if a `random.shuffle()` step were added at the end. If not, then a `sorted()` step would be an improvement over having a result that is neither randomized or sorted.

bbayles:
This PR makes some changes to sample:
* `shuffle` is called before returning the output
* A note about differences from `random.sample` is added to the docstring
* `ValueError` is raised for negative `k` values

Closes #854

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -4,7 +4,7 @@ import warnings
 from collections import Counter, defaultdict, deque, abc
 from collections.abc import Sequence
 from functools import cached_property, partial, reduce, wraps
-from heapq import heapify, heapreplace, heappop
+from heapq import heapify, heapreplace
 from itertools import (
     chain,
     combinations,
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -3597,8 +3597,7 @@ def _sample_weighted(iterator, k, weights, strict):
         else:
             weights_to_skip -= weight
 
-    # Equivalent to [element for weight_key, element i n sorted(reservoir)]
-    ret = [heappop(reservoir)[1] for _ in range(k)]
+    ret = [element for weight_key, element in reservoir]
     shuffle(ret)
     return ret
 
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -23,7 +23,7 @@ from itertools import (
 )
 from math import comb, e, exp, factorial, floor, fsum, log, perm, tau
 from queue import Empty, Queue
-from random import random, randrange, uniform
+from random import random, randrange, shuffle, uniform
 from operator import itemgetter, mul, sub, gt, lt, ge, le
 from sys import hexversion, maxsize
 from time import monotonic
@@ -3557,6 +3557,7 @@ def _sample_unweighted(iterable, k):
             W *= exp(log(random()) / k)
             next_index += floor(log(random()) / log(1 - W)) + 1
 
+    shuffle(reservoir)
     return reservoir
 
 
@@ -3592,19 +3593,21 @@ def _sample_weighted(iterable, k, weights):
             weights_to_skip -= weight
 
     # Equivalent to [element for weight_key, element in sorted(reservoir)]
+    shuffle(reservoir)
     return [heappop(reservoir)[1] for _ in range(k)]
 
 
 def sample(iterable, k, weights=None):
     """Return a *k*-length list of elements chosen (without replacement)
-    from the *iterable*. Like :func:`random.sample`, but works on iterables
-    of unknown length.
+    from the *iterable*. Similar to :func:`random.sample`, but works on
+    iterables of unknown length.
 
     >>> iterable = range(100)
     >>> sample(iterable, 5)  # doctest: +SKIP
     [81, 60, 96, 16, 4]
 
-    An iterable with *weights* may also be given:
+    An iterable with *weights* may also be given (this is not analagous to
+    :func:`random.sample`'s  ``counts`` parameter):
 
     >>> iterable = range(100)
     >>> weights = (i * i + 1 for i in range(100))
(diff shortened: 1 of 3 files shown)


more_itertools/more.py
```diff
@@ -3592,19 +3593,21 @@ def _sample_weighted(iterable, k, weights):
...
     # Equivalent to [element for weight_key, element in sorted(reservoir)]
+    shuffle(reservoir)
     return [heappop(reservoir)[1] for _ in range(k)]
 
 
 def sample(iterable, k, weights=None):
     """Return a *k*-length list of elements chosen (without replacement)
-    from the *iterable*. Like :func:`random.sample`, but works on iterables
-    of unknown length.
+    from the *iterable*. Similar to :func:`random.sample`, but works on
```
rhettinger, line 3602: Consider documenting that if the input iterable is sized k or smaller, `more_itertools.sample()` will return all of the input in shuffled order, possibly giving a result length smaller than *k*.

Alternatively, consider adding a `strict` option to raise an exception as `random.sample()` does. That would let the user decide whether it is an error to say, "choose 100 distinct notes on an 88-key piano."

rhettinger:
Overall, this looks like an excellent set of edits. Thank you.

rhettinger:
Please also fix the bad variable name starting in `sample()` and flowing through to the two private functions.

```
- iterable = iter(iterable)
+ iterator = iter(iterable)
```

Both private functions ONLY work with iterators, not generic iterables because the `take()` step changes state by advancing the iterator and the subsequent enumeration relies on that state change.

rhettinger:
Unless I'm missing something, the variable names *inside* the main function can be changed (leaving the user visible API unchanged).  Something like this:

```
def _sample_unweighted(iterator, k, strict):
    ...

def _sample_weighted(iterator, k, weights, strict):
    ...

def sample(iterable, k, weights=None, strict=False):
    if k < 0:
        raise ValueError('k must be non-negative')
    if k == 0:
        return []

    iterator = iter(iterable)     # <-- CHANGE HERE
    if weights is None:
        return _sample_unweighted(iterator, k, strict)   # <-- CHANGE HERE
    else:
        weights = iter(weights)
        return _sample_weighted(iterator, k, weights, strict)   # <-- CHANGE HERE
```

Would it make sense to add stubs for the private functions to let a type checker verify that iterators are passed in?

```
def _sample_unweighted(
     iterator: Iterator[_T],
     k: int,
     strict: bool = ...,
 ) -> list[_T]: ...

def _sample_weighted(
     iterable: Iterator[_T],
     k: int,
     weights: Iterator[float] | None = ...,
     strict: bool = ...,
 ) -> list[_T]: ...

```

rhettinger:
The explanation of the *weights* parameter seems way off, "The relative weight of each item determines the probability that it appears late in the permutation."

The wording in the Efraimidis paper is better, "In weighted random sampling (WRS) the items are weighted and the probability of each item to be selected is determined by its relative weight."

However even that only explains the probability of being added to the reservoir. From a users point of view, it misses an essential point:  when an item with high weight is selected, the probabilities of the remaining items increase.  

Ideally, there should be some wording that allows a user to predict this outcome:

```
>>> Counter(chain.from_iterable(more_itertools.sample('abc', 2, [98, 1, 1]) for i in range(100_000)))
Counter({'a': 99978, 'c': 50136, 'b': 49886})
```

bbayles:
I'm open to suggestions on how much detail is appropriate. I could link to the papers, perhaps?

rhettinger:
The papers are a difficult read. They only explain the algorithm and don't elaborate on the goal.

Some documentation edit does need to be made because the current sentence is just wrong. There is no notion of "appearing late in the permutation".

Maybe this:

> Weighted selections are made without replacement. After an element is selected, it is removed from the pool and the relative weights of the other elements increase.  The final result is shuffled so that weights only affect the probability of selection but not the position in the final permutation.

bbayles:
Thanks - I incorporated your suggestions. Wikipedia has descriptions of both algorithms, which I've linked for those who are interested in more detail.

image

more_itertools/more.py
```diff
@@ -3597,9 +3597,10 @@ def _sample_weighted(iterator, k, weights, strict):
         else:
             weights_to_skip -= weight
 
-    # Equivalent to [element for weight_key, element in sorted(reservoir)]
-    shuffle(reservoir)
-    return [heappop(reservoir)[1] for _ in range(k)]
+    # Equivalent to [element for weight_key, element i n sorted(reservoir)]
+    ret = [heappop(reservoir)[1] for _ in range(k)]
```
rhettinger, line 3601: Since this gets shuffled later, there is no need for ordering with *heappop()*. 

This line can be simplified to:

`ret = [element for weight_key, element in reservoir]`

bbayles: Hmm, isn't the point of doing the `heappop` to get the first `k` as ordered by the preceeding steps?

rhettinger: Double check me on this, but it looks like reservoir is always of size *k*.  The `take()` step puts in *k* elements. The `heapify` step does not change the size, nor does the `heapreplace()` step.

So the "first k" in the reservoir is the same as the entire reservoir.

bbayles: Ah, you're right! Thanks - I stand corrected.

Issue #849, opened 2024-06-03
TypeError: type 'peekable' is not subscriptable


powellnorma:
It would be great when we could hint the type e.g. like:

```py
def f(it: peekable[str]):
    pass
```

Pandede:
As this repository uses stub files instead of writing the typing hint directly in the Py file, the only way to use generic typing hint is literal string (same as your workaround):
```python
def f(it: 'peekable[str]'):
    pass
```
The similar issue is also stated [here](https://github.com/python/typing/issues/60) and the author of Python @gvanrossum said he likes the "literal string" solution.

bbayles:
I'll accept a PR that makes the relevant change, thanks.

m9810223:
#857 I'm not sure the implementation is ideal.

PR #857, opened 2024-06-13
feat: `peekable` support subscriptable

m9810223:
close https://github.com/more-itertools/more-itertools/issues/849


bbayles (changes requested):
Could you move the type annotations to `more.pyi`?

m9810223:
> Could you move the type annotations to `more.pyi`?

I noticed that the stub file exists.
And the problem is that the class in `more.py` isn't inherited from `Iterator`.

bbayles:
It's OK with me if you change it to inherit from `collections.abc.Iterator` in the main file.

m9810223:
> It's OK with me if you change it to inherit from `collections.abc.Iterator` in the main file.

After testing it locally, I found that the key is to let calss inherit from `Generic`.

bbayles (changes requested):
I would like the type annotations to live in `more.pyi`, please.

m9810223:
> I would like the type annotations to live in `more.pyi`, please.

At the moment it is. 
But it does not work.

https://github.com/more-itertools/more-itertools/blob/4bd0bd876273a8ff2171352c937b91a8c053d00b/more_itertools/more.pyi#L64

m9810223:
OK. But I don't think it's reachable because #849 is a runtime situation.

saaketp:
Instead of inheriting from Generic you could define a [`__class_getitem__`](https://docs.python.org/3/reference/datamodel.html#emulating-generic-types).

Something like this:
```python
class peekable:
   ...
   def __class_getitem__(cls, key):
      return cls
```

m9810223:
@saaketp Thanks! Let me try!

m9810223:
@bbayles

The `__class_getitem__` solution works.

Should we update other classes like `bucket` ... together?

saaketp:
> It's OK with me if you change it to inherit from `collections.abc.Iterator` in the main file.

This actually works too for python 3.9 and above, and is definitely nicer than the `__class_getitiem__` solution.
So maybe we could wait few more months till python 3.8 goes out of support and then do this instead.

stankudrow:
Any updates?

bbayles:
Given that the current release has 3.9 as the minimum supported version, I'm fine with the change described above as long as it lives in the `.pyi` file.

rhettinger:
This would go directly in the class source code and not in a PYI file.  I don't think there is any other way to make it work.  

FWIW, that is what we've done in CPython which does not have otherwise have type annotations or pyi files.  See [`functools.partial()`](https://github.com/python/cpython/blob/d05140f9f77d7dfc753dd1e5ac3a5962aaa03eff/Lib/functools.py#L431) for an example.

m9810223:
I agree with @rhettinger.

> This would go directly in the class source code and not in a PYI file. I don't think there is any other way to make it work.

more_itertools/more.py
```diff
@@ -431,6 +431,9 @@ def __getitem__(self, index):
 
         return self._cache[index]
 
+    def __class_getitem__(cls, key):
+        return cls
+
```
rhettinger, line 436: This should be: `__class_getitem__ = classmethod(GenericAlias)`

Also suggest adding a test to make sure it works: `self.assertEqual(peekable[str].__args__,
(str,))`

PR #866, opened 2024-07-03
Issue 862: change relative_seek() behaviour

dkrikun:
### Issue reference
Closes 862

### Changes
Change relative_seek() behaviour to seek relative to current position (possibly within the cache), as opposed to, seeking to the cache head, which had been implemented previously.

### Checks and tests
Passes make all-checks

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -2823,9 +2823,9 @@ class seekable:
         >>> it.relative_seek(-3)  # Source is at '6', we move back to '3'
         >>> next(it)
         '3'
-        >>> it.relative_seek(-3)  # Source is still at '6', we move back to '3'
+        >>> it.relative_seek(-3)  # Source is at '3', we move back to '1'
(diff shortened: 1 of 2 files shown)


more_itertools/more.py
```diff
@@ -2823,9 +2823,9 @@ class seekable:
         >>> it.relative_seek(-3)  # Source is at '6', we move back to '3'
         >>> next(it)
         '3'
-        >>> it.relative_seek(-3)  # Source is still at '6', we move back to '3'
+        >>> it.relative_seek(-3)  # Source is at '3', we move back to '1'
```
bbayles, line 2826: Should `Source is at '4'`, right? That's the one that would be emitted if `next()` were called.

dkrikun: You are right, should be '4'

dkrikun: Fixed now

Issue #856, opened 2024-06-12
Extend set_partitions with max_size, min_size parameters


JPBotelho:
### Description
Allow an optional parameter to be passed onto set_partitions that sets the maximum size per block in the partition. 
Allow a similar parameter for the minimum size of the blocks.

If there is interest I can have a look at implementing this. There might be more efficient algorithms than the recursive one used now ( see https://link.springer.com/chapter/10.1007/3-540-52921-7_57 ). It might even be possible to use an algorithm that allows for parallelization (split generation space into K, execute iterative algorithm starting at each split).

### References
These correspond to restricted and associated Stirling numbers of the second kind: 
https://setpartitions.org/associatedStirling.html
https://setpartitions.org/restrictedStirling.html

### Examples

Current usage (still valid):
```
iterable = 'abc'
for part in set_partitions(iterable):
    print([''.join(p) for p in part])

['abc']
['a', 'bc']
['ab', 'c']
['b', 'ac']
['a', 'b', 'c']
```
Proposed:
```
iterable = 'abc'
for part in set_partitions(iterable, max_size=2):
    print([''.join(p) for p in part])

['a', 'bc']
['ab', 'c']
['b', 'ac']
['a', 'b', 'c']
```

```
iterable = 'abc'
for part in set_partitions(iterable, min_size=2):
    print([''.join(p) for p in part])

['abc']
```

bbayles:
I'll definitely accept a PR that changes the `set_partitions` implementation from recursive. If you're willing to do that, go for it!

PR #872, opened 2024-07-06
Issue 856: Add argument `min_size` and `max_size` to `set_partitions`

Pandede:
### Issue reference

Closes more-itertools/more-itertools#856

### Changes

Add argument `min_size` and `max_size` to `set_partitions` to limit the size per block in a partition.

### Checks and tests

Done.


bbayles:
Thanks for the PR. Per [my comment here](https://github.com/more-itertools/more-itertools/issues/856#issuecomment-2176400116), I was hoping to get a non-recursive version of `set_partitions`. I don't suppose you have an idea on how to do that?

Pandede:
@bbayles Sure, I may give it a try.

Pandede:
May I ask a question: Is calling `set_partitions(iterable, k=None)` equivalent to calling `partitions(iterable)`?

Pandede:
By comparing the time efficiency between two methods, the recursive one is better than iterative one:
```
Before:  0.7022837000258733
After:  2.7836228999949526
```
<details>

<summary> Sample code </summary>

```python
def set_partitions_recursive(iterable, k=None, min_size=None, max_size=None):
    L = list(iterable)
    n = len(L)
    if k is not None:
        if k < 1:
            raise ValueError(
                "Can't partition in a negative or zero number of groups"
            )
        elif k > n:
            return

    min_size = min_size if min_size is not None else 0
    max_size = max_size if max_size is not None else n
    if min_size > max_size:
        return

    def set_partitions_helper(L, k):
        n = len(L)
        if k == 1:
            yield [L]
        elif n == k:
            yield [[s] for s in L]
        else:
            e, *M = L
            for p in set_partitions_helper(M, k - 1):
                yield [[e], *p]
            for p in set_partitions_helper(M, k):
                for i in range(len(p)):
                    yield p[:i] + [[e] + p[i]] + p[i + 1 :]

    if k is None:
        for k in range(1, n + 1):
            yield from filter(
                lambda z: all(min_size <= len(bk) <= max_size for bk in z),
                set_partitions_helper(L, k),
            )
    else:
        yield from filter(
            lambda z: all(min_size <= len(bk) <= max_size for bk in z),
            set_partitions_helper(L, k),
        )


def set_partitions_iterative(iterable, k=None, min_size=None, max_size=None):
    L = list(iterable)
    n = len(L)
    if k is not None:
        if k < 1:
            raise ValueError(
                "Can't partition in a negative or zero number of groups"
            )
        elif k > n:
            return

    min_size = min_size if min_size is not None else 1
    max_size = max_size if max_size is not None else n
    if min_size > max_size:
        return

    def set_partitions_helper(i, path):
        k = len(path)
        if k == 1:
            yield [L]
        elif k == n:
            yield [[s] for s in L]
        else:
            stack = [(i, path)]
            while stack:
                i, path = stack.pop()

                if i >= n:
                    yield path
                else:
                    for j in range(len(path)):
                        stack.append(
                            (
                                i + 1,
                                path[:j] + [path[j] + [L[i]]] + path[j + 1 :],
                            )
                        )
                        if not path[j]:
                            break

    if k is None:
        for k in range(1, n + 1):
            yield from filter(
                lambda z: all(min_size <= len(bk) <= max_size for bk in z),
                set_partitions_helper(0, [[] for _ in range(k)]),
            )
    else:
        yield from filter(
            lambda z: all(min_size <= len(bk) <= max_size for bk in z),
            set_partitions_helper(0, [[] for _ in range(k)]),
        )


if __name__ == '__main__':
    from timeit import timeit

    setup_code = r'from __main__ import set_partitions_recursive, set_partitions_iterative'
    stmt_before = r'for _ in set_partitions_recursive(range(7)): pass'
    stmt_after = r'for _ in set_partitions_iterative(range(7)): pass'

    print('Before: ', timeit(stmt=stmt_before, setup=setup_code, number=1000))
    print('After: ', timeit(stmt=stmt_after, setup=setup_code, number=1000))
```

</details>

bbayles:
I merged the recursive version. Thanks!

Issue #885, opened 2024-07-27
Type hint for collapse base_type should accept a tuple of types


jbosboom:
`collapse`'s `base_type` argument is currently hinted as `type | None`.  It should accept `tuple[type, ...]`.

`isinstance` also accepts `types.UnionType`.  Passing a union is (was?) slower than a tuple of types, so maybe it's fine to leave unions out as a nudge.

Compare `always_iterable`'s `base_type` argument, which is hinted as `type | tuple[type | tuple[Any, ...], ...] | None`.  I'm surprised by the `Any`, so I haven't gone straight to a pull request.

bbayles:
I'll accept a PR that improves the situation -t hanks.

PR #887, opened 2024-07-28
Standardize type hints for isinstance's second argument

jbosboom:
always_iterable, collapse, and zip_broadcast take arguments that are used as the second argument to isinstance.  Before this commit, collapse's type hint was too narrow (rejected tuples of types) and always_iterable/zip_broadcast's type hint was too broad (accepted Any in nested tuples).  All three also omitted UnionType, introduced in Python 3.10, which can also be the second argument of isinstance (provided it does not contain types.GenericAlias).

This commit standardizes these three functions on [typeshed's hint for isinstance](https://github.com/python/typeshed/blob/ea14235f74a0c8c6e58dac06c1d554f6bac74a4e/stdlib/builtins.pyi#L1446).

---

### Issue reference
#885 

### Changes
See the commit message/PR description.

Previously more.pyi used only `from X import Y`.  This commit imports `sys` because [type checkers are allowed to match sys.versioninfo literally](https://typing.readthedocs.io/en/latest/reference/stubs.html#version-and-platform-checks).  This commit also imports `types` because putting `from types import UnionType` inside the version check causes black to add an ugly blank line.

### Checks and tests
`make all-checks` passed.

### DIFF
--- a/more_itertools/more.pyi
+++ b/more_itertools/more.pyi
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import sys, types
 from types import TracebackType
 from typing import (
     Any,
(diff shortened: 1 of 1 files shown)


bbayles:
Thanks!

PR #892, opened 2024-08-02
Add option `strict` to `sort_together`

### LINKED ISSUE #890: Add a strict option to sort_together()
A *strict* option would detect data size mismatches. This would avoid silently dropping data:

```
>>> iterables = [(4, 3, 2, 1), ('a', 'b', 'c')]
>>> sort_together(iterables)
[(2, 3, 4), ('c', 'b', 'a')]
```

Pandede:
### Issue reference

Closes #890 

### Changes

Add option `strict` to detect the mismatch dimension of iterables.

### Checks and tests

Done.


bbayles:
This will consume the iterables, right? So if you don't pass in `lists` or `tuples`, nothing will happen?

Pandede:
@bbayles I just recognize that using `zip_equal` is the simplest solution. I may modify the solution in next commits.

bbayles (approved):

bbayles:
This looks good - if you'll run the formatter (`make format`), I'll merge it. Thanks!

PR #899, opened 2024-08-15
Optimize all_equal recipe

### LINKED ISSUE #896: all_equal() improvement
The `all_equal()` recipe in the main itertools docs was chosen because it is a one liner, it is reasonably clear, and it nicely shows off how various tools can be chained together in a functional style.  

For the production version in `more-itertools`, we can do better (faster startup overhead and fewer dependencies):

```python
def all_equal(iterable, key=None):
    "Returns True if all the elements are equal to each other."
    # all_equal('4٤௪౪໔', key=int) → True

    iterator = groupby(iterable, key=key)
    try:
        next(iterator)
        next(iterator)
    except StopIteration:
        return True
    return False
```

Here are some timings that demonstrate reduced overhead for each of the cases (zero groups, one group, two or more groups):

```
% python3.13 -m timeit -s 'from itertools import islice, groupby' -s 'key=None' -s 'iterable=""' 'len(list(islice(groupby(iterable, key), 2))) <= 1' 
1000000 loops, best of 5: 233 nsec per loop
% python3.13 -m timeit -s 'from itertools import islice, groupby' -s 'key=None' -s 'iterable="a"' 'len(list(islice(groupby(iterable, key), 2))) <= 1'
1000000 loops, best of 5: 256 nsec per loop
% python3.13 -m timeit -s 'from itertools import islice, groupby' -s 'key=None' -s 'iterable="ab"' 'len(list(islice(groupby(iterable, key), 2))) <= 1'
1000000 loops, best of 5: 289 nsec per loop

% python3.13 -m timeit -s 'from itertools import islice, groupby' -s 'key=None' -s 'iterable=""' 'it=groupby(iterable)' 'try:' ' next(it)' ' next(it)' 'except StopIteration:' ' True' 'else:' ' False'
2000000 loops, best of 5: 150 nsec per loop
% python3.13 -m timeit -s 'from itertools import islice, groupby' -s 'key=None' -s 'iterable="a"' 'it=groupby(iterable)' 'try:' ' next(it)' ' next(it)' 'except StopIteration:' ' True' 'else:' ' False' 
2000000 loops, best of 5: 198 nsec per loop
% python3.13 -m timeit -s 'from itertools import islice, groupby' -s 'key=None' -s 'iterable="ab"' 'it=groupby(iterable)' 'try:' ' next(it)' ' next(it)' 'except StopIteration:' ' True' 'else:' ' False'
2000000 loops, best of 5: 174 nsec per loop
```

Downstream code such as `iequals()` will benefit most because it uses `all_equal()` in a loop.

bbayles:
This PR updates `all_equal` with @rhettinger's new version.

Closes #896

### DIFF
--- a/more_itertools/recipes.py
+++ b/more_itertools/recipes.py
@@ -218,13 +218,11 @@ def all_equal(iterable, key=None):
         True
 
     """
-    iterator = groupby(iterable, key=key)
-    try:
-        next(iterator)
-        next(iterator)
-    except StopIteration:
-        return True
-    return False
+    iterator = groupby(iterable, key)
+    for first in iterator:
+        for second in iterator:
+            return False
+    return True
 
 
 def quantify(iterable, pred=bool):
--- a/more_itertools/recipes.py
+++ b/more_itertools/recipes.py
@@ -218,7 +218,13 @@ def all_equal(iterable, key=None):
         True
 
     """
-    return len(list(islice(groupby(iterable, key), 2))) <= 1
+    iterator = groupby(iterable, key=key)
+    try:
+        next(iterator)
+        next(iterator)
+    except StopIteration:
+        return True
+    return False
 
 
 def quantify(iterable, pred=bool):


pochmann3:
Did you overlook my benchmark where this was **slower** than the current version for the case of one group? (Which is the most relevant one for the mentioned usage un `iequals`.)

bbayles:
Thanks for the comment. To be totally transparent, I am going to favor Raymond's suggestions in almost every case – this library is a wrapper around his work, and unless I have a big disagreement with one of his ideas, I'm going to implement them. You have a good track record of contributing to this library as well, and as such I am very open to your suggestion too. That said, I generally discount microbenchmarks on a single machine with un-representative inputs, which is what appears to be the basis for the "this is slower" claim.

rhettinger:
Pochmann later posted a for/for variant that is much better than his other proposals. After changing the variable name from `groups` to `iterator`, I'm happy with it:

```
def all_equal(iterable, key=None):
    iterator = groupby(iterable, key)
    for first in iterator:
        for second in iterator:
            return False
    return True
```

I don't think we can do better than this.  It is short and relies only on the most basic of Python features.  It is a little less straight-forward than the `try/except` variant but only has a small and tolerable bit of weirdness.  This is the one I recommend.

rhettinger (approved):

pochmann3:
> I generally discount microbenchmarks on a single machine with un-representative inputs, which is what appears to be the basis for the "this is slower" claim.

As shown in the issue, I also reproduced my findings on another machine. And reproduced Raymond's disagreeing findings on the same machine *when measuring his command line way without a function*. I still don't know why measuring that way gives such very different results, but that way isn't the realistic way. The real `all_equal` is a function.

About *"un-representative inputs"*: They do represent all cases: No groups, one group, more than one group. And all solutions under consideration use `groupby`. So those inputs *are* fully representative for the differences of the solutions.

I also like the normal version of my "gross" hack:

```python
def all_equal(iterable, key=None):
    seen_one = False
    for _ in groupby(iterable, key):
        if seen_one:
            return False
        seen_one = True
    return True
```

It's similarly fast:

```
iterable = ()
True  137 Â± 0.3 ns  Stefan_hack
True  140 Â± 0.3 ns  Stefan_seen
True  147 Â± 0.7 ns  Stefan_loops
True  306 Â± 0.9 ns  current
True  356 Â± 0.8 ns  Raymond

iterable = (1,)
True  205 Â± 0.3 ns  Stefan_hack
True  208 Â± 0.7 ns  Stefan_seen
True  221 Â± 0.8 ns  Stefan_loops
True  355 Â± 1.3 ns  current
True  435 Â± 1.0 ns  Raymond

iterable = (1, 2)
False  251 Â± 0.4 ns  Stefan_loops
False  252 Â± 0.9 ns  Stefan_seen
False  255 Â± 0.8 ns  Stefan_hack
False  329 Â± 1.1 ns  Raymond
False  396 Â± 0.9 ns  current

Python: 3.13.0rc1+ (heads/3.13:009f9efe658, Aug 15 2024, 14:30:32) [Clang 18.1.6 (Fedora 18.1.6-3.fc40)]
```

<details><summary>Benchmark script</summary>

```python
def current(iterable, key=None):
    return len(list(islice(groupby(iterable, key), 2))) <= 1

def Raymond(iterable, key=None):
    iterator = groupby(iterable, key=key)
    try:
        next(iterator)
        next(iterator)
    except StopIteration:
        return True
    return False

def Stefan_hack(iterable, key=None):
    for _ in groupby(iterable, key):
        if iterable is None:
            return False
        iterable = None
    return True

def Stefan_loops(iterable, key=None):
    iterator = groupby(iterable, key)
    for first in iterator:
        for second in iterator:
            return False
    return True

def Stefan_seen(iterable, key=None):
    seen_one = False
    for _ in groupby(iterable, key):
        if seen_one:
            return False
        seen_one = True
    return True


funcs = [current, Raymond, Stefan_hack, Stefan_loops, Stefan_seen]

iterables = (), (1,), (1, 2)

from timeit import timeit
from statistics import mean, stdev
import sys
import random
from itertools import *

for iterable in iterables:
    print(f'{iterable = }')
    times = {f: [] for f in funcs}
    def stats(f):
        ts = [t * 1e9 for t in sorted(times[f])[:10]]
        return f' {mean(ts):3.0f} ± {stdev(ts):3.1f} ns '
    for _ in range(100):
        random.shuffle(funcs)
        for f in funcs:
            t = timeit(lambda: f(iterable), number=1000) / 1000
            times[f].append(t)
    for f in sorted(funcs, key=stats):
        print(f(iterable), stats(f), f.__name__)
    print()

print('Python:', sys.version)
```

[Attempt This Online!](https://ato.pxeger.com/run?1=nVXBjtMwEL1xyIlPmFtjFEoLF4joFYkLQiy3qorSZLyNNrEj21ltVPVLuOwF7vwCn8HXMLaTxi0tAnJoE49n5r2ZN_aXb21vdlI8Pn7tDH_--ueTpyVyKDqlUJi4MqjybY0J3GG_-iAFsjQCehSaTgmoUcR1pWmjrqsC41slu3bbn_qxBF4yxuDtCpZRZMN_yvtGivJ6eGcwUsEKLoZc2bBup1G9d7GPwIcBM_myPy3jQ4GtgRsj2_fOUEkxxRnYfVYdhmzf5bVGz-DGIM9FtsuLu-ssOBHIoBKXObApX8VhNEGlwYaYjL8BOHqNLivnEZ0jD4HWUrb6_-vNjnx4pbSxnEaXCag1ayyosZftF4lcA6yRpHUVr7Vm9El4p1D_Uu4xwF-UOch1LgiPOuKdKDSZ18PYJKPAk1AnyUkvkpDoJopGrDZOTBMTLxP_S7NDGZRswFQNVlT7ppXKDF_eog0JWJuKUAzWBnOR0HqJ99GwpHs9vqpclLLxvjaxkbI-uj6jdFTJSZDi-K59uVpV0dnAZ_tAgYfZMI8Ey3LY8xTWGy8ZG8LV6OC22C5bxDrmQVOMqyClhyW-cX5OZpogYRm7sGu-Yet0udhszieVz2BvKcdGs_TVfMEP8OM77B3_YW1Ja0LD7FQqVIlbjJeLRYDEV2eudx3nNcYOOTtR-cToVD-GKPi-xHXebMs8BX5UIbVTdM0W1YrSLRi8APt_6j-wnOdti3Q-mmDugmK4zH4mXB0D7ENrwqRjqRPg8ywTeYNZxoI-krz8y-yjuwrSWWK1Mr9HpelUZP5iGO6H8Z74BQ)

</details>

rhettinger:
Stefan, take the `key=key` out of the "Raymond" variant.  Also we really don't care about the empty iterator case, it is a distractor.  (Side note: passing in a `lambda :` is possibly the worst way to use *timeit*. Instead, prefer the `setup="from __main__ import test"` style shown in the examples at the bottom of the *timeit* docs.)

Bo, aside from the two other hacky variants proposed, I don't think there are any wrong choices here.  My order of preference is 1) the `try/next/next` proposed in the issue, 2) the `for first/for second` in the current PR, or 3) just leaving it as-is.  

Though not currently the fastest, the `try/next/next` most clearly expresses the algorithmic intent and doesn't ask the interpreter for anything beyond that.  As the interpreter evolves, `try/next/next` will become the fastest (looking up builtins will be as fast as looking up globals/locals , catching exceptions will become cheaper, can catching StopIteration may be special cased).  That said, we're only talking a few tens of nanoseconds outside the inner loop, so none of this really matters much.  I don't care about it enough to endlessly joust with Stefan.

PR #903, opened 2024-08-20
Reduce groupby.__next__ calls in all_equal

### LINKED ISSUE #902: One more nit on all_equal()
I missed something during the code review.  There is an extra unnecessary `next` call in the first for-loop.  This is easily solved by adding another `return True`.

```
def all_equal(iterable, key=None):
    iterator = groupby(iterable, key)
    for first in iterator:
        for second in iterator:
            return False
        return True                          # <== This was missed
    return True
```

For example, `all_equal('a')` calls the `groupby.__next__` three times.

This is can be made visible by replacing `groupby` with `verbose_groupby`:

```
class verbose_groupby(itertools.groupby):

    def __next__(self):
        print('N')
        return super().__next__()
```

bbayles:
This PR implements @rhettinger's suggestion in issue 902.

Closes #902

### DIFF
--- a/tests/test_recipes.py
+++ b/tests/test_recipes.py
@@ -2,11 +2,12 @@ from decimal import Decimal
 from doctest import DocTestSuite
 from fractions import Fraction
 from functools import reduce
-from itertools import combinations, count, permutations
+from itertools import combinations, count, groupby, permutations
 from operator import mul
 from math import factorial
 from sys import version_info
 from unittest import TestCase, skipIf
+from unittest.mock import patch
 
 import more_itertools as mi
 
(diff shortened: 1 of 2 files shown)


tests/test_recipes.py
```diff
@@ -158,6 +159,22 @@ def test_key(self):
         self.assertTrue(mi.all_equal('4٤໔４৪', key=int))
         self.assertFalse(mi.all_equal('Abc', key=str.casefold))
 
+    @patch('more_itertools.recipes.groupby', autospec=True)
```
bbayles, line 162: It feels like the mock object's `call_count` should be able to be used here, but I couldn't get it to actually work.

rhettinger (approved):

PR #917, opened 2024-10-10
Issue 916: lt only for is_sorted

### LINKED ISSUE #916: is_sorted() should only depend on less-than comparisons
The `sorted()` function is [documented to only use less-than comparisons](https://docs.python.org/3/library/functions.html#sorted).  The `is_sorted()` predicate should do the same:

```
def is_sorted(iterable, key=None, reverse=False, strict=False):
    it = iterable if (key is None) else map(key, iterable)
    a, b = tee(it)
    next(b, None)
    if reverse:
        b, a = a, b
    return all(map(lt, a, b)) if strict else not any(map(lt, b, a))
```

Note:  This breaks some of the tests with *NaN* values when `strict=True`.  However, I think all of the *NaN* tests are dubious.

For example, `math.nan <= 0 <= math.nan <= 0` is `False` but `is_sorted(['nan', 0, 'nan', 0], key=float)` currently returns `True`.

In general, it doesn't make sense to claim a dataset containing a *NaN* is sorted.  When `data = [8.0, 4.0, 0.0, math.nan, 6.0, 2.0]`, the call `is_sorted(sorted(data))` nonsensically returns `False`.

bbayles:
This PR updates `is_sorted` to use only `lt` comparisons.

Closes #916

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -25,7 +25,7 @@ from itertools import (
 from math import comb, e, exp, factorial, floor, fsum, log, log1p, perm, tau
 from queue import Empty, Queue
 from random import random, randrange, shuffle, uniform
-from operator import itemgetter, mul, sub, gt, lt, le
+from operator import itemgetter, mul, sub, gt, lt
 from sys import hexversion, maxsize
 from time import monotonic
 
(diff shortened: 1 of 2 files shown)


rhettinger:
May I suggest adding a test, something like this:

```
class IntLT:
    "Only implement the __lt__ rich comparison."
    def __init__(self, x):
        self.x = x
    def __lt__(self, other):
        return self.x < other.x

# Verify that is_sorted() works when only less-than is defined
data = list(map(IntLT, range(100)))
for reverse, strict in product([False, True], repeat=2):
    random.shuffle(data)
    ordered = sorted(data, reverse=reverse)
    assert is_sorted(ordered, reverse=reverse, strict=strict)
```

PR #922, opened 2024-12-02
Issue 920: Add is_prime, nth_prime

### LINKED ISSUE #920: Add is_prime()
Consider adding an `is_prime()` function. It is not an iterator but does fit nicely with the other math tools.

Here's some code (tested) that gets the job done and that has fast paths for smaller inputs:

```
from random import randrange
from functools import lru_cache

# https://en.wikipedia.org/wiki/Miller–Rabin_primality_test#Testing_against_small_sets_of_bases
perfect_tests = [
    (2047, (2,)),
    (9080191, (31, 73)),
    (4_759_123_141, (2, 7, 61)),
    (1122004669633, (2, 13, 23, 1662803)),
    (2152302898747, (2, 3, 5, 7, 11)),
    (3474749660383, (2, 3, 5, 7, 11, 13)),
    (341550071728321, (2, 3, 5, 7, 11, 13, 17)),
    (3825123056546413051, (2, 3, 5, 7, 11, 13, 17, 19, 23)),
    (318665857834031151167461, (2, 3, 5, 7, 11, 13, 17, 19, 23, 29, 31, 37)),
    (3317044064679887385961981, (2, 3, 5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41)),
]

@lru_cache
def shift_to_odd(n):
    'Return s, d such that 2**s * d == n'
    s = ((n - 1) ^ n).bit_length() - 1
    d = n >> s
    assert (1 << s) * d == n and d & 1
    return s, d

def strong_probable_prime(n, base):
    assert n > 2 and n & 1
    s, d = shift_to_odd(n - 1)
    x = pow(base, d, n)
    if x == 1 or x == n - 1:
        return True
    for _ in range(s - 1):
        x = x * x % n
        if x == n - 1:
            return True
    return False

def is_prime(n):
    if n < 17:
        return n in {2, 3, 5, 7, 11, 13}
    if not (n&1 and n%3 and n%5 and n%7 and n%11 and n%13):
        return False
    for limit, bases in perfect_tests:
        if n < limit:
            break
    else:
        bases = [randrange(2, n - 1) for i in range(32)]
    return all(strong_probable_prime(n, base) for base in bases)
```

bbayles:
This PR adds the functions discussed in #920:
* `is_prime`
* `nth_prime`

I'm putting these in `recipes.py` based on [this comment](https://github.com/more-itertools/more-itertools/issues/920#issuecomment-2512673450).

Closes #920


rhettinger (approved):

rhettinger:
I think `nth_prime()` should go into `more.py` rather than `recipes.py`.  Otherwise, this looks good.

more_itertools/recipes.py
```diff
@@ -1075,3 +1077,89 @@ def totient(n):
...
+    """Return ``True`` if *n* is prime and ``False`` otherwise.
+
+    >> is_prime(37)
+    True
+    >> is_prime(39)
+    False
+    >> is_prime(18_446_744_073_709_551_557)
+    True
+
+    This function uses the Miller-Rabin primality test, which can return
```
rhettinger, line 1133: Consider making a more precise statement. The test is exact for `n < 3.3 * 10**24`. For larger values, the test is probabilistic with a false positive rate no worse than `1 in 2 ** 64` and with no false negatives.

For higher reliability, the test can be run multiple times.

more_itertools/recipes.py
```diff
@@ -1075,3 +1077,89 @@ def totient(n):
     for prime in set(factor(n)):
         n -= n // prime
     return n
+
+
+# Miller–Rabin primality test: Testing against small sets of bases
+# https://w.wiki/CGwr
```
rhettinger, line 1083: For the runs of small primes (most of the entries in the table), a more authoritative source is https://oeis.org/A014233

more_itertools/recipes.py
```diff
@@ -1075,3 +1077,89 @@ def totient(n):
...
+    ),
+]
+
+
+@lru_cache
+def _shift_to_odd(n):
+    'Return s, d such that 2**s * d == n'
+    s = ((n - 1) ^ n).bit_length() - 1
+    d = n >> s
+    assert (1 << s) * d == n and d & 1
```
rhettinger, line 1106: Let's strengthen this to match the full post condition:

`assert (1 << s) * d == n and d & 1 and s >= 0`

more_itertools/recipes.py
```diff
@@ -1075,3 +1077,89 @@ def totient(n):
...
+    return s, d
+
+
+def _strong_probable_prime(n, base):
+    assert n > 2 and n & 1
+    s, d = _shift_to_odd(n - 1)
```
rhettinger, line 1112: Also add another precondition:

`assert 2 <= base < n`

This will detect cases where the `_perfect_tests` table is misordered. For example, if the `(2047, (2,))` entry is omitted, `is_prime(31)` would give an incorrect result.

more_itertools/recipes.py
```diff
@@ -1075,3 +1077,89 @@ def totient(n):
...
         n -= n // prime
     return n
+
+
+# Miller–Rabin primality test: Testing against small sets of bases
+# https://w.wiki/CGwr
+_perfect_tests = [
+    (2047, (2,)),
+    (9080191, (31, 73)),
+    (4_759_123_141, (2, 7, 61)),
```
rhettinger, line 1087: We should be consistent. Either use underscores in all of the large numbers of none of them.  I slight favor none-of-them.

more_itertools/recipes.py
```diff
@@ -1075,3 +1077,89 @@ def totient(n):
...
+            return True
+    return False
+
+
+def is_prime(n):
+    """Return ``True`` if *n* is prime and ``False`` otherwise.
+
+    >> is_prime(37)
+    True
+    >> is_prime(39)
```
rhettinger, line 1128: Perhaps write `39` as `3 * 13` to make it clear that this is composite.

I don't want to insult the intelligence of the reader but you know how it goes: https://www.gdargaud.net/Humor/OddPrime.html

more_itertools/recipes.py
```diff
@@ -1075,3 +1077,89 @@ def totient(n):
...
+# https://w.wiki/CGwr
+_perfect_tests = [
+    (2047, (2,)),
+    (9080191, (31, 73)),
+    (4_759_123_141, (2, 7, 61)),
+    (1122004669633, (2, 13, 23, 1662803)),
+    (2152302898747, (2, 3, 5, 7, 11)),
+    (3474749660383, (2, 3, 5, 7, 11, 13)),
+    (341550071728321, (2, 3, 5, 7, 11, 13, 17)),
+    (3825123056546413051, (2, 3, 5, 7, 11, 13, 17, 19, 23)),
```
rhettinger, line 1092: Perhaps replace this one with:

`(18446744073709551616, (2, 325, 9375, 28178, 450775, 9780504, 1795265022)),`

That covers all of `2 ** 64` with only seven rounds.

more_itertools/recipes.py
```diff
@@ -1075,3 +1076,76 @@ def totient(n):
...
+    """Return ``True`` if *n* is prime and ``False`` otherwise.
+
+    >> is_prime(37)
+    True
+    >> is_prime(3 * 13)
+    False
+    >> is_prime(18_446_744_073_709_551_557)
+    True
+
+    This function uses the Miller-Rabin primality test, which can return falae
```
rhettinger, line 1136: `falae` -> `false`

more_itertools/more.py
```diff
@@ -4977,3 +4980,23 @@ def doublestarmap(func, iterable):
     """
     for item in iterable:
         yield func(**item)
+
+
+def _nth_prime_ub(n: int) -> float:
+    "Upper bound for the nth prime (counting from 1)."
+    # https://en.wikipedia.org/wiki/Prime-counting_function#Inequalities
+    return n * math.log(n * math.log(n)) if n >= 6 else 11.1
```
rhettinger, line 4988: There is a `from math import log` on line 25, so the `math.log()` can be replaced with just `log()` which  looks nicer.

more_itertools/recipes.py
```diff
@@ -1075,3 +1076,76 @@ def totient(n):
...
+
+# Miller–Rabin primality test: https://oeis.org/A014233
+_perfect_tests = [
+    (2047, (2,)),
+    (9080191, (31, 73)),
+    (4759123141, (2, 7, 61)),
+    (1122004669633, (2, 13, 23, 1662803)),
+    (2152302898747, (2, 3, 5, 7, 11)),
+    (3474749660383, (2, 3, 5, 7, 11, 13)),
+    (341550071728321, (2, 3, 5, 7, 11, 13, 17)),
```
rhettinger, line 1089: Suggest removing lines 1089 and 1090 since the new `2**64` bases now give more coverage for just seven rounds.

more_itertools/recipes.py
```diff
@@ -1075,3 +1076,76 @@ def totient(n):
...
+    s = ((n - 1) ^ n).bit_length() - 1
+    d = n >> s
+    assert (1 << s) * d == n and d & 1 and s >= 0
+    return s, d
+
+
+def _strong_probable_prime(n, base):
+    assert n > 2 and n & 1
+
+    s, d = _shift_to_odd(n - 1)
```
rhettinger, line 1111: Minor nit:  I would swap these two lines (1111 and 1112) so that the two precondition checks are at the top before the processing begins.

more_itertools/recipes.py
```diff
@@ -1075,3 +1076,76 @@ def totient(n):
...
+    (18446744073709551616, (2, 325, 9375, 28178, 450775, 9780504, 1795265022)),
+    (
+        3317044064679887385961981,
+        (2, 3, 5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41),
+    ),
+]
+
+
+@lru_cache
+def _shift_to_odd(n):
```
rhettinger, line 1100: Is it worth adding annotations to the helper functions so that the type checker can fully check the main function?

`def _shift_to_odd(n: int) -> tuple[int, int]:`

more_itertools/recipes.py
```diff
@@ -1075,3 +1076,76 @@ def totient(n):
...
+    'Return s, d such that 2**s * d == n'
+    s = ((n - 1) ^ n).bit_length() - 1
+    d = n >> s
+    assert (1 << s) * d == n and d & 1 and s >= 0
+    return s, d
+
+
+def _strong_probable_prime(n, base):
```
rhettinger, line 1108: Perhaps annotate this one as well:

`def _strong_probable_prime(n: int, base: int) -> bool:`

rhettinger:
This looks excellent. Thank you.

rhettinger (approved):

PR #926, opened 2024-12-05
Add loops recipe

### LINKED ISSUE #923: Sync with itertools recipes:  loops()
Recently a new `loops()` recipe was added to the [CPython main docs](https://docs.python.org/3/library/itertools.html#itertools-recipes).

```
def loops(n):
    "Loop n times. Like range(n) but without creating integers."
    # for _ in loops(100): ...
    return repeat(None, n)
```

Simple test:

```
>>> all(list(loops(n)) == [None] * n for n in range(-10, 10))
True
```

It is intended to be used like this:

```
>>> for _ in loops(5):
...     print('hello world')
... 
hello world
hello world
hello world
hello world
hello world
```

bbayles:
This PR closes #923. The `loops` function from the `itertools` docs is added - thanks @rhettinger for the contribution


more_itertools/recipes.py
```diff
@@ -1146,3 +1147,17 @@ def is_prime(n):
     else:
         bases = [randrange(2, n - 1) for i in range(32)]
     return all(_strong_probable_prime(n, base) for base in bases)
+
+
+def loops(n):
+    """Returns an iterable with *n* elements for efficient looping.
+    Like ``range(n)``, but doesn't create integegers.
```
rhettinger, line 1154: `integegers` -> `integers`

rhettinger: Also, I think there shouldn't be a comma because there aren't two independent clauses.

tests/test_recipes.py
```diff
@@ -1370,3 +1370,10 @@ def test_special_primes(self):
         ):
             with self.subTest(n=n):
                 self.assertTrue(mi.is_prime(n))
+
+
+class LoopsTests:
```
rhettinger, line 1375: The should inherit from `TestCase`.

tests/test_recipes.py
```diff
@@ -1370,3 +1370,10 @@ def test_special_primes(self):
...
+    def test_basic(self):
+        self.asserTrue(
```
rhettinger, line 1377: `self.asserTrue` -> `self.assertTrue`

more_itertools/recipes.py
bbayles: I asked a robot about the comma:

```
Prompt:
Here's a bit of documentation:

> Like ``range(n)`` but doesn't create integers.

The copyeditor says: don't put a comma in between these clauses. Is he right?

Answer: 
Yes, the copyeditor is right.

You don't need a comma between those two clauses because they are joined by a conjunction ("but").

```

You're right! But just as a check...

```
Prompt:
Here's a bit of documentation:

> Like ``range(n)`` but doesn't create integers.
The copyeditor says: put a comma in between these clauses. Is he right?

Answer:
The copyeditor is correct.

The sentence "Like range(n) but doesn't create integers" is a compound sentence.  Compound sentences have two independent clauses (meaning each could stand alone as a sentence).
```

...thanks robot.

rhettinger (approved):

PR #931, opened 2024-12-19
Set number of rounds for is_prime to 64

### LINKED ISSUE #930: More rounds for is_prime() ?
The paper [Primality Testing Under Adversarial Conditions](https://eprint.iacr.org/2018/749.pdf) recommends increasing the number of Miller-Rabin rounds to 64.

They note that in non-adversarial conditions, such as generating random large primes, there is no performance degradation because composites are usually weeded-out in the first few rounds. So the extra work is in certifying an actual prime.

```
diff --git a/more_itertools/recipes.py b/more_itertools/recipes.py
index cf61fc6..b1a7e4e 100644
--- a/more_itertools/recipes.py
+++ b/more_itertools/recipes.py
@@ -1134,8 +1134,7 @@ def is_prime(n):
     This function uses the Miller-Rabin primality test, which can return false
     positives for very large inputs. For values of *n* below 10**24
     there are no false positives. For larger values, there is less than
-    a 1 in 2**64 false positive rate. Multiple tests can reduce the chance
-    of a false positive.
+    a 1 in 2**128 false positive rate.
     """
     if n < 17:
         return n in {2, 3, 5, 7, 11, 13}
@@ -1145,7 +1144,7 @@ def is_prime(n):
         if n < limit:
             break
     else:
-        bases = [randrange(2, n - 1) for i in range(32)]
+        bases = [randrange(2, n - 1) for i in range(64)]
     return all(_strong_probable_prime(n, base) for base in bases)

```

bbayles:
This PR closes #930 by doubling the number of rounds for `is_prime`.

Thanks to @rhettinger for the patch.

### DIFF
--- a/more_itertools/recipes.py
+++ b/more_itertools/recipes.py
@@ -1134,7 +1134,7 @@ def is_prime(n):
     This function uses the Miller-Rabin primality test, which can return false
     positives for very large inputs. For values of *n* below 10**24
     there are no false positives. For larger values, there is less than
-    a 1 in 2**64 false positive rate. Multiple tests can reduce the chance
+    a 1 in 2**128 false positive rate. Multiple tests can reduce the chance
     of a false positive.
     """
     if n < 17:
@@ -1145,7 +1145,7 @@ def is_prime(n):
         if n < limit:
             break
     else:
-        bases = [randrange(2, n - 1) for i in range(32)]
+        bases = [randrange(2, n - 1) for i in range(64)]
     return all(_strong_probable_prime(n, base) for base in bases)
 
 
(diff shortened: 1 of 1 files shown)


more_itertools/recipes.py
```diff
@@ -1134,7 +1134,7 @@ def is_prime(n):
     This function uses the Miller-Rabin primality test, which can return false
     positives for very large inputs. For values of *n* below 10**24
     there are no false positives. For larger values, there is less than
-    a 1 in 2**64 false positive rate. Multiple tests can reduce the chance
+    a 1 in 2**128 false positive rate. Multiple tests can reduce the chance
```
rhettinger, line 1137: Perhaps while we're here, change "can" to "can further".   One in 2**128 is already so rare that it borders on "impossible".

rhettinger (approved):

Issue #937, opened 2024-12-27
Include derangements


debruijn:
### Description
I'd like to propose to add `derangements` (and a few variants) to `more_itertools`. [Derangements](https://en.wikipedia.org/wiki/Derangement) of x are the subset of all permutations of x for which no element has index equal to its value (e.g. exclude (0, 2, 1) since 0 is at index-0 but include (1, 2, 0) since no element has index equal to its value).

More detailed, my proposal is to add three functions:
- `derangements(iterable, r)` that gets all derangements incl duplicates, like `permutations(iterable, r)`
- `distinct_derangements(iterable, r)` that removes duplicates, analog to `distinct_permutations`'s relation to `permutations`
- `derangements_range(n)` that deals with the specific common case of generating derangements from a range, which allows for a faster implementation

### References
Derangements can come up in a variety of applications, such as finding all ways to do Secret Santa without individuals getting to buy a present for themselves (which was the motivation for me thinking about this over Christmas due to a Standupmaths video) or all ways to let students correct each others tests. This can be generalized to use cases with duplicate numbers in the input, such as assigning a batch of Pull Requests to reviewers not reviewing their own PR.

### Example code
The core idea can be coded by filtering out from all permutations:
```
def derangements(iterable, r=None):
  for p in permutations(iterable, r=r):
      if any(x == i for i, x in enumerate(p)): continue
      yield p
```

### Example output
```
>>> sorted(derangements_range(4))
[(1, 0, 3, 2), (1, 2, 3, 0), (1, 3, 0, 2), (2, 0, 3, 1), (2, 3, 0, 1), (2, 3, 1, 0), (3, 0, 1, 2), (3, 2, 0, 1), (3, 2, 1, 0)]
>>> sorted(derangements(range(4), r=3))
[(1, 0, 3), (1, 2, 0), (1, 2, 3), (1, 3, 0), (2, 0, 1), (2, 0, 3), (2, 3, 0), (2, 3, 1), (3, 0, 1), (3, 2, 0), (3, 2, 1)]

>>> sorted(derangements([0, 0, 1, 2]))
[(1, 0, 0, 2), (1, 0, 0, 2), (1, 2, 0, 0), (1, 2, 0, 0), (2, 0, 0, 1), (2, 0, 0, 1), (2, 0, 1, 0), (2, 0, 1, 0)]
>>> sorted(distinct_derangements([0, 0, 1, 2]))
[...]

PR #938, opened 2024-12-27
Add derangements

debruijn:
### Issue reference
#937 

### Changes
Include `derangements`, `derangements_range` and `distinct_derangements` into `more_itertools` as presented in the linked issue.

Some explanations:
- I have tried to make this as similar to `distinct_permutations` as possible regarding usage and documentation. Because of this, I have put it close to that function in `more.py`.
- For `derangements_range(n)` there is an implementation that makes use of knowing the input to avoid testing each element, and achieves a runtime of about 1/4th to 1/3rd of `derangements(range(n))`.
  - This is done using a Dynamic Programming solution, that uses the output of `n-1` and `n-2`:
  - `n-1` is extended with this new element by putting it on each position with the displaced element put at the back which is always possible
  - The other solutions are all those in which the new element swaps out with a single element in the `n-1` case that violates the restriction. To construct these, use an adjusted `n-2` solution and extend it by swapping in the two missing elements.
- For `derangements` and `distinct_derangements` the current implementation is simply filtering out non-deranged elements of `permutations` or `distinct_permutations`. In this case, the benefit for the user is in convenience and clarity. 
  - Further speed-up of these two might be possible as well. Initial exploration of this ended up with an algorithm that was about as fast as this filtering solution but way more convoluted, but I might give that another try at some point. (Unless someone else beat me to it)

### Checks and tests
Done, all pass.


bbayles:
Thanks for the suggestion. I'm good with `derangements` and `distinct_derangements`, but the `range` variant is probably out of scope for this library. If you can update the PR with those two in it, I'll include them in the next release.

debruijn:
Thank you! I have removed `derangements_range`. 

For my understanding: is it out of scope because it doesn't take an iterable as input? If so, I think that makes sense, and I might look for a combinatorics package to include it in (or make one myself).

PR #940, opened 2024-12-31
Issue 939: Speed up factor

### LINKED ISSUE #939: Clean-up and speed-up factor()
Moves the small prime reduction before the Pollard's Rho step. This makes the code a faster and partially restores lazy evaluation.  

Uses a precomputed table of small primes instead of recurring calls to `sieve()`.  This makes it cost effective to run trial divisions up to 211 so that no further reductions are needed for factors below 211².  This reduces the number of calls to `is_prime()` and `pollard()`.

The cutoff of 211 was found empirically to give the best performance. That is the point where O(∜n) Pollard's Rho overtakes  O(√n) trial division.  Cutoffs from 127 to 211.

Eliminates the dependency on `sieve()` and removes the `_factor_trial()` helper function in favor of in-line code.

**New code:**

```python
def factor(n):
    """Yield the prime factors of n in sorted order.

    >>> list(factor(360))
    [2, 2, 2, 3, 3, 5]

    Finds small factors with trial division and
    large factors with Pollard's rho algorithm.
    """

    small_primes = (2, 3, 5, 7, 11, 13, 17, 19, 23, 29, 31, 37,
        41, 43, 47, 53, 59, 61, 67, 71, 73, 79, 83, 89, 97, 101,
        103, 107, 109, 113, 127, 131, 137, 139, 149, 151, 157,
        163, 167, 173, 179, 181, 191, 193, 197, 199)  # sieve(211)

    trial_division_boundary = 211 ** 2

    # Corner case reduction
    if n < 1:
        return

    # Trial division reduction
    for prime in small_primes:
        while not n % prime:
            yield prime
            n //= prime

    # Pollard's rho reduction
    primes = []
    todo = [n] if n > 1 else []
    for n in todo:
        if n < trial_division_boundary or is_prime(n):
            primes.append(n)
        else:
            fact = _factor_pollard(n)
            todo += (fact, n // fact)
    yield from sorted(primes)
```

**Test and timing:**

```python
if __name__ == '__main__':

    from more_itertools import consume, factor as factor_baseline
    from time import perf_counter as time

    for factor_function in factor, factor_baseline:
        start = time()
        for n in range(10 ** 7):
            consume(factor_function(n))
        print(time() - start)
    print()
```

**Timing results:**

* New factor function:  31.3 seconds
* Old factor function:  65.1 seconds

bbayles:
This PR closes #939. The improvements are given in the issue by @rhettinger - many thanks.

### DIFF
--- a/more_itertools/recipes.py
+++ b/more_itertools/recipes.py
@@ -1045,7 +1045,8 @@ def factor(n):
     This function uses trial division for when *n* is less than 1000 and
     Pollard's rho algorithm for larger inputs.
     """
-    small_primes = (
+    # tuple(sieve(211))
+    small_primes
         2,
         3,
         5,
--- a/more_itertools/recipes.py
+++ b/more_itertools/recipes.py
@@ -1045,18 +1045,77 @@ def factor(n):
     This function uses trial division for when *n* is less than 1000 and
     Pollard's rho algorithm for larger inputs.
     """
+    small_primes = (
+        2,
+        3,
+        5,
+        7,
+        11,
+        13,
+        17,
+        19,
+        23,
+        29,
+        31,
+        37,
+        41,
+        43,
+        47,
+        53,
+        59,
+        61,
+        67,
+        71,
+        73,
+        79,
+        83,
+        89,
+        97,
+        101,
+        103,
+        107,
+        109,
+        113,
+        127,
+        131,
+        137,
+        139,
+        149,
+        151,
+        157,
+        163,
+        167,
+        173,
+        179,
+        181,
+        191,
+        193,
+        197,
+        199,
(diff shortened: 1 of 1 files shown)


more_itertools/recipes.py
```diff
@@ -1045,18 +1045,77 @@ def factor(n):
...
+        163,
+        167,
+        173,
+        179,
+        181,
+        191,
+        193,
+        197,
+        199,
+    )
```
rhettinger, line 1095: Perhaps add  this comment: `# tuple(sieve(211))`. That will help mentally connect the small primes list to the calculation that follows.

rhettinger (approved):

PR #934, opened 2024-12-20
Add ruff

stankudrow:
### Issue reference

Addresses the issue #933 .
Must follow the PR #942 .
Allows to close the PR #852 .

### Changes
Tons of them anout formatting and linting the code according to the configuration (which is discussable) from the `pyproject.toml` file:
- docs 
- more_itertools
- tests

### Checks and tests
- `pytest .` - no changes in code and tests apart refactoring
- `uv run --with ruff ruff check --fix .` - no error code reported

Screenshot from 2024-12-20 16-51-01

UPD: warning gone (the [comment](https://github.com/more-itertools/more-itertools/pull/934#issuecomment-2580878173)).

**Notes**:
- adopting the `ruff` tool can make the `black` formatting package redundant;
- the same about flake8 and many other linters.

**Updates**.

Some changes cover a total elimination of Py3.8 support ( I should have run `pyupgrade py39-plus ...` in the #928 ).

PR #947, opened 2025-01-17
Move from flake8 and black to ruff

### LINKED ISSUE #933: Add ruff formatter/linter
The [ruff](https://docs.astral.sh/ruff/) package can be added as an optional dependency which can be useful for keeping the code clean and meeting PEP conventions.

bbayles:
This PR switches from `flake8` and `black` to `ruff`. It is based on @stankudrow's code in #934.

Closes #933


MarcinKonowalczyk:
Could we also remove the [flake8] section in setup.cfg? That's the only leftover of flake8 as far as my grep through the repo goes.

stankudrow:
Finally the gitignore repo got this merged: https://github.com/github/gitignore/pull/4289 .

tests/test_more.py
```diff
@@ -1915,7 +1915,7 @@ def test_deprecation(self):
             )
 
         (warning,) = caught
-        assert warning.category == DeprecationWarning
+        self.assertTrue(warning.category is DeprecationWarning)
```
rhettinger, line 1918: Consider using [`assertIs`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertIs) for this line or simplify the entire test with [`assertWarns`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertWarns).

PR #948, opened 2025-01-19
Add minor ruff-oriented fixes.

stankudrow:
Continues the following pull requests:

- #947 (add ruff formatter and linter)

This PR aims the goals as:

- enrich the project configuration concerning `ruff` (https://docs.astral.sh/ruff/configuration/) - add as many rules as it seems enough for this PR;
- add more `ruff` rules (the default set is `select = ["E4", "E7", "E9", "F"]` - not sufficient IMO);
- lint the code according to these rules (this may not exclude the refactoring which is not induced by ruff);
- keep this PR as small as possible.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+TARGET_DIRS := more_itertools tests
+
 .PHONY: all-checks
 all-checks: requirements coverage check docs package
 
@@ -8,18 +10,23 @@ requirements:
 
 .PHONY: check
 check:
-	ruff format --check .
-	ruff check more_itertools tests
+	ruff format --check ${TARGET_DIRS}
+	ruff check ${TARGET_DIRS}
 	stubtest more_itertools.more more_itertools.recipes
 
 .PHONY: format
 format:
-	ruff format .
+	ruff format ${TARGET_DIRS}
 
 .PHONY: coverage
 coverage:
 	coverage run --include="more_itertools/*.py" -m unittest
 	coverage report --show-missing --fail-under=99
+
+.PHONY: lint
+lint:
+	ruff format ${TARGET_DIRS}
+	ruff check --fix ${TARGET_DIRS}
 
 .PHONY: test
 test:
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -3312,8 +3312,8 @@ def set_partitions(iterable, k=None, min_size=None, max_size=None):
     ['a', 'b', 'c']
 
     """
-    L = list(iterable)
-    n = len(L)
+    lst = list(iterable)
+    n = len(lst)
     if k is not None:
         if k < 1:
             raise ValueError(
@@ -3334,10 +3334,10 @@ def set_partitions(iterable, k=None, min_size=None, max_size=None):
         elif n == k:
             yield [[s] for s in L]
         else:
-            e, *M = L
-            for p in set_partitions_helper(M, k - 1):
+            e, *m = L
+            for p in set_partitions_helper(m, k - 1):
                 yield [[e], *p]
-            for p in set_partitions_helper(M, k):
+            for p in set_partitions_helper(m, k):
                 for i in range(len(p)):
                     yield p[:i] + [[e] + p[i]] + p[i + 1 :]
 
@@ -3345,12 +3345,12 @@ def set_partitions(iterable, k=None, min_size=None, max_size=None):
         for k in range(1, n + 1):
             yield from filter(
                 lambda z: all(min_size <= len(bk) <= max_size for bk in z),
-                set_partitions_helper(L, k),
+                set_partitions_helper(lst, k),
             )
     else:
         yield from filter(
             lambda z: all(min_size <= len(bk) <= max_size for bk in z),
-            set_partitions_helper(L, k),
+            set_partitions_helper(lst, k),
         )
 
 
@@ -3649,12 +3649,12 @@ def _sample_unweighted(iterator, k, strict):
     reservoir = list(islice(iterator, k))
     if strict and len(reservoir) < k:
         raise ValueError('Sample larger than population')
-    W = 1.0
+    w = 1.0
 
     with suppress(StopIteration):
         while True:
-            W *= exp(log(random()) / k)
-            skip = floor(log(random()) / log1p(-W))
+            w *= exp(log(random()) / k)
+            skip = floor(log(random()) / log1p(-w))
             element = next(islice(iterator, skip, None))
             reservoir[randrange(k)] = element
 
@@ -3725,10 +3725,10 @@ def _sample_counted(population, k, counts, strict):
         raise ValueError('Sample larger than population')
 
     with suppress(StopIteration):
-        W = 1.0
+        w = 1.0
         while True:
-            W *= exp(log(random()) / k)
-            skip = floor(log(random()) / log1p(-W))
+            w *= exp(log(random()) / k)
+            skip = floor(log(random()) / log1p(-w))
             element = feed(skip)
             reservoir[randrange(k)] = element
 
@@ -4932,10 +4932,10 @@ def dft(xarr):
 
     See :func:`idft` for the inverse Discrete Fourier Transform.
     """
-    N = len(xarr)
-    roots_of_unity = [e ** (n / N * tau * -1j) for n in range(N)]
-    for k in range(N):
-        coeffs = [roots_of_unity[k * n % N] for n in range(N)]
+    l = len(xarr)
+    roots_of_unity = [e ** (n / l * tau * -1j) for n in range(l)]
+    for k in range(l):
+        coeffs = [roots_of_unity[k * m % l] for m in range(l)]
         yield _complex_sumprod(xarr, coeffs)
 
 
@@ -4952,11 +4952,11 @@ def idft(Xarr):
 
     See :func:`dft` for the Discrete Fourier Transform.
     """
-    N = len(Xarr)
-    roots_of_unity = [e ** (n / N * tau * 1j) for n in range(N)]
-    for k in range(N):
-        coeffs = [roots_of_unity[k * n % N] for n in range(N)]
-        yield _complex_sumprod(Xarr, coeffs) / N
+    l = len(Xarr)
+    roots_of_unity = [e ** (n / l * tau * 1j) for n in range(l)]
+    for k in range(l):
+        coeffs = [roots_of_unity[k * m % l] for m in range(l)]
+        yield _complex_sumprod(Xarr, coeffs) / l
 
 
 def doublestarmap(func, iterable):
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -2154,6 +2154,7 @@ class numeric_range(abc.Sequence, abc.Hashable):
 
     def __init__(self, *args):
         argc = len(args)
+        cls_name = self.__class__.__name__
         if argc == 1:
             (self._stop,) = args
             self._start = type(self._stop)(0)
@@ -2165,16 +2166,16 @@ class numeric_range(abc.Sequence, abc.Hashable):
             self._start, self._stop, self._step = args
         elif argc == 0:
             raise TypeError(
-                f'numeric_range expected at least 1 argument, got {argc}'
+                f'{cls_name} expected at least 1 argument, got {argc}'
             )
         else:
             raise TypeError(
-                f'numeric_range expected at most 3 arguments, got {argc}'
+                f'{cls_name} expected at most 3 arguments, got {argc}'
             )
 
         self._zero = type(self._step)(0)
         if self._step == self._zero:
-            raise ValueError('numeric_range() arg 3 must not be zero')
+            raise ValueError('{cls_name}() arg 3 must not be zero')
         self._growing = self._step > self._zero
 
     def __bool__(self):
@@ -2272,11 +2273,10 @@ class numeric_range(abc.Sequence, abc.Hashable):
         return numeric_range, (self._start, self._stop, self._step)
 
     def __repr__(self):
+        cls_name = self.__class__.__name__
         if self._step == 1:
-            return f"numeric_range({self._start!r}, {self._stop!r})"
-        return (
-            f"numeric_range({self._start!r}, {self._stop!r}, {self._step!r})"
-        )
+            return f"{cls_name}({self._start!r}, {self._stop!r})"
+        return f"{cls_name}({self._start!r}, {self._stop!r}, {self._step!r})"
 
     def __reversed__(self):
         return iter(
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -2150,7 +2150,7 @@ class numeric_range(abc.Sequence, abc.Hashable):
 
     """
 
-    _EMPTY_HASH = hash(range(0, 0))
+    _EMPTY_HASH = hash(range(0))
 
     def __init__(self, *args):
         argc = len(args)
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -794,7 +794,6 @@ def distinct_permutations(iterable, r=None):
     if r is None:
         r = size
 
-    # functools.partial(_partial, ... )
     algorithm = _full if (r == size) else partial(_partial, r=r)
 
     if 0 < r <= size:
@@ -1211,7 +1210,7 @@ def interleave_evenly(iterables, lengths=None):
         errors = [e - delta for e, delta in zip(errors, deltas_secondary)]
 
         # those iterables for which the error is negative are yielded
-        # ("diagonal step" in Bresenham)
+        # ("diagonal step" in Bresenham)  # noqa: ERA001
         for i, e_ in enumerate(errors):
             if e_ < 0:
                 yield next(iters_secondary[i])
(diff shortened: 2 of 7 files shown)


Makefile
```diff
@@ -8,19 +10,24 @@ requirements:
...
 
 .PHONY: coverage
 coverage:
 	coverage run --include="more_itertools/*.py" -m unittest
 	coverage report --show-missing --fail-under=99
 
+.PHONY: lint
+lint:
+	ruff format ${TARGET_DIRS}
+	ruff check --fix ${TARGET_DIRS}
```
stankudrow, line 29: This came in handy.

tests/test_recipes.py
```diff
@@ -643,19 +645,19 @@ def test_no_replacement(self):
...
         combo = mi.random_combination_with_replacement(items, len(items) * 2)
         self.assertEqual(2 * len(items), len(combo))
         if len(set(combo)) == len(combo):
             raise AssertionError("Combination contained no duplicates")
 
     def test_pseudorandomness(self):
-        """ensure different subsets of the iterable get returned over many
-        samplings of random combinations"""
+        """Ensure different subsets of the iterable get returned over many
+        samplings of random combinations."""
```
stankudrow, line 660: An artifact left from giving a try to the pydocstyle (D) rules. There is a strikingly great number of them throughout the source code and ignoring test files does not help out.

more_itertools/more.py
```diff
@@ -2152,10 +2150,11 @@ class numeric_range(abc.Sequence, abc.Hashable):
 
     """
 
-    _EMPTY_HASH = hash(range(0, 0))
+    _EMPTY_HASH = hash(range(0))
 
     def __init__(self, *args):
         argc = len(args)
+        cls_name = self.__class__.__name__
```
stankudrow, line 2157: Dragged from the #934 .

pyproject.toml
```diff
@@ -52,7 +52,26 @@ target-version = "py39"
...
+    "ERA",  # eradicate
+    "F",  # pyflakes
+    "I",  # isort
+    "ICN",  # flake8-import-conventions
+    "ISC",  # flake8-implicit-str-concat
+    "N",  # pep8-naming
+    "PIE",  # flake8-pie
+    "T20",  # flake8-(p)print
+    "UP",  # pyupgrade
+]
```
stankudrow, line 67: The default configuration is posted here: https://docs.astral.sh/ruff/configuration/ . As you can see, the default selected set of rules is `select = ["E4", "E7", "E9", "F"]`, which is not enough in my opinion.

pyproject.toml
```diff
@@ -52,7 +52,26 @@ target-version = "py39"
...
+ignore = [
+    "E731",  # lambda-assignment
+    "E741",  # ambiguous-variable-name
+    "N801",  # CapWords convention
+    "N802",  # function name should be lowercase
+    "N803",  # argument name should be lowercase
```
stankudrow, line 73: I think including the PEP8 naming conventions are of high importance, but I had ignore these rules because of multiple complaints from `ruff` and fixing them would pretty tiresome and added tons of changes.

more_itertools/more.py
```diff
@@ -2274,11 +2273,10 @@ def __reduce__(self):
...
 
     def __repr__(self):
+        cls_name = self.__class__.__name__
         if self._step == 1:
-            return f"numeric_range({self._start!r}, {self._stop!r})"
-        return (
-            f"numeric_range({self._start!r}, {self._stop!r}, {self._step!r})"
-        )
+            return f"{cls_name}({self._start!r}, {self._stop!r})"
+        return f"{cls_name}({self._start!r}, {self._stop!r}, {self._step!r})"
```
stankudrow, line 2279: Drawn from the #934 .

tests/test_more.py
```diff
@@ -5395,59 +5396,59 @@ def test_key_partially_hashable(self):
...
         self.assertEqual(output, [1, 2])
-        self.assertEqual(list(mi.duplicates_everseen(input)), output)
+        self.assertEqual(list(mi.duplicates_everseen(input_)), output)
 
     def test_vs_duplicates_everseen_key(self):
-        input = 'HEheHEhe'
+        input_ = 'HEheHEhe'
         output = [
-            e for e, j, u in mi.classify_unique(input, str.lower) if not u
+            e for e, j, u in mi.classify_unique(input_, str.lower) if not u
```
stankudrow, line 5431: For not shadowing the `input` builtin function.

tests/test_more.py
```diff
@@ -5927,16 +5928,16 @@ def test_basic(self):
         # Example calculation from:
         # https://en.wikipedia.org/wiki/Discrete_Fourier_transform#Example
         xarr = [1, 2 - 1j, -1j, -1 + 2j]
-        Xarr = [2, -2 - 2j, -2j, 4 + 4j]
-        self.assertTrue(all(map(cmath.isclose, mi.dft(xarr), Xarr)))
-        self.assertTrue(all(map(cmath.isclose, mi.idft(Xarr), xarr)))
+        xarr2 = [2, -2 - 2j, -2j, 4 + 4j]
```
stankudrow, line 5931: Or maybe `yarr` or something else, please feel free to propose better variants.

tests/test_recipes.py
```diff
@@ -580,10 +582,10 @@ def test_list_with_repeat(self):
 
 
 class RandomPermutationTests(TestCase):
-    """Tests for ``random_permutation()``"""
+    """Tests for ``random_permutation()``."""
 
     def test_full_permutation(self):
-        """ensure every item from the iterable is returned in a new ordering
+        """Ensure every item from the iterable is returned in a new ordering.
```
stankudrow, line 588: Please refer to this comment: https://github.com/more-itertools/more-itertools/pull/948/files#r1921578029 .

rhettinger:
No.  I disagree with this. Some of the ruff "rules" are BS and make the code worse, not better. The purpose of using a tool is to help improve code not to appease to a tool.  In the case of "ruff", there's an obvious pun about the tail wagging the dog.

Variable names like `Xarr`, `W` and `N` were chosen for a reason. They have a strong meaning in the problem domain.

more_itertools/more.py
```diff
@@ -1897,7 +1895,7 @@ def getter(obj):
                 return obj[i]
             except IndexError:
                 # basically if we have an iterable like
-                # iter([(1, 2, 3), (4, 5), (6,)])
+                # iter([(1, 2, 3), (4, 5), (6,)])  # noqa: ERA001
```
rhettinger, line 1898: I object to `ERA001` being enabled. A Google search shows a high false positive rate.  Also, the premise of the rule, "commented-out code is bad" is itself dubious.

We really don't want to garbage up the code with a bug on `noqa`. That defeats the purpose of making the code more readable.

more_itertools/more.py
```diff
@@ -2152,10 +2150,11 @@ class numeric_range(abc.Sequence, abc.Hashable):
 
     """
 
-    _EMPTY_HASH = hash(range(0, 0))
+    _EMPTY_HASH = hash(range(0))
```
rhettinger, line 2153: This edit is okay, but arguably `range(0, 0)` is clearer that this is an empty range.

Mostly, we should defer to the intensions of the person who wrote this could rather than blindly making substitutions based on ruff recommendations which don't actually know the author's intent.

more_itertools/more.py
rhettinger: This group of edits is entirely unnecessary and do not improve the user experience.

It makes instances a little larger and slows instantiation.

If Bo does want to factor-out the class name to make this easily more subclassable (maybe he does, maybe he doesn't), the usual way to get the class name with `to compute `{type(self).__name__}` in the relevant f-strings.

more_itertools/more.py
```diff
@@ -3313,8 +3311,8 @@ def set_partitions(iterable, k=None, min_size=None, max_size=None):
     ['a', 'b', 'c']
 
     """
-    L = list(iterable)
-    n = len(L)
+    lst = list(iterable)
```
rhettinger, line 3314: This is another dubious ruff rule.  `L` and `M` are perfectly valid variable names.

more_itertools/more.py
```diff
@@ -3335,23 +3333,23 @@ def set_partitions_helper(L, k):
         elif n == k:
             yield [[s] for s in L]
         else:
-            e, *M = L
-            for p in set_partitions_helper(M, k - 1):
+            e, *m = L
```
rhettinger, line 3336: Please stop making blind substitutions.  In classical and idiomatic choices of variable names, we use `m` and `n` for populations sizes, `k` and `r` for subpopulation sizes, `p` and `q` for paired variables, `i`, `j`, `k` for indices, etc.

In this case, using `m` makes the code harder to read.  If `M` has to go, then switch to longer names to accurately describe the code.  Perhaps `head` and `tail` for `e` and `M`.

more_itertools/more.py
```diff
@@ -3650,12 +3648,12 @@ def _sample_unweighted(iterator, k, strict):
     reservoir = list(islice(iterator, k))
     if strict and len(reservoir) < k:
         raise ValueError('Sample larger than population')
-    W = 1.0
+    w = 1.0
```
rhettinger, line 3651: NO!  The `W` was chosen for a reason.  See the paper referenced in the comments.  This was another blind substitution make without any understanding of the code or the author's intent.

more_itertools/more.py
```diff
@@ -4953,11 +4951,11 @@ def idft(Xarr):
 
     See :func:`dft` for the Discrete Fourier Transform.
     """
-    N = len(Xarr)
-    roots_of_unity = [e ** (n / N * tau * 1j) for n in range(N)]
-    for k in range(N):
-        coeffs = [roots_of_unity[k * n % N] for n in range(N)]
-        yield _complex_sumprod(Xarr, coeffs) / N
+    l = len(Xarr)
```
rhettinger, line 4954: NO!  The formula here is *always* expressed with a capital `N` and a lowercase `n`'

tests/test_more.py
```diff
@@ -21,15 +20,15 @@
     product,
     repeat,
 )
-from operator import add, mul, itemgetter
-from pickle import loads, dumps
+from operator import add, itemgetter, mul
```
rhettinger, line 23: No one cares about the components of a from-import being alphabetized.

In some ways, this is worse.  The `add` and `mul` are semantically related.

tests/test_more.py
```diff
@@ -890,56 +889,58 @@ def test_reverse(self):
 class BucketTests(TestCase):
     def test_basic(self):
         iterable = [10, 20, 30, 11, 21, 31, 12, 22, 23, 33]
-        D = mi.bucket(iterable, key=lambda x: 10 * (x // 10))
+        bucket = mi.bucket(iterable, key=lambda x: 10 * (x // 10))
```
rhettinger, line 892: NO!  This is uses the same name for the function and its result.  This no better than writing `list = __builtins__.list`.

tests/test_more.py
```diff
@@ -5395,59 +5396,59 @@ def test_key_partially_hashable(self):
         )
 
     def test_vs_unique_everseen(self):
-        input = 'AAAABBBBCCDAABBB'
-        output = [e for e, j, u in mi.classify_unique(input) if u]
+        input_ = 'AAAABBBBCCDAABBB'
```
rhettinger, line 5399: No.  The advice to not shadow builtins is mainly applicable in the global namespace where it is very easy to create confusion.  For local variable names, the advice to not shadow a builtin is only applicable when if the code is easily misread. In this case, it is very clear that `input` is input data and not the `input` builtin function.

Note, `input_` is harder to read. It always slows down someone reading the code.

IMO, this edit makes the code worse.

tests/test_more.py
```diff
@@ -5891,10 +5892,10 @@ def test_hash_count(self):
         hash_count = 0
 
         class Str(str):
-            def __hash__(true_self):
+            def __hash__(self):
```
rhettinger, line 5895: I don't know that the author's intent was here.  Remember that `self` is a convention, not a requirement.  There are a number of occasions where some other name is clearer in a specific context.

The Chesterson's Fence rule suggests deferring to the original author.

tests/test_more.py
rhettinger: No.

tests/test_recipes.py
```diff
@@ -492,26 +492,28 @@ def test_first(self):
     def test_multiple(self):
         """ensure can catch multiple exceptions"""
 
-        class Fiz(Exception):
+        class FizzError(Exception):
```
rhettinger, line 495: No.

In this case, adding `Error` to the name makes the code worse.  The exceptions are used as a named signal rather than an error casing.

rhettinger (changes requested):

rhettinger:
@bbayles I recommend that almost none of these changes be accepted and that the ruff rules that flagged them not be enabled.  For new code, all ruff rules could be checked in an advisory capacity just to make sure that the author hasn't overlooked something but forcing these edits mostly doesn't make sense.

The tools serve you, not the other way around.

bbayles:
Yeah, I appreciate the suggestions here, but I put what I wanted from the original in #947 and left out the rest purposefully.

more_itertools/more.py
stankudrow: The `type(self)`.__name__` was pointed out [here](https://github.com/more-itertools/more-itertools/pull/948#discussion_r1921606445), duly noted.

more_itertools/more.py
stankudrow: Thank you.

tests/test_more.py
stankudrow: I played with the following example:

```python
import array

# If the argument has the name `array`,
# the AttributeError sends the message:
# "'array.array' object has no attribute 'array'"".
def run_example(arr: array.array) -> None:
    print(f"Arg = {arr}")
    # If the `array_` is renamed to `array`,
    # the UnboundLocalError kicks, saying:
    # "cannot access local variable 'array' where it is not associated with a value"
    array_ = array.array("B", [1, 2, 3]) 
    # Hm, this trailing underscore makes the variable name kinda flaky,
    # it draws attention, increasing cognitive load.
    print(f"New array = {array_}")
    print(f"array.array: {array.array("b", [-1])}")

run_example(array.array("f", [42.21]))
```

Could you verify my reasoning about:
- the case about the `array: array.array` parametre -> the AttributeError is raised because the parametre is a local variable and the [LEGB](https://realpython.com/python-scope-legb-rule/) rule is in effect, so, when resolving, the interpreter will stop in the local scope, therefore, the `array.array` import in the outer scope is never reached Am I right?
- the situation about the `UnboundLocalError`. From this section (https://docs.python.org/3/reference/expressions.html#evaluation-order) we know, that "while evaluating an assignment, the right-hand side is evaluated before the left-hand side." And this section (https://docs.python.org/3/faq/programming.html#why-am-i-getting-an-unboundlocalerror-when-the-variable-has-a-value) conveys nice use cases about the `UnboundLocalError`. So, we have the assignment `array = array.array("B", [1, 2, 3])`. If the right-hand side is evaluated first, why it cannot be resolved up to the `array.array` entity? A guess is that the `array` from the left-hand side is already visible in the local scope. Is that correct? Could you possibly what I misunderstand?

Then I tried this example:

```python
import itertools

def run_example(chain):
    print(f"Arg = {chain}")  # OK, the local one -> expected
    chain = itertools.chain(chain)
    print(f"chain = {chain} object")  # still OK -> overwritten purposely

run_example(array.array("f", [42.21]))
```

It worked. I would really appreciate your ideas about this `chain = itertools.chain(chain)` assugnment. Suppose the objection is like "the left-hand `chain` variable is the noun, the right-hand `chain` is the verb and this verb is accessed via the `itertools` module, so no confusion or scope resolution failure occurs at all". What counterexample is viable here?

Raymond, thank you so much, I enjoyed this R&D and it taught me more than one lesson.

tests/test_more.py
stankudrow: My first question was like "Can the self parameter be None?" I wrote the following example (it may be incorrect):

```python
def test_hash_count():
    hash_count = 0

    class Str(str):
        # you can write `self = None`
        def __hash__(self):
            nonlocal hash_count
            hash_count += 1
            return super.__hash__(self)
        
    def test_hash_count():
    hash_count = 0

    class Str(str):
        # you can write `self = None`
        def __hash__(self):
            nonlocal hash_count
            hash_count += 1
            return super.__hash__(self)

    for string in ("", "hello"):
        s = Str(string)
        print(f"s = {s}")
        print(f"hash(s) = {hash(s)} -> count = {hash_count}")
        print(f"Str.__hash__(s) = {Str.__hash__(s)} -> count = {hash_count}")
        print(f"Str.__hash__(None) = {Str.__hash__(None)} -> count = {hash_count}")
        print()  # next line

test_hash_count()
```

So, this example shows that `None` can be passed in lieu of an instance. Honestly, I couldn't find cases when Python can replace self with None, so if you know the answer, I will be happy (no sarcasm or derision) to read it. I cannot get the intent of `true_self` instead of `self`:

- if it's about non-None objects, what is a possible use case?
- if it's about True objects, is that the case?

What am I missing?

PR #960, opened 2025-01-31
Sync _fsumprod() with the CPython implementation.

rhettinger:
The `math.sumprod` function does more than just an `fsum`. It also uses lossless multiplication.

Syncing up the implementation so that both give the same results that are more accurate that the current implementation.  This is especially useful in the context of `dft()` and `idft()` which are both numerically sensitive unless there is a single-rounding `sumprod()`.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -162,7 +162,7 @@ __all__ = [
 try:
     from math import sumprod as _fsumprod
 
-except ImportError:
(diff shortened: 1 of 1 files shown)


bbayles:
The linter is complaining about the switch from `fsum` to `math.fsum`, but once that's fixed I'll take this.

bbayles (approved):

PR #964, opened 2025-02-11
Sync-up recipes with the itertools docs

rhettinger:
Resync the recipes with those in the itertools docs which have had a number of updates over the years.

* Switched to from-imports making the code more readable and sometimes a little faster.
* Get the code to match minor improvement in the published version.
* Add comments where the more-itertools version intentionally differs.


bbayles:
These are all good, but I'd rather not change the parameter names. Can you leave those as-is? I'll change those at the next major release.

PR #979, opened 2025-04-23
Remove setup.py file

cdce8p:
### Changes
The project uses `flit_core` as build backend. Including the `setup.py` file should not be necessary anymore.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,3 +1,0 @@
-from setuptools import setup
-
-setup()


bbayles:
We switched to `flit` to make it easier for packagers (in particular [Gentoo](https://github.com/more-itertools/more-itertools/issues/593)) to build the project. But my understanding at the time (2022) was that other packagers still depended on `setup.py`, so I left it in. Do you have a sense for how that ecosystem has evolved?

cdce8p:
As far as I'm aware all major build frontends, mainly pip, poetry, uv, do support independent build systems now. So it should be safe to remove at this point.

_Anecdotally, I've removed it on a number of other projects some time ago and so far I haven't seen any complains._

PR #985, opened 2025-05-01
Include `derangements` by itself

bbayles:
This PR grabs the commits from #946 and then drops the functions other than `derangements`. I'm still where I was when I made [this comment](https://github.com/more-itertools/more-itertools/issues/937#issuecomment-2587478939), and I think that function would be a nice addition to the library.

@debruijn, are you OK with this? The other function implementations are good, but I think this is the core.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -869,39 +869,12 @@ def derangements(iterable, r=None):
         [(2, 0), (2, 3), (3, 0)]
 
     Elements are treated as unique based on their position, not on their value.
-    If the input elements are unique, there will be no repeated values within a permutation.
+    If the input elements are unique, there will be no repeated values within a
+    permutation.
     """
-    pool = tuple(iterable)
-    xs = tuple(zip(pool))
+    xs = tuple(zip(iterable))
     for ys in permutations(xs, r=r):
-        if any(map(operator.eq, xs, ys)):
-            continue
-        yield tuple(y[0] for y in ys)
-
-
-def distinct_derangements(iterable, r=None):
-    """Yield successive distinct derangements of the elements in *iterable*.
-
-        >>> for d in distinct_derangements(['Alice', 'Bob', 'Carol', 'Alice']):
-        ...    print(', '.join(d))
-        Bob, Alice, Alice, Carol
-        Carol, Alice, Alice, Bob
-
-    Equivalent to yielding from ``set(derangements(iterable))``, except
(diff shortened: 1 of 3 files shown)


pochmann3:
Do I see correctly that this is the current version?

```python
def derangements(iterable, r=None):
    """
    ...
    Elements are treated as unique based on their position, not on their value.
    ...
    """
    pool = tuple(iterable)
    xs = tuple(zip(pool))
    for ys in permutations(xs, r=r):
        if any(map(operator.eq, xs, ys)):
            continue
        yield tuple(y[0] for y in ys)
```

That seems wrong and the combination of zip and eq doesn't make sense to me. Using eq doesn't lead to *"Elements are treated as unique based on their position, not on their value."* And it makes the zip pointless.

The output of
```
for d in derangements([0, False, 1, True]):
    print(d)
```
is:
```
(1, True, 0, False)
(1, True, False, 0)
(True, 1, 0, False)
(True, 1, False, 0)
```
With `operator.is_` as in my original it becomes:
```
(False, 0, True, 1)
(False, 1, True, 0)
(False, True, 0, 1)
(1, 0, True, False)
(1, True, 0, False)
(1, True, False, 0)
(True, 0, False, 1)
(True, 1, 0, False)
(True, 1, False, 0)
```

[Attempt This Online!](https://ato.pxeger.com/run?1=VVFLTsQwDBXbnsLqqpVKNewQ0iyZJSt2oxHKUJeJlNohcdCUq7CZDRyEY3Aakv6g3kR-fvHzsz--bC8npsvlM0h7fftz9d067kALOmE2HnRn2QlYdF0QJZrJZxPGEVTCDrKswRaamNELdkjii9RAHQ1W4LYPTFjeZRAjz_Phret6eO_NyAflEMShEmxAeQikXwPCUfmYM4GcUDuw7HWaoAJi-YPflAlYr_rOOjZ6gC1IsAaXmcqhdPZL4V3bIjHLsdJGS310TivXxdknM25ykkK3oKgvOmWLeRe19k8VJGrvy3_UFM9MoingAvYaTTPN0O83h1E5CcfPWZayJmWrze43FeyU8XG1NxU8uoCHScc6TVI05XjK6aLzZX8B)

bbayles:
Yours is the implementation I was endorsing [here](https://github.com/more-itertools/more-itertools/issues/937#issuecomment-2587478939); I'll change to use it.

PR #988, opened 2025-05-12
Issue 986: argmin and argmax

### LINKED ISSUE #986: argmin() and argmax()
These two functions are common in machine learning.  

[Numpy implements them](https://numpy.org/doc/2.2/reference/generated/numpy.argmin.html) as returning the index of the first occurrence of a minimum or maximum value in an array. For more_itertools, the functions would accept any iterable or iterator and make only a single pass over the data one value at a time.

```
from operator import itemgetter

def argmin(iterable, *, key=None):
    "Index of the first occurence of a minimum value in an iterable."
    if key is not None:
        iterable = map(key, iterable)
    return min(enumerate(iterable), key=itemgetter(1))[0] 

def argmax(iterable, *, key=None):
    "Index of the first occurence of a maximum value in an iterable."
    if key is not None:
        iterable = map(key, iterable)
    return max(enumerate(iterable), key=itemgetter(1))[0] 
```

Some tests:

```
data = [10, 2, 20, 5, 17, 4]
assert argmin(data) == 1
assert argmax(data) == 2

data = [10, -2, -20, 5, 17, 4]
assert argmin(data, key=abs) == 1
assert argmax(data, key=abs) == 2

data = [[0]*10, [0]*5, [0]*3, [0]*12, [0]*2, [0]*3]
assert argmin(data, key=len) == 4
assert argmax(data, key=len) == 3

assert argmin([10, 10, 20, 10]) == 0
assert argmax([30, 30, 20, 30]) == 0
```

bbayles:
Re: https://github.com/more-itertools/more-itertools/issues/986, this PR adds `argmin` and `argmax`. Thanks to @rhettinger for the implementations.

Closes #986


more_itertools/more.py
```diff
@@ -53,6 +53,8 @@
     'all_unique',
     'always_iterable',
     'always_reversible',
+    'argmax',
+    'argmin',
```
rhettinger, line 57: Does this also need to go into `docs/api.rst`?

rhettinger (approved):

rhettinger:
This makes me happy.  Thank you.

PR #993, opened 2025-05-16
Fix the return type of more_itertools.batched

qobilidop:
### Issue reference

This is a trivial bug.

### Changes

Fixes the return type of `more_itertools.batched`. The tuple there should be a tuple of potentially multiple items, rather than a tuple of a single item.

### Checks and tests

`make all-checks` passes.


bbayles:
For reference, here's the annotation for the standard library's version: https://github.com/python/typeshed/blob/0125fe28489590dd341576147b54dc164e1b016a/stdlib/itertools.pyi#L326

PR #1000, opened 2025-06-16
Speed-up _complex_sumprod()

rhettinger:
Replace generator expressions with equivalent calls to `map`.

This runs about 40% faster on my Apple M1 Max


ben42code:
@rhettinger 
While checking the performance improvement, I checked if the results differ between the current and new implementation.
Got a perfect match on my large test case✅

bbayles (approved):

PR #1002, opened 2025-06-17
Add an option for nth_prime() to be approximated

rhettinger:
This makes `nth_prime()` usable with larger values of *n*.

```
% time python3.14 -c 'from more_itertools import nth_prime; nth_prime(100_000_000)' 
14.30s user 0.51s system 99% cpu 14.820 total

% time python3.14 -c 'from more_itertools import nth_prime; nth_prime(100_000_000, approximate=True)'
python3.14 -c   0.03s user 0.01s system 88% cpu 0.046 total
```

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -35,7 +35,9 @@ from .recipes import (
     _zip_equal,
     UnequalIterablesError,
     consume,
+    first_true,
     flatten,
+    is_prime,
     nth,
     powerset,
     sieve,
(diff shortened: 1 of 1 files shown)


more_itertools/more.py
```diff
@@ -5061,21 +5063,45 @@ def doublestarmap(func, iterable):
...
         raise ValueError
+
     limit = math.ceil(_nth_prime_ub(n + 1))
-    return nth(sieve(limit), n)
+
+    if not approximate or n <= 1_000_000:
+        return nth(sieve(limit), n)
+
+    # Round to closest odd within the bounds
+    ub = floor(limit)
```
pochmann3, line 5100: I think `floor` doesn't make sense, as `limit` is a `ceil` value.

pochmann3: I think you misunderstood. What I mean is that `limit` comes from `limit = math.ceil(...)` a few lines above. So what good does `floor` do here?

pochmann3: Actually if `P_n <= 155.3` is intended, then the `math.ceil()` seems just wrong already. Not in that case, but for example for 156.3, which `math.ceil()` would make 157 and then the floor() does nothing and you return 157 which isn't `<= 156.3`.

rhettinger: Okay, that makes sense.

PR #1004, opened 2025-06-18
Fix broken link

rhettinger:


bbayles:
I removed the link when I merged #1008 since I thought it was permanently dead. Glad it's not! If you'd like me to keep the reference I will.

PR #1008, opened 2025-06-18
Issue 1005: Document difference from builtin min() and max().

rhettinger:


more_itertools/more.py
```diff
@@ -4696,6 +4696,16 @@ def minmax(iterable_or_value, *others, key=None, default=_marker):
...
     Otherwise ``ValueError`` is raised.
 
+    This function is intended for finding extreme values in the input.
+    It differs the builtin `min` and `max` functions in several ways:
+
+    1) Unlike ``(min(data), max(data))``, it makes a single pass over the input.
+    2) Comparisons only call `__lt__`, unlike the builtin `max` function
+       which calls ``__gt__`` instead.
+    3) Ties are handled differently. The builtin functions always return the
+       first minimum or maximum value found in the input.  In contrast,
```
pochmann3, line 4706: This function also always returns the first minimum. It's only the max that can be a later one.

(And it can only be the element immediately after the first max and only if they're a considered pair.)

bbayles:
Thanks - I made some edits, but appreciate the clarification on `max` and deleting the dead ActiveState link.

PR #1023, opened 2025-06-28
Add fast path for derangements()

rhettinger:
Add a fast path for the common case where the input elements are distinct.


bbayles:
I took @pochmann3's implementation and @rhettinger's docstring additions. Thanks!

Issue #1032, opened 2025-07-19
Improve typehints for always_iterable


maltevesper:
I would like completions to work when using always iterable, i.e.

```
class MyClass:
    def somemethod(self):
        ...

for x in always_iterable(MyClass()): # I would like the typechecker to know that x is MyClass
    ...

(I am sure there are more functions which could have more accurate typehints, I want to keep this issue small, to provide a PR which matches it)

PR #1033, opened 2025-07-19
Increase always_iterable typing accuracy.

maltevesper:
### Issue reference
closes more-itertools/more-itertools#1032

### Changes
Attempt at increasing the typehint precision for always Iterable.

while the general case is still slightly flaky, I am not sure how to write better typehints in light of the second attribute (on which the return value depends). However, I added several overloads which should help with use without the second argument.

### Checks and tests
 - [x] ran the checks

### DIFF
--- a/more_itertools/more.pyi
+++ b/more_itertools/more.pyi
@@ -452,7 +452,7 @@ def unzip(iterable: Iterable[Sequence[_T]]) -> tuple[Iterator[_T], ...]: ...
 def divide(n: int, iterable: Iterable[_T]) -> list[Iterator[_T]]: ...
 @overload
 def always_iterable(
-    obj: None, base_type: _ClassInfo | None = ...
+    obj: None, base_type: _ClassInfo | None
 ) -> Iterator[Never]: ...
 @overload
 def always_iterable(obj: bytes) -> Iterator[bytes]: ...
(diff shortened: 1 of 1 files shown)


bbayles:
Hmm, I don't understand why you've chosen `bytes` here and not other types?

maltevesper:
> Hmm, I don't understand why you've chosen `bytes` here and not other types?

You are right, I missed `str`. I just added that.

So why did I special case `str` and `bytes`?
This has to do with the [function signature definition](https://github.com/more-itertools/more-itertools/blob/master/more_itertools/more.py#L2061):
`def always_iterable(obj, base_type=(str, bytes)):`

The issue is, that I have not found a way with typehints to reflect the dependency of the return type on `base_type`. So in the general case I only know that the function either returns `Iterable[obj]` or `obj`. But in case `base_type` is not passed in, I can use my knowledge of `base_type` (str, bytes) to type hint those overloads more accurately.

While I am not saying this is perfect, here is a very basic look at what these type hints give me in vscode with pylance:
image

bbayles:
Could you explain why `Sequence[_T]` producing `Iterator[_T]` is needed? The last overload would seem to cover it?

JamesParrott:
@maltevesper  Hi I didn't notice you'd raised this PR, and charged right ahead and took a shot at this too in PR #1050  -sorry for the duplication.  I liked your `Iterator[Never]`.  But I decided to match the `base_type is None` path first, then afterwards handle everything the isinstance could return True from, with a generic type[T]:
```py
@overload
def always_iterable(obj: _T2, base_type: type[_T2])
``` 

I did think using both `Sequence[_T]` together with `Iterable[_T]` could together describe everything that `isinstance` could return True for is passed explicit iterables -i.e. `Sequence` could fill in the gap mentioned in [the docs](https://docs.python.org/3/library/collections.abc.html#collections.abc.Iterable), and "detect classes that iterate with the `__getitem__` method".  But this isn't necessary if None is matched first.  Admittedly more overloads are needed my way.

maltevesper:
@bbayles try removing it and look at what happens to
`x_list_iterable = more_itertools.always_iterable(x_list)`
without the sequence overlaod the single argument overload `def always_iterable(obj: _T) -> Iterator[_T]: ...` is taken, resolving it to `Iterator[List[int]]`.

But as a matter of fact you revealed more issues. The same issue applies to Mappings(dict) and generators.
At first I thought adding mapping and generator to the list would be the way to go, but it turns out that Iterable rather than sequence covers all these cases.

maltevesper:
@bbayles I split the __getitem__ support into its own commit, so that that change could be easily separated.

bbayles:
Superseded by https://github.com/more-itertools/more-itertools/pull/1107. Thanks!

PR #1040, opened 2025-08-01
running_median() with windowed data

rhettinger:
Open questions:

* Possibly rename the `iterable` parameter to `data`, consistent with `statistics.median`? Or keep as-is to match `more_itertools` conventions and to emphasize the lazy evaluation which is the principal use case for `running_median()`?
* Happy with the `O(n)` but mostly fast steps to maintain a sorted window?  Or add more complex `O(log n)` code ([IndexableSkiplist](https://code.activestate.com/recipes/576930-efficient-running-median-using-an-indexable-skipli/), [blist](https://pypi.org/project/blist/), [SortedContainers](https://grantjenks.com/docs/sortedcontainers/), etc)?  Based on Grant Jenks' notes, the current list insort/bisect/del technique can be expected to win for window sizes up to several thousand.

Solved questions:

* Named the size parameter `maxlen` because it caps the size of the window and also allows smaller sizes, like the `maxlen` parameter for deque.
* Let `running_median` start yielding values before the window is full.  This is more convenient to use.  Also, it invariant that the unwindowed case gives the same result as having a window larger than the input data.


more_itertools/recipes.py
```diff
@@ -1383,6 +1365,72 @@ def _running_median_minheap_only(iterable):  # pragma: no cover
...
+    On version Python 3.13 and prior, max-heaps are simulated with
+    negative values. The negation causes Decimal inputs to apply context
+    rounding, making the results slightly different than that obtained
+    by statistics.median().
+    """
+
+    iterator = iter(iterable)
+
+    if window is not None:
+        window = index(window)
```
bbayles, line 1430: Why `index()` here?

rhettinger: Perhaps there should also be a test to reject non-positive window sizes.

bbayles: Should we be using it on `islice_extended`?

rhettinger: The checks are already in the downstream calls, so incorrect arguments are getting rejected. However, that only occurs after the generator starts running. Ideally, the arguments would be checked on instantiation, but I think `islice_extended` is probably fine as-is and it hasn't caused any known problems in the real world.

```python
>>> iterator = islice_extended('abcdefgh', -4.0, -1)  # Error not detected on instantiation
>>> list(iterator)  # Error detected when run
...
TypeError: an integer is required
```

rhettinger:
For the record, the issue with `Decimal` context rounding is that double negation does not always round-trip:

```python
>>> from decimal import Decimal
>>> from math import pi
>>> Decimal(pi)
Decimal('3.141592653589793115997963468544185161590576171875')
>>> - - Decimal(pi)
Decimal('3.141592653589793115997963469')
>>> Decimal(pi) == - - Decimal(pi)
False
```

bbayles:
> Possibly rename the iterable parameter to data, consistent with statistics.median? Or keep as-is to match more_itertools conventions and to emphasize the lazy evaluation which is the principal use case for running_median()?

+0 for as-is on this one.

> Happy with the O(n) but mostly fast steps to maintain a sorted window? Or add more complex O(log n) code (IndexableSkiplist, blist, SortedContainers, etc)? Based on Grant Jenks' notes, the current list insort/bisect/del technique can be expected to win for window sizes up to several thousand.

+1 on the current implementation - it's reasonably clear what's going on, and the other options are probably overkill for this general purpose library.

bbayles:
This is great, thanks - can't wait to find a place to use it.

Issue #1047, opened 2025-08-07
[maintenance] Use pre-commit + tox instead of Makefile


ktbarrett:
The Makefile specifically calls out `python3 -m pip`, etc. and that doesn't work in my environments whether that's uv venv or a venv created with a recent version of Python.

tox is more than capable of setting requirements and running tests or building docs. And pre-commit can handle things like running ruff and checking the package and do so automatically on every commit made. pre-commit can also be run in CI easily.

JamesParrott:
On Windows 11, for me at least, neither `uv venv` (0.7.11) nor an official Python 3.13.5 (`python -m venv`) include `python3` launchers in their venvs anymore, and python3 falls back to a link to the MS App Store even when either of those newer venvs is activated.

ktbarrett:
It doesn't work on some Linux systems by default. You need packages like `python-is-python3`.

Also if toxic doesn't suffice, why is there a toxfile in the repo? But I'm not a huge fan either, I prefer nox which looks more like a Makefile, but is Python, and handles all the environment better than Makefiles.

bbayles:
> It doesn't work on some Linux systems by default. You need packages like python-is-python3.

Not once a virtual environment is activated, right?

> Also if tox doesn't suffice, why is there a toxfile in the repo?

I forgot about this, whoops - it's from many years ago.

Anyway, I amend my earlier statement to:
* l'll take a PR that changes `python3` to `python`
* And that mentions that you should activate your virtual environment to use the `Makefile` to run tests

ktbarrett:
> Not once a virtual environment is activated, right?

At least with uv, `python -m pip` will still not work as a Python `pip` package is never installed. `uv pip` is a pip-like tool that includes other uv features and integrations.

Is there any interest in moving to [`nox`](https://nox.thea.codes/en/stable/) instead of Makefiles? It would certainly help with environment management and those poor souls stuck on Windows that don't have easy access to a Unix environment with `make`.

PR #1052, opened 2025-08-08
Switch testing to use nox

ktbarrett:
### Issue reference
xref #1047. For your consideration. No feelings will be hurt if it's rejected, but this is an improvement over the previous flow IMO.

### Changes
Replaces `tox` and `make` with `nox`. nox can automatically install and use multiple versions of Python, which replaces the functionality of `nox` and can function like a Makefile while better handling Python virtual environment creation and removes the dependency on non-Python `make`.

### Checks and tests
👍


bbayles:
Thanks for looking at this, but I'm -1 on the switch. I may revisit the PR template to specify how to run tests more explicitly.

PR #1056, opened 2025-08-13
Issue 1044: Add new tool: extract()

rhettinger:


tests/test_more.py
```diff
@@ -6273,3 +6273,104 @@ def test_key(self):
...
+        data = list('abcdefghijklmnopqrstuvwxyz')
+
+        # Test iterator inputs, increasing and decreasing indices, and repeats.
+        self.assertEqual(
+            list(extract(iter(data), iter([7, 4, 11, 11, 14]))),
+            ['h', 'e', 'l', 'l', 'o'],
+        )
+
+        # Empty indices
+        self.assertEqual(list(extract(iter([data]), iter([]))), [])
```
pochmann3, line 6290: `[data]` looks odd/unintended, at least I see no reason for wrapping `data` in another list. (Btw I also don't see why `data` *is* a list instead of just the string.)

PR #1058, opened 2025-08-15
Set minimum version to Py3.10. 

rhettinger:
Python 3.14 will be final in a few days (it is currently at rc2).  Suggest bumping up the minimum supported version to Py3.10.

This eliminates the deprecated `zip_equal` adapter and its companion `_zip_strict` in favor of the builtin `zip` function which directly supports the `strict` option.  Also `more_itertools.pairwise` is no longer needed.


more_itertools/more.py
```diff
@@ -1954,10 +1918,10 @@ def sort_together(
                 *get_key_items(zipped_items)
             )
 
-    zipper = zip_equal if strict else zip
-    return list(
-        zipper(*sorted(zipper(*iterables), key=key_argument, reverse=reverse))
-    )
+    transposed = zip(*iterables, strict=strict)
```
pochmann3, line 1921: I suggest more meaningful names instead of `transposed` etc, inspired by the function name: `together`, `sorted_together`, and `separated`.

bbayles:
I made a reminder for myself to merge this in October, after 3.9 goes EOL.

PR #1066, opened 2025-08-28
Improve count_cycle()

rhettinger:
Prefer an iterator stack to nested generator expressions.

Baseline:
```
% python3.14 -m timeit -s 'from more_itertools import consume, count_cycle' 'consume(count_cycle(range(1000), 3))'
5000 loops, best of 5: 93 usec per loop
```

Patched:
```
% python3.14 -m timeit -s 'from more_itertools import consume, count_cycle' 'consume(count_cycle(range(1000), 3))'
10000 loops, best of 5: 36.4 usec per loop
```

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -2437,9 +2437,7 @@ def count_cycle(iterable, n=None):
     if not seq:
         return iter(())
     counter = count() if n is None else range(n)
-    return zip(
-        chain.from_iterable(map(repeat, counter, repeat(len(seq)))), cycle(seq)
-    )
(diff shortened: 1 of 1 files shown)


pochmann3:
Unless you dislike the dependency again, you could shorten your

```python
    return zip(chain.from_iterable(map(repeat, counter, repeat(len(seq)))), cycle(seq))
```

to:

```python
    return zip(repeat_each(counter, len(seq)), cycle(seq))
```

(Another one I wrote was `return chain.from_iterable(map(zip, map(repeat, counter), repeat(seq)))`, but I think yours is better.)

PR #1072, opened 2025-09-02
Issue 1070: Add random_derangement()

rhettinger:


rhettinger:
> why did you move this to the recipes module?

Because I'm also adding this to CPython docs :-)

more_itertools/recipes.py
```diff
@@ -1469,3 +1470,20 @@ def running_median(iterable, *, maxlen=None):
...
+
+def random_derangement(iterable):
+    """Return a random derangement of elements in the iterable.
+
+    Equivalent to but much faster than ``choice(list(derangements(iterable)))``.
+
+    """
+    seq = tuple(iterable)
+    if len(seq) < 2:
+        raise ValueError('derangments require at least two values')
```
Peiffap, line 1483: nit
```suggestion
        raise ValueError('derangements require at least two values')
```

pochmann3, line 1483: Only length 1 is impossible. Length 0 is perfectly fine, it has one derangement, the empty permutation.

bbayles:
Anything more on this besides the typo fix noted above? Should the length 1 case be handled?

PR #1075, opened 2025-09-28
Issue 1074: Add `zip_mappings()`

### LINKED ISSUE #1074: Add `zip_mappings`
### Description
I would like to propose an itertools that combines the functionality of the `items` method on mappings and `zip`. Like `items`, it would yield tuples with keys and values. There would be one result for each unique key in the union of the keys of all the mappings. Each key would be the first element of the tuple as in `items`. Then, like `zip`, the remaining _n_ elements of the tuple would be the corresponding values of the _n_ mappings with a default of `None` or another value specified by the user.

### References
`items` is useful when one is constructing a new dictionary by applying some function to the values of an existing dictionary, as in `new_dict = {key: function(value) for key, value in existing_dict.items()}`. The proposed itertool generalizes this to functions of multiple arguments that are supposed to come from different dicts, which is a use case I have encountered.

### Examples
```
from itertools import chain
from more_itertools import unique
from operator import methodcaller
def zip_mappings(*mappings, default=None):
    for key in unique(chain(*mappings)):
        yield key, *map(methodcaller("get", key, default), mappings)
```
A type check for `collections.abc.Mapping` could be added.

gtkacz:
### Issue reference

more-itertools/more-itertools#1074

### Changes
The new `zip_mappings` function allows users to iterate over the union of keys from multiple mappings, yielding a tuple for each key containing the key itself and the corresponding values from each mapping. Closes #1074 

### Checks and tests

✔️

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -64,6 +64,7 @@ Python iterables.
 |                        | `zip_offset <https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.zip_offset>`_,                                                                         |
 |                        | `zip_equal <https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.zip_equal>`_,                                                                           |
 |                        | `zip_broadcast <https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.zip_broadcast>`_,                                                                   |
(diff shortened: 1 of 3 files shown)


bbayles:
#1074 was deferred, so closing - thanks for the contribution.

PR #1101, opened 2025-11-13
Add a monotonic option to extract()

rhettinger:
Currently, the *indices* are fully consumed before starting. That arises from the need to sort the indices into non-decreasing order.

Add a `monotonic=True` option that allows *indices* to be consumed lazily when the indices are known to already be in non-decreasing order.  For that case, neither sorting nor buffering are needed, and infinite inputs can be supported.


bbayles:
Nice enhancement - thanks!

PR #1107, opened 2025-11-23
Issue 1032: Improve common case typehints for always_iterable

rhettinger:
Rather than try to handle every case, this PR aims at the common cases:

1) `obj` is None
2) `base_type` is None
3) `base_type` is the default `(str, bytes)`
4) `base_type` is exactly `str` or exactly `bytes`

Only those exact matches are handled. Other cases continue to default to `Iterable[Any]`.

This PR also removes the outdated Python 3.9 support in the definition of `_ClassInfo`.

------------------

The following examples are now all improved, giving the expected `reveal_type` instead of a non-specific `Iterable[Any]`.

```python
from more_itertools import always_iterable
from typing import reveal_type

i1 = always_iterable(None)  # Iterator[tuple[()]]
reveal_type(i1)
assert list(i1) == []

i2 = always_iterable(b'abc', None)  # typing.Iterator[builtins.int]
reveal_type(i2)
assert list(i2) == [97, 98, 99]

i3 = always_iterable('abc', None)  # typing.Iterator[builtins.str]
reveal_type(i3)
assert list(i3) == ['a', 'b', 'c']

i4 = always_iterable(['abc'], None)  # typing.Iterator[builtins.str]
reveal_type(i4)
assert list(i4) == ['abc']

i5 = always_iterable(10, None)  # typing.Iterator[builtins.int]
reveal_type(i5)
assert list(i5) == [10]

i6 = always_iterable([10], None)  # typing.Iterator[builtins.int] 
reveal_type(i6)
assert list(i6) == [10]

i7 = always_iterable(b'abc')  # typing.Iterator[builtins.bytes]
reveal_type(i7)
assert list(i7) == [b'abc']  

i7b = always_iterable(b'abc', bytes)  # typing.Iterator[builtins.bytes]
reveal_type(i7b)
assert list(i7b) == [b'abc'] 

i8 = always_iterable('abc')  # typing.Iterator[builtins.str]
reveal_type(i8)
assert list(i8) == ['abc']  

i8b = always_iterable('abc', str)  # typing.Iterator[builtins.str]
reveal_type(i8b)
assert list(i8b) == ['abc']  

i9 = always_iterable(10)  # typing.Iterator[builtins.int]
reveal_type(i9)
assert list(i9) == [10] 

i10 = always_iterable([10])  # typing.Iterator[builtins.int]
reveal_type(i10)
assert list(i10) == [10] 

i10b = always_iterable([10], str)  # typing.Iterator[builtins.int]
reveal_type(i10b)
assert list(i10b) == [10]

i10c = always_iterable([10], bytes)  # typing.Iterator[builtins.int]
reveal_type(i10c)
assert list(i10c) == [10]
```


rhettinger:
Here's a tool to generate test cases to make sure nothing else changed and that mypy checks that passed before will still pass.

```python
from more_itertools import powerset, always_iterable
from itertools import product
from contextlib import redirect_stdout

bts = ('int', 'bytes', 'str', 'list', 'dict')
vals = (None, 1, b'abc', 'abc', list('abc'), tuple('abc'), {'a': 1, 'b': 2})

exprs = []
for val, bt in product(vals, powerset(bts)):
    bt = tuple(sorted(bt, reverse=True))
    if not bt:
        expr = f'always_iterable({val!r}, None)'
    elif len(bt) == 1:
        expr = f'always_iterable({val!r}, {bt[0]})'
    else:
        expr = f'always_iterable({val!r}, ({", ".join(bt)},))'
    exprs.append(expr)

results = [list(eval(expr)) for expr in exprs]

with open('always_iterable_tests.py', 'w') as test_file:
    with redirect_stdout(test_file):
        print('from typing import reveal_type')
        print('from more_itertools import always_iterable\n')
        for i, expr, result in zip(count(), exprs, results):
            print(f'iterator{i} = {expr}')
            print(f'reveal_type(iterator{i})')

            test = f'assert list({expr}) == {result!r}'
            exec(test)
            print(test)
            print()
```

bbayles (approved):

Issue #1113, opened 2025-12-05
Latin hypercube sample products


rx-dwoodward:
### Description

A Cartesian product search strategy akin to `itertools.product` that instead of depth-first searching through the Cartesian product space, searches via a latin hypercube / diagonal sampling. This has the property that after sampling `len(longest_iterable)` times, you will have sampled the minimum number of products possible to ensure you have sampled at least every iterable element once. Likewise with after `10 * len(longest_iterable)` samples, you will have sampled every iterable element at least 10 times. 

This is useful as its a more explorative way of sampling through the Cartesian product than the 'depth-first' approach done by `itertools.product`. The main use of this is when the full Cartesian product space is prohibitively large to hold in memory altogether. Note this implementation samples uniquely and is (relatively) stateless.

I have code for this and figured it might be useful to chuck into `more_itertools` - I also have a code snippet to execute pseudo-random sampling of the Cartesian product space, also in a stateless and unique way, happy to add that too if interested!

### References
https://stackoverflow.com/questions/61434268/how-to-make-itertools-product-random heres a SO post about itertools product random sampling - on this Cartesian product space in the post you could ofc enumerate all possible combinations, but for larger iterables this would become expensive memory wise relatively quickly.

Additionally for myself, I wanted to find a way of generating combinations but seeing all elements in the minimum number of samples possible (for further downstream processing) in a stateless and unique way and this was the solution to that problem.

### Examples
```
from more_itertools import hypercube_sample_products

# total Cartesian product space: 33,750,000
iterables = [list(range(50)), list(range(600)), list(range(45)), list(range(25))]

samples = hypercube_sample_products(*iterables)

for i, sample in enumerate(samples):   
     print(sample)
```

and the output would be akin to:

```
(0, 0, 0, 0)
(1, 1, 1, 1)
(2, 2, 2, 2)
...
(24, 24, 24, 24)
(25, 25, 25, 0)
...
(44, 44, 44, 19)
(45, 45, 0, 20)
...
(49, 49, 4, 24)
(0, 50, 5, 0)
...
...
(49, 599, 14, 24)
(0, 0, 15, 0)
...
...
...
```
and so on. The way the algorithm works it ensures that when a lowest common multiple 'period' comes back round (in the above case every 1800 samples) there is an offset such that the period never ends up repeating.

Let me know if you think this would be suitable to add to `more_itertools` !

rx-dwoodward:
[...]

PR #1114, opened 2025-12-05
feat: add hypercube_product sampling

rx-dwoodward:
### Issue reference
(Issue reference)
[more-itertools/more-itertools#XXXX](https://github.com/more-itertools/more-itertools/issues/1113)

### Changes

This PR adds a minimal implementation for hypercube sampling products of a Cartesian product space in a maximally explorative way by sampling from the diagonal of the Cartesian product hypercube. 

See issue linked for more details. 

Looking back at `itertools.product` they do state that the elements will be held in memory (as they are here) so I could always do `sequences = tuple(map(tuple, iterables))` as the first step and proceed now knowing the sizes of all the finite iterables from there, but this current implementation doesn't require eager consumption of each iterable immediately. 

Depends what y'all think is most readable / preferable, lmk!

### Checks and tests
✅

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -22,7 +22,19 @@ from itertools import (
     zip_longest,
     product,
 )
-from math import comb, e, exp, factorial, floor, fsum, log, log1p, perm, tau
+from math import (
+    comb,
+    e,
+    exp,
+    factorial,
+    floor,
+    fsum,
+    lcm,
+    log,
+    log1p,
+    perm,
+    tau,
+)
 from math import ceil
 from queue import Empty, Queue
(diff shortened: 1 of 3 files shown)


bbayles:
I'm going to close this, following my comment here:
> I'd like to have something that gives something other than the depth-first order of product, but I don't think I want to have both the easy-to-describe random functions in PR https://github.com/more-itertools/more-itertools/pull/1116 AND a relatively inaccessible "hypercube" method.

I'm not sure this quite fits the bill as it stands now.

@rx-dwoodward said:
> That being said if this is too niche of a use-case and you do not think its worth adding that's also fine

Appreciate it, and thanks for your work on this all the same.

PR #1116, opened 2025-12-21
Issue 1113: Add random ordered combinatoric functions

rhettinger:
These functions fulfill a need for random ordered exploration in the search space for the combinatoric functions.


pochmann:
Are these intended only for small inputs or also large ones? For large ones, `nth_permutation` can be slow partly because of the final line popping from all over the place:

```python
    return tuple(map(pool.pop, result))
```

Instead of a pop from index `i`, which shifts all later elements, you could pop the *last* element and put it at index `i`. Not for the regular `nth_permutation`, because that messes with the order of the permutations. But for your random order, that doesn't matter. (Might even improve the randomness?)

rhettinger:
@bbayles Here's a smaller ask.  Add `random_ordered_range()` to serve as primitive for exploring data in random order without duplicates. It works with indexable collections and combinatorics.

Since the last edit, I removed the combinatoric specific variants. Also, I bumped up the batch size to 1024 for better quality while still limiting max memory consumption.

bbayles (approved):

PR #1119, opened 2026-01-02
Simplify nth_xxx code and fix invalid exception.

rhettinger:
Like what was done for `nth_combination`, fix the `ValueError` exception for a large value of *r*.  When `r >= n`, the number of `combinations` and `permutations` are zero.  For `combinations_with_replacement`, a large *r* gives more results:

```python
len(list(permutations('abc', r=5)))
0
len(list(combinations('abc', r=5)))
0
len(list(combinations_with_replacement('abc', r=5)))
21
```

Also, repeat the simplifications for error checks for all the nth variants.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -22,20 +22,8 @@ from itertools import (
     zip_longest,
     product,
 )
-from math import (
-    comb,
-    e,
-    exp,
-    factorial,
-    floor,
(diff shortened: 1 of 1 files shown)


bbayles (approved):

bbayles:
Many thanks!

PR #1125, opened 2026-02-02
More fixes for repeat with iterator arguments

rhettinger:


bbayles:
Thanks! GitHub Actions is having problems, but nothing is broken with this PR.

bbayles (approved):

PR #1128, opened 2026-02-20
Fix numeric_range slicing with negative step returning empty range

bysiber:
## Summary

`numeric_range.__getitem__` returns an empty range when sliced with a negative step (e.g., `[::-1]`), even though `reversed()` works correctly on the same range.

## Problem

When a slice has a negative step, the default values for `None` start and stop should be adjusted — start should default to the last element and stop to before the first element. But the current code always defaults `None` start to `self._start` and `None` stop to `self._stop`, regardless of the step direction.

```python
from more_itertools import numeric_range

nr = numeric_range(0, 10, 2)  # [0, 2, 4, 6, 8]

list(reversed(nr))   # [8, 6, 4, 2, 0] — correct
list(nr[::-1])       # [] — should be [8, 6, 4, 2, 0]

# Compare with built-in range:
r = range(0, 10, 2)
list(r[::-1])        # [8, 6, 4, 2, 0] — correct
```

The issue is in the `__getitem__` method. For `[::-1]`:
- `start` defaults to `self._start` (0), should be the last element (8)
- `stop` defaults to `self._stop` (10), should be before the first element (-2)
- Result: `numeric_range(0, 10, -2)` which is empty, instead of `numeric_range(8, -2, -2)`

## Fix

Use `slice.indices()` to correctly normalize the start/stop/step accounting for the step direction, then convert the integer indices back to the range's value space. This approach handles all combinations of positive/negative/None start, stop, and step correctly — matching built-in `range` behavior.

### DIFF
--- a/tests/test_more.py
+++ b/tests/test_more.py
@@ -2856,6 +2856,11 @@ class NumericRangeTests(TestCase):
             ((1.0, 9.0, 1.5), slice(None, -10, 3), (1.0, 1.0, 4.5)),
             ((1.0, 9.0, 1.5), slice(None, 10, 3), (1.0, 9.0, 4.5)),
             (
+                (10.0, 0.0, -2.0),
+                slice(None, None, -1),
+                (2.0, 12.0, 2.0),
+            ),
+            (
                 (Decimal("1.0"), Decimal("9.0"), Decimal("1.5")),
                 slice(1, -1, None),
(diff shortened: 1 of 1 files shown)


bysiber:
Added a test case that slices a negative-step range with `[::-1]` — this would have returned an empty range before the fix. Thanks for the suggestion!

PR #1131, opened 2026-03-14
add tee_filtered

### LINKED ISSUE #1130: Propose a new itertool: tee_filtered
### Description

While `itertools.tee` is useful for splitting a stream, it suffers from severe memory bloat if the branches are consumed at different speeds or if one branch skips a large number of elements. `tee_filtered` could solves this in some situations, by only appending elements to a branch's queue if the elements satisfy the specific predicate.

### References

This is an actual problem met by myself:

```python
from collections import deque
import itertools
from typing import Callable, Iterable, Iterator, Sequence


# Simulates a time-consuming process.
def lazy_process(input: Iterable[int]):
    # It process batch by batch.
    # Therefore, we prefer 'lazy_process([1, 1])' rather than 2 times of 'lazy_process([1])'.
    # (This is a still simplified version. It might also involve caching or something else, 
    #  so we generally prefer to pass in all the content to be processed, 
    #  rather than passing it in chunks.)
    def a_time_consuming_operation(value_batch: Sequence[int]) -> Sequence[int]:
        ...
        return value_batch
    
    def auto_batch_size() -> int:
        ...
        return 4

    input_batch: Sequence[int]
    for input_batch in itertools.batched(input, auto_batch_size()):
        result_batch: Sequence[int] = a_time_consuming_operation(input_batch)

        i: int
        r: int
        for i, r in zip(input_batch, result_batch):
            # Here the actual result is 'i * r', so actually if the input is zero, we could skip the operation.
            # However, this function might provided by others and we can't optimize it.
            yield i * r


# We assume that the data source is also a very complex generator.
def input_generator():
    yield from ([0] * 100 + [1] * 100 + [2] * 100 + [3] * 100)


def run_the_process_with_tee():
    given_values: Iterable[int] = input_generator()
    # We would like to skip those zeros (we know the result is zero).
    # A solution might be:
    
    values1: Iterable[int]
    values2: Iterable[int]
    values1, values2 = itertools.tee(given_values)
    result: Iterator[int] = iter(lazy_process((v for v in values2 if v != 0)))

    input: int
    for input in values1:
        print("input", input)
        if input == 0:
            print("output:", 0)
        else:
            print("output:", next(result))

    # However, if there are 100 zeros at the beginning, they will be storaged in 'tee', until a non-zero value occurs.
    # If the inputs are very complex (and that's why we use a generator for input), this could fill up the memory.
```

I seem to have encountered this problem multiple times in a certain project of mine. Maybe because in that project, it's easy to have both caching and batch processing at the same time.

### Possible Implementation

Actually I have solved this problem by (Note that this implemention is not thread-safe; I don't know if we require it to be thread-safe):

```python
def tee_filtered[T](iterable: Iterable[T], 
                    *filters: Callable[[T], bool]) -> tuple[Iterator[T], ...]:
    it: Iterator[T] = iter(iterable)
    queues_and_filters: list[tuple[deque[T], Callable[[T], bool]]] = [
        (deque(), filter) for filter in filters
    ]

    def generator(q: deque[T]):
        while True:
            while len(q) == 0:
                try:
                    next_value: T = next(it)
                except StopIteration:
                    return
                
                queue: deque[T]
                filter: Callable[[T], bool]
                for queue, filter in queues_and_filters:
                    if filter(next_value):
                        queue.append(next_value)
            yield q.popleft()

    return tuple(generator(queue) for queue, _ in queues_and_filters)
```

In fact, it works quite well in my scenario, but I feel like I don't have a place to define this tool, so I might as well submit it to this project.

### Examples

```python
def run_the_process_with_tee_filtered():
    given_values: Iterable[int] = input_generator()
    
    values1: Iterable[int]
    values2: Iterable[int]
    values1, values2 = tee_filtered(given_values, 
                                    lambda _: True, 
                                    lambda x: x != 0)
    result: Iterator[int] = iter(lazy_process(values2))

    input: int
    for input in values1:
        print("input", input)
        if input == 0:
            print("output:", 0)
        else:
            print("output:", next(result))
```

yueyinqiu:
### Issue reference

more-itertools/more-itertools#1130

closes #1130

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,6 @@ venv/
 
 # Ruff
 .ruff_cache/
+
+# uv
+uv.lock
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -164,6 +164,7 @@ __all__ = [
     'substrings_indexes',
     'synchronized',
     'takewhile_inclusive',
+    'tee_filtered',
     'time_limited',
     'unique_in_window',
     'unique_to_each',
@@ -5433,3 +5434,23 @@ class _concurrent_tee:
                     link[1] = [None, None]
         value, self.link = link
         return value
+
+
+def tee_filtered(iterable, *filters):
+    it = iter(iterable)
+    queues_and_filters = [(deque(), filter) for filter in filters]
+
+    def generator(q):
+        while True:
+            while len(q) == 0:
+                try:
+                    next_value = next(it)
+                except StopIteration:
+                    return
+
+                for queue, filter in queues_and_filters:
+                    if filter(next_value):
+                        queue.append(next_value)
+            yield q.popleft()
+
+    return tuple(generator(queue) for queue, _ in queues_and_filters)
--- a/more_itertools/more.pyi
+++ b/more_itertools/more.pyi
@@ -136,6 +136,7 @@ __all__ = [
     'substrings_indexes',
     'synchronized',
     'takewhile_inclusive',
+    'tee_filtered',
     'time_limited',
     'unique_in_window',
     'unique_to_each',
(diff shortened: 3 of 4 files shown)


bbayles:
Deferred - see my comment in [the issue](https://github.com/more-itertools/more-itertools/issues/1130#issuecomment-4076092283). Thanks for the PR all the same.

PR #1144, opened 2026-04-02
Issue 1143: Another change to always_iterable type hints

### LINKED ISSUE #1143: How to adapt to typing change for always_iterable?
Over in setuptools I noticed that the typing tests started failing with more-itertools 11.0:

```
setuptools/config/expand.py:125: error: Argument 2 to "join" has incompatible type "PathLike[str] | Iterable[str | PathLike[str]]"; expected "str | PathLike[str]"  [arg-type]
```

See e.g. [this CI run](https://github.com/pypa/setuptools/actions/runs/23911615010/job/69734806337?pr=5207). If I pin more-itertools to `<11.0`, the issue goes away.

I asked my resident typing expert @jorenham, and he thinks the issue might be this code:

https://github.com/more-itertools/more-itertools/blob/5ea81c48727f9906879e30b352772bc3869c4e88/more_itertools/more.pyi#L444-L447

> This looks like the issue: The obj: _T will also match obj: Iterable[_T] , so I can see why mypy would infer -> Iterator[_T] | _T because of the overlap

bbayles:
This PR updates the type hints for `always_iterable`.

These previously changed in https://github.com/more-itertools/more-itertools/pull/1107 in response to https://github.com/more-itertools/more-itertools/issues/1032.

@maltevesper, I don't suppose you could see if your original issue is still resolved with this change?

Closes #1143, I hope.


rhettinger:
I don't think this is what we want.

These now give `Iterator[Any]`:

* `i4 = always_iterable(["abc"], None)  # typing.Iterator[str]`
* `i6 = always_iterable([10], None)  # typing.Iterator[int]`
* `i10 = always_iterable([10])  # typing.Iterator[builtins.int]`
* `i10b = always_iterable([10], str)  # typing.Iterator[int]`
* `i10c = always_iterable([10], bytes)  # typing.Iterator[int]`

The first new test case should give `Iterable[int]` but now gets `Iterator[Any]`:

```python
def test_optional_iterable(data: list[int] | None):
    it = always_iterable(data)
    reveal_type(it)
```    

The second new test case should also give `Iterable[int]` but instead reveals as `Iterator[Union[int, None]]`:

```python
def test_optional_scalar(data: int | None):
    it = always_iterable(data)
    reveal_type(it)
```

PR #1154, opened 2026-04-12
Add __class_getitem__ to peekable for generic type subscript support

### LINKED ISSUE #849: TypeError: type 'peekable' is not subscriptable

DORI2001:
Fixes #849.

## Summary

- Adds `__class_getitem__ = classmethod(types.GenericAlias)` to the `peekable` class
- Adds `import types` to the top of `more.py`
- This allows `peekable[str]` to work as a generic alias at runtime, matching how `list[str]`, `functools.partial[str]`, and other CPython classes work

## Motivation

Currently `peekable[str]` raises `TypeError: type 'peekable' is not subscriptable`. This is the same pattern as CPython's `functools.partial` which uses `__class_getitem__ = classmethod(GenericAlias)`.

## Testing

```python
from more_itertools import peekable

# Before: TypeError: type 'peekable' is not subscriptable
# After:
print(peekable[str])  # more_itertools.more.peekable[str]

# Normal usage still works:
p = peekable(['a', 'b', 'c'])
print(next(p))  # 'a'
```

## Notes

This follows the same approach rejected PR #857 attempted, but uses `classmethod(types.GenericAlias)` instead of `def __class_getitem__(cls, key): return cls`, which is the correct CPython-style implementation that returns a proper `GenericAlias` object (not just the class itself).


DORI2001:
Hi! Just wanted to check in on this PR — would love to get feedback if there's anything to address. Thanks!

bbayles:
[This comment](https://github.com/more-itertools/more-itertools/pull/857/changes#r1918916534) has a request for a test; let's do that - thanks.

DORI2001:
@bbayles Done — added a `test_class_getitem` test to `PeekableTests` that asserts:

- `peekable[str]` returns a `types.GenericAlias` instance
- `__origin__` is `peekable`
- `__args__` is `(str,)`

Also registered `__class_getitem__ = classmethod(types.GenericAlias)` in `more.pyi` to clear the `stubtest` failure that was breaking CI on Python 3.10.

bbayles:
Thanks for the update. That test is still failing. You can try it locally with `make check` - it should run the same command that GitHub Actions does.

DORI2001:
@bbayles Fixed — the assignment form `__class_getitem__ = classmethod(types.GenericAlias)` triggers a mypy `[var-annotated]` error in `.pyi` stub files. Replaced it with the decorated method form:

```python
@classmethod
def __class_getitem__(cls, item: Any) -> types.GenericAlias: ...
```

That should clear the stubtest failure.

PR #1157, opened 2026-05-03
Add generator support to serialize()

rhettinger:
Sync with edits in the CPython versions:

* Make the attributes private
* Support generator methods: throw, send, close


bbayles:
The code looks good. Would it make sense to have examples in the docstring that illuminate support for these methods?

rhettinger:
No good, concise examples come to mind. The other generator methods are rarely used.

bbayles:
> The other generator methods are rarely used.

This was why I thought examples would be useful, ha.

I'll think about what might make a good illustration.

PR #1167, opened 2026-06-12
Fix broken test

rhettinger:
* The `Str` test class has a bug.   It's `super.__hash__` should be `super().__hash__`. 
* The expected number of sets produced should be `2**4` because there are 4 distinct elements in the input.
* The expected number of `hash()` calls should be 7 because there are 7 elements in the input.

### DIFF
--- a/tests/test_more.py
+++ b/tests/test_more.py
@@ -6302,14 +6302,16 @@ class PowersetOfSetsTests(TestCase):
         hash_count = 0
 
         class Str(str):
-            def __hash__(true_self):
+            def __hash__(self):
                 nonlocal hash_count
                 hash_count += 1
-                return super.__hash__(true_self)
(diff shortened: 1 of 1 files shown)


tests/test_more.py
```diff
@@ -6302,14 +6302,16 @@ def test_hash_count(self):
...
         class Str(str):
-            def __hash__(true_self):
+            def __hash__(self):
                 nonlocal hash_count
                 hash_count += 1
-                return super.__hash__(true_self)
+                return super().__hash__()
 
+        # Seven input values should have 7 hash calls.
+        # Four distinct input values should have 2 ** 4 powersets.
```
pochmann: Not 16 powersets but 16 subsets. It's one powerset.

bbayles (approved):

Issue #1168, opened 2026-06-13
Propose a new itertool: primes


zakaria-zoulati:
### Issue reference
more-itertools/more-itertools#XXXX

### Changes

Adds `primes()` — an **unbounded, lazy prime number generator**.

Unlike the existing `sieve(n)`, which requires a known upper limit and builds a
table proportional to that limit, `primes()` runs forever using a *segmented*
Sieve of Eratosthenes. Primes are produced one block at a time, so memory use
stays proportional to the square root of the largest prime yielded rather than
the prime itself.

```python
>>> from itertools import islice, takewhile
>>> list(islice(primes(), 10))
[2, 3, 5, 7, 11, 13, 17, 19, 23, 29]
>>> list(takewhile(lambda p: p < 30, primes()))
[2, 3, 5, 7, 11, 13, 17, 19, 23, 29]

pochmann:
How does yours compare to the implementations in the prior issue?

https://github.com/more-itertools/more-itertools/issues/987

zakaria-zoulati:
> How does yours compare to the implementations in the prior issue?
> 
> [#987](https://github.com/more-itertools/more-itertools/issues/987)

Thanks @pochmann — I hadn't fully accounted for #987 before opening this.

Comparing honestly: my segmented `bytearray` version doesn't beat the
solutions there. rhettinger's `sieve(n=None)` reuses the existing `sieve`
machinery and is the cleaner API (no new public name), and pochmann3's
dict version is far more memory-efficient (~0.1 MB vs my per-block
allocations) and faster.

Given that, a separate `primes()` doesn't seem like the right call. The
better path is the `sieve(n=None)` design from #987. I'm happy to either
close this in favor of #987, or — if it's still wanted — open a fresh PR
implementing `sieve(n=None)` per rhettinger's snippet, with tests and the
docstring update. Which would you prefer?

bbayles:
I'm open to making `sieve(n=None)` generate primes without an explicit upper bound. I think we should either do something very simple, like Raymond's "With a little more effort the islice rescan can be avoided" implementation, or something really efficient like the [iterative version](https://github.com/more-itertools/more-itertools/issues/987#issuecomment-3169091435) of pochmann's recursive generator.

I'm still slightly dissatisfied with the translation from recursive-to-iterative on that, but if we can make something cleaner then I think it would be the winner.

bbayles:
Current state of the art:

```python
@dataclass(slots=True)
[...]

PR #1169, opened 2026-06-13
Add primes generator

zakaria-zoulati:
### Issue reference
more-itertools/more-itertools #1168 

### Changes

Adds `primes()` — an **unbounded, lazy prime number generator**.

Unlike the existing `sieve(n)`, which requires a known upper limit and builds a
table proportional to that limit, `primes()` runs forever using a *segmented*
Sieve of Eratosthenes. Primes are produced one block at a time, so memory use
stays proportional to the square root of the largest prime yielded rather than
the prime itself.

```python
>>> from itertools import islice, takewhile
>>> list(islice(primes(), 10))
[2, 3, 5, 7, 11, 13, 17, 19, 23, 29]
>>> list(takewhile(lambda p: p < 30, primes()))
[2, 3, 5, 7, 11, 13, 17, 19, 23, 29]

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -333,6 +333,7 @@ The tools focus on math with integers.
 .. autofunction:: is_prime
 .. autofunction:: multinomial
 .. autofunction:: sieve
+.. autofunction:: primes
 .. autofunction:: totient
 
 
--- a/more_itertools/recipes.py
+++ b/more_itertools/recipes.py
@@ -69,6 +69,7 @@ __all__ = [
     'polynomial_derivative',
     'powerset',
     'prepend',
+    'primes',
     'quantify',
     'reshape',
     'random_combination_with_replacement',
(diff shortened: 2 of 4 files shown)


DroneBetter:
I would say don't do this because it duplicates functionality from `sympy.ntheory.generate.Sieve` which is already quite considerably optimised

jonathandung:
At the risk of bikeshedding, there isn't supposed to be a space in your issue reference.

jonathandung:
Also, just my two cents, but it's not like `more_itertools` is the only library that can achieve the math stuff (by which I refer to [these](https://more-itertools.readthedocs.io/en/stable/api.html#math)), and specialized libraries do in fact do them much faster. However, they are, after all, specialized, and possibly not pure Python. I think it's perfectly fine to duplicate functionality here.

bbayles:
See https://github.com/more-itertools/more-itertools/issues/1168#issuecomment-4735308620 for the current thinking on this proposal

PR #1172, opened 2026-06-16
Annotate `return self` with `typing.Self`

jorenham:
### Issue reference
No issue

### Changes

This narrows the return types of `__iter__` and `__enter__` methods that `return self` at runtime as ` -> typing.Self`.

This matters because classes in question aren't `@final` so can therefore be subclassed. If those subclasses then don't override `__iter__` or `__enter__`, then the return type of these inherited methods will now be more accurate.

Another advantage is that `-> Self` clearly describes the runtime behavior (i.e. `return self`) in a natural way, so it can also be seen as a documentation improvement in that sense.

### Checks and tests
All green :)


bbayles (approved):

bbayles:
Let's change this to `typing.Self` later this year when Python 3.10 goes EOL.

jorenham:
> Let's change this to `typing.Self` later this year when Python 3.10 goes EOL.

Yea exactly. Ruff (pyupgrade) has an autofix for that, so that should be easy.

PR #1173, opened 2026-06-16
Use (co)variant type params in the stubs where appropriate

jorenham:
### Issue reference
No issue

### Changes

When a generic class-bound type parameter (a `TypeVar`) is only used in output positions, it should be covariant. This esnures this is the case, matching the variance of the corresponding generic `collection.abc` (super)class type params.

For `numeric_range.{count,index}` the `_T_co` cannot be used, because that would make it invariant, and also wouldn't be accurate (too narrow). This is why in `collections.abc.Sequence` these method parameters are annotated with `Any`. Following this example, I also changed these `numeric_range` methods to use `Any` so that they nicely match the signatures of their `Sequence` superclass.

### Checks and tests
All green :)


bbayles:
I don't get the `count` and `index` changes. You said:
> This is why in `collections.abc.Sequence` these method parameters are annotated with `Any`. Following this example, I also changed these `numeric_range` methods to use `Any` so that they nicely match the signatures of their `Sequence` superclass.

`collections.abc.Sequence` does indeed use `Any` ([ref](https://github.com/python/typeshed/blob/5e927a88f7c37b2592371a419cf4dd355a3ae4b7/stdlib/typing.pyi#L690-L691)), but here we have:

```
    def count(self: numeric_range[_T, Any], value: _T) -> int: ...
    def index(self: numeric_range[_T, Any], value: _T) -> int: ...  # type: ignore[override]
```

Could you help me understand the discrepancy?

jorenham:
> I don't get the `count` and `index` changes. You said:
> 
> > This is why in `collections.abc.Sequence` these method parameters are annotated with `Any`. Following this example, I also changed these `numeric_range` methods to use `Any` so that they nicely match the signatures of their `Sequence` superclass.
> 
> `collections.abc.Sequence` does indeed use `Any` ([ref](https://github.com/python/typeshed/blob/5e927a88f7c37b2592371a419cf4dd355a3ae4b7/stdlib/typing.pyi#L690-L691)), but here we have:
> 
> ```
>     def count(self: numeric_range[_T, Any], value: _T) -> int: ...
>     def index(self: numeric_range[_T, Any], value: _T) -> int: ...  # type: ignore[override]
> ```
> 
> Could you help me understand the discrepancy?

Oh woops yea I forgot to undo that little experiment, sorry about that.

PR #1174, opened 2026-06-16
Use `_typeshed` protocols in the stubs

jorenham:
### Issue reference

No issue (refactor only)

### Changes

This replaces the custom helper protocols in the `more.pyi` stubs with existing protocols from the type-check-only `_typeshed` stdlib module. 

The leading `_` in `_typeshed` might seem fishy, but this module is used heavily by a lot of stub libraries, including the typeshed stubs themselves, and its API has been very stable for a long time now. So it's not something I'd worry about (and even in the worst case where `_typeshed` gets deleted or something, then some users will see some additional squiggly lines; that's all).

### Checks and tests

Everything is still green.

### DIFF
--- a/more_itertools/more.pyi
+++ b/more_itertools/more.pyi
@@ -24,11 +24,17 @@ from typing import (
     Callable,
     Generic,
     Literal,
+    TypeAlias,
     TypeVar,
     overload,
     type_check_only,
 )
 from typing_extensions import Protocol, Self
+from _typeshed import (
+    SupportsDunderLT,
+    SupportsGetItem,
+    Viewable as _SizedIterable,
+)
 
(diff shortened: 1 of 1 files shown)


bbayles:
Could you explain why this is better?

jorenham:
> Could you explain why this is better?

Less clutter in the stubs, no more need to maintain or test custom `Protocol`s, and the `_typeshed`  protocols are battle-tested and recognizable.

bbayles:
Not sure about this one. I will meditate on the proposal.

bbayles:
Based on what other projects are doing, I'm going to go with this PR. Thanks for the proposal.

PR #1178, opened 2026-06-16
Remove `threading` import and make `queue` import lazy

jonathandung:
### Changes
Since the `_thread` module is documented to be available whenever `threading` is, and `threading` takes 10 ms to load whereas `_thread` takes <100 microseconds, I believe it desirable to replace all instances of `threading.Lock` with the functionally equivalent `_thread.allocate_lock()` and remove the `threading` import.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -1,6 +1,7 @@
 import math
 import types
 
+from _thread import allocate_lock
 from collections import Counter, defaultdict, deque
 from collections.abc import Sequence
 from contextlib import suppress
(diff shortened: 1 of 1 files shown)


jonathandung:
I just realized that the `queue` is imported already, which loads `threading` rather than `_thread`. I will repurpose this PR to make `queue` a [PEP 810](https://peps.python.org/pep-0810/) lazy import as well, which will cause `threading` to only be imported when the deprecated `callback_iter` class is instantiated.

bbayles (approved):

Issue #1180, opened 2026-06-17
`groupby_transform` overloads does not cover all calling patterns


jonathandung:
#568 added overloads for `groupby_transform`, but did not account for the cases where some arguments were not passed.

PR #1181, opened 2026-06-17
Completely type `groupby_transform`

jonathandung:
### Issue reference
more-itertools/more-itertools#1180

### Changes
Add overloads for `groupby_transform` to cover cases where some parameters are not passed


bbayles:
I'll accept that this is more correct, but at a certain point, what are we doing with these type annotations?

PR #1182, opened 2026-06-17
Support negative start/stop in iter_index for general iterables

gaoflow:
### Problem

`iter_index` accepts negative `start` / `stop` for sequences (the fast `.index`
path) but crashes for general iterables (the slow `islice` path):

```python
from more_itertools import iter_index

list(iter_index('AABCADEAF', 'A', -3))        # [7]
list(iter_index(iter('AABCADEAF'), 'A', -3))  # ValueError: Indices for islice() must be None or an integer...
```

### Cause

`iter_index` has two code paths:

- **Fast path** (objects with `.index`, e.g. `str`/`list`): negatives work via
  `str.index` / `list.index`, with the usual from-the-end semantics.
- **Slow path** (general iterables): `islice(iterable, start, stop)` rejects
  negative indices, so any negative `start`/`stop` raises.

### Why it's a bug (oracle)

`IterIndexTests` runs every case `for wrapper in (list, iter)` and asserts the
two paths produce identical output — i.e. the maintained invariant is that the
sequence and general-iterable paths agree. Negative bounds break that invariant:
the `list` wrapper yields a result, the `iter` wrapper raises.

### Fix

When `start` or `stop` is negative on the slow path, materialize the iterable to
a tuple and use its `.index`, so both paths share the same from-the-end
semantics. Materialization only happens on the (rare) negative-index path; the
common non-negative slow path stays fully lazy and is untouched.

```python
seq_index = getattr(iterable, 'index', None)
if seq_index is None and (start < 0 or (stop is not None and stop < 0)):
    iterable = tuple(iterable)
    seq_index = iterable.index
```

### Tests

Added `IterIndexTests.test_negative_start_and_stop`, covering negative `start`,
negative `stop`, and combined negative bounds, each checked for both the `list`
and `iter` wrappers (matching the existing test style). Verified red before the
fix (the four cases raise) and green after; full suite `891 passed`. A
differential grid (sequences x values x `start`/`stop` in `-15..15`) shows zero
remaining fast-vs-slow disagreements.

---

*Disclosure: I prepared this patch with the help of an AI assistant, under my
direction. I found the inconsistency via fast-vs-slow differential testing,
confirmed the oracle against the existing tests, and verified the fix
red-to-green myself.*

### DIFF
--- a/docs/versions.rst
+++ b/docs/versions.rst
@@ -5,6 +5,12 @@ Version History
 .. automodule:: more_itertools
    :noindex:
 
+Unreleased
+----------
+
+* Changes to existing functions:
+    * :func:`iter_index` was fixed to accept negative *start* and *stop* with general iterables (thanks to gaoflow)
+
 11.1.0
 ------
 
--- a/more_itertools/recipes.py
+++ b/more_itertools/recipes.py
@@ -889,6 +889,13 @@ def iter_index(iterable, value, start=0, stop=None):
 
     """
     seq_index = getattr(iterable, 'index', None)
+    if seq_index is None and (start < 0 or (stop is not None and stop < 0)):
+        # islice() rejects negative indices, but the fast path (below) accepts
+        # them with the usual from-the-end semantics. Materialize so that both
+        # paths agree for negative *start* / *stop*.
+        iterable = tuple(iterable)
+        seq_index = iterable.index
+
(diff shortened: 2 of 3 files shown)


bbayles:
Thanks for the fix. I'll add the changes to the versions doc when it's time for a release.

PR #1183, opened 2026-06-18
Optimize `minmax` to avoid creating an intermediate tuple of positional arguments

jonathandung:
### Changes

If a lot of positional arguments are passed to `minmax`, all but the first are collected into a tuple, which then has the first prepended to it. This incurs the extra cost of a tuple creation and concatenation. This can be avoided by changing the signature.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -4830,8 +4830,10 @@ def minmax(*iterable, key=None, default=_marker):
     `recipe <https://code.activestate.com/recipes/577916-fast-minmax-function>`__ by
     Raymond Hettinger.
     """
-    if not iterable: raise TypeError('`minmax()` expected one or more positional arguments.')
-    if len(iterable) == 1: iterable = iterable[0]
+    if not iterable:
+        raise TypeError('`minmax()` expected one or more positional arguments.')
+    if len(iterable) == 1:
+        iterable = iterable[0]
 
     it = iter(iterable)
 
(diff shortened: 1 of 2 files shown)


pochmann:
Please show benchmark / results that show that this is indeed an optimization.

jonathandung:
```python
import random, timeit
a = list(range(1000000))
random.shuffle(a)
print(timeit.timeit(f'minmax({', '.join(map(str, a))})', 'from more_itertools import minmax', number=10))
```

The above snippet materializes a string passing in 1 million positional arguments to `minmax`. I ran it on the master branch, checked out to my branch, and ran it there.

Without optimization: 0.6787999s/run
With optimization: 0.4586191s/run

Output of `python -VV`: `Python 3.14.6 (tags/v3.14.6:c63aec6, Jun 10 2026, 10:26:10) [MSC v.1944 64 bit (AMD64)]`
Output of `python -m platform`: `Windows-11-10.0.26200-SP0`

This appears to be a significant optimization, but nobody would realistically star an iterable of that many items when the iterable could simply be passed in as the sole argument, hence my calling this a micro-optimization in the last commit.

pochmann:
Do you also see an improvement with something realistic? And how does it affect *common* cases? Does it at least not make them slower?

jonathandung:
```bash
python -m timeit -s "from more_itertools import minmax; from random import shuffle; a = list(range(100)); shuffle(a)" "minmax(a, key=hash)"
```

gives 5.12 microseconds on master and 5.13 microseconds on my branch.

jonathandung:
As to whether I see an improvement with something realistic, I've already said that this is a micro-optimzation, the likes of which I've seen merged.

bbayles:
Passing on this one, but thanks. Rule of thumb: either 10x faster or 10x more beautiful.

Issue #1179, opened 2026-06-16
Incorrect annotation for `initial` in second overload of `difference`


jonathandung:
Since `difference` is meant to be the inverse of `accumulate` (given a suitable `func`), and `difference` only checks if `initial` is `None` to decide whether to yield the first item, I was wondering if a better approach could be used (defaulting `initial` to `_marker` to indicate that it was not passed, because `None` is sometimes a valid value, and raising `ValueError` in the case that `initial` was passed and not equal to the first item).

JamesParrott:
Currently `initial` might as well be  bool.  I don't see there is much value in validation so without knowing more about the potential use cases it's -1 from me.

```py
    a, b = tee(iterable)
    try:
        first = [next(b)]
    except StopIteration:
        return iter([])

    if initial is not None:
        return map(func, b, a)

    return chain(first, map(func, b, a))
```
https://github.com/more-itertools/more-itertools/blob/c5538f9ff7f3fa92ac0fe378b5c70486494c040b/more_itertools/more.py#L2868

jonathandung:
I see why passing `initial=True` would be useful, but it may be detrimental (in that users would pass `initial=False` and the same effect would be unexpectedly achieved) for users who don't read the documentation thoroughly. Also, the stubs [disagree](https://github.com/more-itertools/more-itertools/blob/c5538f9ff7f3fa92ac0fe378b5c70486494c040b/more_itertools/more.pyi#L561) with this calling pattern.

> because `None` is sometimes a valid value

in keeping the existing behaviour, this concern also goes unaddressed.

jonathandung:
If this change still doesn't go through, I'm happy to edit the stubs to reflect that.

bbayles:
`initial` probably should have been a `bool`, but I don't think we need to change it just to change it.

If the types are wrong, let's fix them. I'll accept a PR that makes them more accurate.

For `_marker`, we will eventually have [sentinel](https://docs.python.org/3.15/library/functions.html#sentinel).

jonathandung:
Original:

```python
@overload
def difference(
    iterable: Iterable[_T],
    func: Callable[[_T, _T], _U] = ...,
    *,
    initial: None = ...,
) -> Iterator[_T | _U]: ...
@overload
def difference(
    iterable: Iterable[_T], func: Callable[[_T, _T], _U] = ..., *, initial: _U
) -> Iterator[_U]: ...
```

[...]

PR #1187, opened 2026-06-19
Improve difference typing for initial

gaoflow:
Refs #1179

## Summary
- add an overload for `difference(..., initial=...)` when the default function is used
- type explicit-function `initial` values as the accumulated input item type instead of the callback result type
- document the typing improvement in the unreleased notes

## Tests
- `uv run --no-project --with mypy python -m mypy -c 'from more_itertools import difference\nfrom typing import Iterator, assert_type\nassert_type(difference([10, 11, 13], initial=True), Iterator[int])'`\n- `uv run --no-project --with mypy mypy --strict more_itertools/more.pyi`\n- `PYTHONPATH=. uv run --no-project --with mypy stubtest more_itertools.more more_itertools.recipes`\n- `uv run --no-project python -m unittest -v`\n- `uv run --no-project --with ruff ruff check more_itertools tests`\n- `uv run --no-project --with ruff ruff format --check .`\n- `git diff --check`\n\nAI assistance was used under my direction.

### DIFF
--- a/docs/versions.rst
+++ b/docs/versions.rst
@@ -10,6 +10,7 @@ Unreleased
 
 * Changes to existing functions:
     * :func:`iter_index` was fixed to accept negative *start* and *stop* with general iterables (thanks to gaoflow)
+    * The type hints for :func:`difference` were improved for calls with *initial* set.
 
 11.1.0
 ------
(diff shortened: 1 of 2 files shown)


bbayles:
Somehow I'm skeptical of the 67 PRs in 58 repositories from this user
image

PR #1190, opened 2026-06-23
Add map_keys, map_values, filter_keys, filter_values

r1cksync:
## Summary

This PR adds four small, composable utilities for transforming and filtering dict-like **mappings**, complementing the existing `join_mappings` tool. They cover an extremely common need (reshaping a dict) with a readable, named API instead of an ad-hoc dict comprehension.

| Function | Description |
| --- | --- |
| `map_values(func, mapping)` | Return a new dict with `func` applied to each **value**; keys unchanged. |
| `map_keys(func, mapping)` | Return a new dict with `func` applied to each **key**; values unchanged. Raises `ValueError` on a key collision so a value is never silently dropped. |
| `filter_values(pred, mapping)` | Return a new dict keeping only items whose **value** satisfies `pred` (defaults to a truth test). |
| `filter_keys(pred, mapping)` | Return a new dict keeping only items whose **key** satisfies `pred` (defaults to a truth test). |

### Use cases
- `map_values(str.upper, d)` / `map_values(len, d)` — normalize or summarize values without touching keys.
- `map_keys(str.lower, headers)` — canonicalize keys; the collision check turns a silent data-loss bug into an explicit error.
- `filter_values(None, d)` — drop entries with empty/falsy values (e.g. cleaning a config dict).
- `filter_keys(lambda k: k.startswith("x_"), d)` — select a namespaced subset of a mapping.

These mirror the `map`/`filter` pairing already familiar from the rest of the library and sit naturally next to `join_mappings` in the *Summarizing* section.

## Checklist
- [x] Implementations in `more_itertools/more.py` with complete docstrings and runnable doctest examples
- [x] Added to `__all__`
- [x] Type stubs added to `more_itertools/more.pyi` with precise annotations (`stubtest` passes)
- [x] Documented in `docs/api.rst` and the `README.rst` feature table
- [x] Dedicated `TestCase` per function in `tests/test_more.py` (normal, empty, single-element, error, and identity/new-dict cases)
- [x] `ruff format --check` and `ruff check` clean
- [x] `sphinx-build -W` builds with no warnings
- [x] Full `tests/test_more.py` suite passes (607 tests)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -79,6 +79,10 @@ Python iterables.
 |                        | `run_length <https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.run_length>`_,                                                                         |
 |                        | `map_reduce <https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.map_reduce>`_,                                                                         |
 |                        | `join_mappings <https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.join_mappings>`_,                                                                   |
+|                        | `map_keys <https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.map_keys>`_,                                                                             |
+|                        | `map_values <https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.map_values>`_,                                                                         |
+|                        | `filter_keys <https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.filter_keys>`_,                                                                       |
+|                        | `filter_values <https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.filter_values>`_,                                                                   |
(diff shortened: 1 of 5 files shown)


bbayles:
Thanks for the PR. FYI, the Pull Request template for this repo has this text.
```
### Issue reference
```

I would have liked if you could have made an issue for these proposals.

My current inclination is against adding these new functions, as I think one-line dict comprehensions are what most people use / would want to use. I can see the point of `map_keys` not silently swallowing duplicates, however.

Can you point to some real code on GitHub or StackOverflow that would look better by using these functions instead of using comprehensions?

bbayles:
Deferred per my [June 23rd comment](https://github.com/more-itertools/more-itertools/pull/1190#issuecomment-4781356913).

PR #1196, opened 2026-07-01
speed up `nth_permutation` by binary splitting for int->factoradic and Fenwick tree for map(pool.pop,result)

DroneBetter:
### Changes
insertionsort can be seen as a variant of mergesort, that repeatedly merges a list of length `k` with a list of length `1` instead of another of length `k`, and it takes longer because the time to merge is proportional to the sum of the lengths, making sorting amenable to divide-and-conquering.

likewise, it's faster to multiply numbers together that are about the same size than disparate (especially with Python's `int.__mul__` using Karatsuba).

you can compute `math.factorial` with binary splitting by `8! = ((1*2)*(3*4))*((5*6)*(7*8))`. (actually Python does it more cleverly `16! = (16)*(8)*(4*12)*(2*6*10*14)*(1*3*5*7*9*11*13*15) = 1*1*(1*3)*((1*3)*(5*7))*(((1*3)*(5*7))*((9*11)*(13*15)))<<15` , where the first half of each multiplicand has already been computed; see [the preamble to `factorial_partial_product`](https://github.com/python/cpython/blob/a5be0d81cb74fabe563b8317f94d80aa01c9bbf9/Modules/mathintegermodule.c#L493)).

`nth_permutation` converts the index `i` into a factoradic (before using that as a sequence of indexes to `pop` from the `pool`); this PR does this conversion by divide-and-conquering in reverse, with `*` replaced with `divmod`.

further, repeatedly `pop`ing from within the list (and shifting everything after the `pop` location) is asymptotically slow (quadratic time, since each `pop`); this PR sets up a Fenwick tree, for which setting a value to 0 (so it no longer contributes to the sum) and finding the first location at which the partial sum exceeds a given value are both logarithmic-time, so the total becomes [linearithmic](https://en.wiktionary.org/wiki/linearithmic). (This only takes effect when the input iterable's length is greater than a threshold of `n>=10**5`, since it takes a while to become faster.)

### Checks and tests
all pass, although I had to add an exception for `nth_permutation` to output `()` when `n=0` to pass the empty case. It no longer multiplies the index by `factorial(n)//c` to bring it from `∈ range(perm(n,r))` to `∈ range(factorial(n))`, so the library's only use of `math.factorial` is now gone; when I ran `make all-checks`, ruff wanted me to remove it so I have.

#### Performance
In a little informal test, for getting permutations of size-`500` subsets of a list of `10**3` elements, the first change (binary splitting) is about thrice as fast (from `65536` tests).
```py
from more_itertools.more_itertools import nth_permutation as go
from cool_itertools.more_itertools import nth_permutation as gn
from math import perm
from random import randrange
from time import time
n=10**3
r=n>>1
l=1<<10
tests=[randrange(perm(n,r)) for _ in range(l)]
t=time()
h=hash(tuple(go(range(n),r,i) for i in tests))
print(res:=-t+(t:=time()),res/l)
print(h==hash(tuple(gn(range(n),r,i) for i in tests)),res/(res:=time()-t),'times faster')
print(res,res/l)
```
returns
```
1.3155200481414795 0.0012846875470131636
True 3.069992288424068 times faster
0.4285092353820801 0.0004184660501778126
```
and (with `r=n>>1`) this grows to about 9.5 times faster for `n=10**4`, and about 18 times faster for `n=10**5`; it seems experimentally to shave off a factor of between n^(1/5) and n^(1/3) from the complexity exponent, depending on size. (More is probably possible pretty easily (in particular by changing the splitting from a binary tree to a "Stirling's-approximation-aware" tree to ensure the parts are somewhat more evenly spaced), but this seems a pretty good time/complexity tradeoff for now.)

meanwhile, (with `r=n>>1`) the Fenwick tree change (vs. just the splitting) breaks even at about `n=10**5` (where it's about 1.4 times as fast) and is much faster at about `n=10**6` (where this increases to about 3.3).

(I strongly recommend the first part even if the second is deemed too complicated for too little payoff)


DroneBetter:
btw I am also willing to expand it into a method (together with a `.index`) as part of a reimplementation of `itertools.permutations` into a full-fledged class here, which I have done in [my library](https://github.com/DroneBetter/dronery) already (as well as for `combinations` and `combinations_with_replacement`); would this be worth an issue or subsequent PR?

(I mentioned the possibility of doing this as part of itertools in [this discuss.python.org thread](https://discuss.python.org/t/making-combinations-combinations-with-replacement-and-permutations-indexable/107834) to which the consensus seemed to be that it was lacking motivation, a friend pointed me here.)

DroneBetter:
also, would there be any interest at all in making the aforementioned change to choose the split locations to equidistribute length as best as possible at each iteration, or is this good enough? (note that currently the divmodding step accounts for less than half the total time until about `n=1<<16`, with `r=n>>1`)

DroneBetter:
I added the `fenwick` class from my library in full here to make its invocations in `nth_permutation` clearer; I will use it in `permutation_index` also.

the changes from last version are:
* popping is faster than fenwicking at small scales, now also large ones too; there are multiple smaller sub-pool lists that are popped from, with their remaining lengths tracked by a Fenwick tree. (this reduces the threshold at which fenwick mode is faster to somewhere between 32768 and 65536.)
* if a number is short enough, the overhead of unbalanced divmods is negligible, so for inputs beneath `2**128` just do the original loop
* we need to approximately solve the equation `n!/(n-y)! = (n-y)!/(n-x)!` (ie. `(n-y)! = √(n!*(n-x)!)`) for the bisections to split the interval into a quotient and remainder whose bit_lengths are about equal; using `y = x/2 - (x/2)**2/(2*(n-x/2)*log(n-x/2))` (by using Stirling's approximation (and throwing away the √n) then taking the logarithm, truncating series expansions (around the point `x/2`) and solving the resulting equation exactly) instead of just `x/2` like before produces a better upper bound on the true best value of `y`; see Desmos,
image

all the constants used as thresholds are based on what seemed to work best on my laptop, but I was unthorough and test results are somewhat variable depending on memory pressure so they may be somewhat optimisable further.

profiling script (all accumulated changes vs. the original version):
```python
from more_itertools.more_itertools.more import nth_permutation as go
from cool_itertools.more_itertools.more import nth_permutation as gn
from dronery import fenwick

from math import perm
from random import randrange
from time import time
for i in range(1,18):
    n=1<<i
    r=n>>1
    l=1<<17
    tests=[randrange(perm(n,r)) for _ in range(l)]
    t=time()
    h=hash(tuple(go(range(n),r,i) for i in tests))
    res=-t+(t:=time())
    hh=hash(tuple(gn(range(n),r,i) for i in tests))
    rres=-t+time()
    assert(h==hh)
    print(f'n = {n}, r = {r}, {int(res/rres*1000)/1000} times faster')
```
results:
```
#131072 tests per entry
n = 2, r = 1, 0.755 times faster
n = 4, r = 2, 0.885 times faster
n = 8, r = 4, 1.15 times faster
n = 16, r = 8, 1.37 times faster
n = 32, r = 16, 1.595 times faster
n = 64, r = 32, 1.574 times faster
n = 128, r = 64, 1.616 times faster
n = 256, r = 128, 2.11 times faster
n = 512, r = 256, 2.473 times faster
#from here just 1024 tests per entry
n = 1024, r = 512, 3.747 times faster
n = 2048, r = 1024, 4.341 times faster
n = 4096, r = 2048, 6.308 times faster
n = 8192, r = 4096, 12.98 times faster
n = 16384, r = 8192, 13.403 times faster
n = 32768, r = 16384, 17.25 times faster
n = 65536, r = 32768, 21.534 times faster
```
and that was as far as it got overnight

more_itertools/more.py
```diff
@@ -4287,39 +4360,39 @@ def nth_permutation(iterable, r, index):
...
-                split[2 * i + 1] = d
-
-    if n < 10**5:  # determined experimentally
+    l = r.bit_length()
+
+    def rec(index, a, n):
+        if index.bit_length() < 128:
+            index = (index, 0)
+            return [
+                (index := divmod(index[0], u))[1]
```
SuffixAutomata: I'm fond of this line of code

DroneBetter: (you see darcy(the creature who commented) is also the one who proposed that)

JamesParrott:
I'm not saying it's worth writing tests for this.  I'm just pointing out the existing tests haven't even been run against this in CI, let alone new tests being written, and the test suite fails due to coverage being < 97%.  That's a sign of how complicated  this new code is (more complicated than I even want to review, personally).

But FYI, please can you enable Github Actions on your forks?  Then, at the very least we can see the results of the test suite on  PRs, without a maintainer having to approve them to be run on the parent project?  Github still provide free runner-minutes for public repos AFAIK (and several hundred for your private ones).   

Otherwise, if one of us doesn't run them:
https://github.com/JamesParrott/more-itertools/actions/runs/29725699457
it adds one more task in Bo's TODO list, instead of making the maintainer's life easier.

JamesParrott (changes requested):
Perhaps it's not worth writing tests for this, if this PR is not going to be accepted for some other reason.

But in future, please run the tests somewhere where we can all see the results.  This is easy and free for public repos and forks, by enabling Github Actions in your fork.

I made a branch in my fork based on this PR and while the tests all seem to pass, coverage drops below 97%, causing a secondary failure:

https://github.com/JamesParrott/more-itertools/actions/runs/29725699457

DroneBetter:
i have fenwickectomied it so the only changes are how it turns integers into factoradics (n = 65536 was the threshold i set to start using it, so all tests n ≤ 32768 from my previous comment still hold).

i have turned on Github Actions and it passes everything (cf. [the checks in my branch](https://github.com/DroneBetter/more-itertools/actions/runs/29741186511)); however, all test cases are too small and go through the `if index.bit_length() < 128` shortcut so the fun part doesn't get invoked.

JamesParrott:
> however, all test cases are too small and go through the if index.bit_length() < 128 shortcut so the fun part doesn't get invoked.

Add some bigger test cases then:  https://github.com/DroneBetter/more-itertools/blob/1e94975d54e55dd37d53e1c5d086ad773d32f7f5/tests/test_recipes.py#L745
Show everybody what happens when the fun part does get invoked.  

It's good if you're nervous about modifying the test suites (as that could be a way of cheating the tests).  But leaving the existing ones alone and adding more test cases is healthy, and should be encouraged.  Even when the PR doesn't delete the entire current tested implementation, and replace it with a new one.

If it's not worth testing, it's not worth adding to a stable Open Source library with existing high test coverage, IMHO.

I wasn't insisting you delete the class you were fond of by the way.  But now that you've done so anyway, personally I do find it much easier to read, so thanks.

bbayles:
What's the expected performance curve with the latest code?

Assuming it's stellar, we will want the test suite to exercise the new part.

pochmann:
@bbayles Why is the current code doing the following anyway?

https://github.com/more-itertools/more-itertools/blob/3a259350f28cd17f09b4f0af21c6ebc03cf725cf/more_itertools/more.py#L4308

Since `c = perm(n, r)`, isn't `factorial(n) // c` the same as `factorial(n - r)`?

Can't that whole line be replaced with this?

```python
    q = index * factorial(n - r)
```

pochmann:
After I tried my above simplification, I noticed that even that multiplication with `factorial(n - r)` was pointless. The loop after it divided that away anyway, doing nothing with the remainders. So my linked PR removes all that. With this PR's benchmark, I also get a >3x speedup. That's for `n=10**3`. This PR still seems to be a bit faster, though, and more so for larger `n`. So it can still make sense even if my PR gets merged, but its advantage shrinks (smaller speedup, and  larger code complication).

DroneBetter:
oops i forgot to rerun ruff after that last edit (which is why it is upset on 3.10 but not 3.11 or 3.13 or 3.14 for some reason)

anyway yes here is the last performance curve i did up to 1024 (after changing the threshold to begin bisecting [currently on line 4296](https://github.com/DroneBetter/more-itertools/blob/de96dff304bf670253c76b9f9f6a31e57c982ecc/more_itertools/more.py#L4296) from 512 to 1024)
image
image
the blue line is the PR in its current state, the orange is what if `_perm_rec` didn't exist and no bisection happened at all

the reason why blue isn't much better than orange for `permutation_index` is that converting the permutation into factoradic form is much more expensive than converting the factoradic into an integer, and so most of the gains come from using an integer as a bitset

DroneBetter:
~~wait sorry i will make one comparing this against the present version with pochmann's smaller change~~ update: pochmann's change turns the blue line into the green line for `nth_permutation`.

at `n=1024`, this PR is about 1.45 times faster for `nth_permutation` and remains >7 times faster than the current implementation for `permutation_index`.

i would like a bit to readd the `fenwick` class from my first commits (albeit as a primarily-user-facing thing described in the docs) and then incorporate it into the permutation-factoradic conversion again for this PR, should I make a PR for that or an issue first detailing what it will contain?

bbayles:
My rule of thumb is "if it's not 10x faster, it needs to be 10x more elegant" for changing something that currently works. I don't think we've met either bar yet, but I'm open to reviewing more implementations.

If we don't wind up merging the code, I'm happy to link a writeup about alternate methods in the docs.

PR #1197, opened 2026-07-01
Use `divmod` for `nth_product` instead of computing quotient and remainder separately

jonathandung:
### Changes

Essentially inspired by the implementation of `nth_permutation`, which updates the variable `q` using `divmod`.


bbayles:
This 2.5% worse in terms of performance on my machine, alas.

jonathandung:
I suppose it's because of the extra tuple assignment to `rem`. I'll close this.

Issue #1198, opened 2026-07-01
Is there any reason why the documentation for `pairwise` and `callback_iter` doesn't use the dedicated ReST `deprecated` directive?


jonathandung:

JamesParrott:
Because it's neither here not there, and we all wish to avoid busywork?  The warnings are perfectly clear, and make sense in the docstrings too.

bbayles:
What's it look like? If you can generate the docs with it to demonstrate, I'm open to using it.

karlhillx:
Opened #1202 — swaps both  blocks for .

jonathandung:
I've already opened one previously.

PR #1199, opened 2026-07-02
Use `.. deprecated::` directive for `callback_iter` and `pairwise`

jonathandung:
### Issue reference
more-itertools/more-itertools#1198


jonathandung:
It would be great if Read the Docs PR builds were enabled.

bbayles (approved):

Issue #1204, opened 2026-07-06
Document how to accomplish functionality equivalent to rust's `itertools::Itertools::duplicates`


jonathandung:
### Description
This itertool would yield all duplicate items in an iterable, giving every duplicate exactly once. It would possibly also take a `key` parameter to determine if items are indeed duplicates.

### References
Inspired by the rust `itertools` crate, which [provides this as an iterator adaptor](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.duplicates).

### Examples

```pycon
>>> list(duplicates([5, 11, 24, 35, 23, 42, 11, 56, 19, 18, 27, 27], key=10 .__rfloordiv__))
[23, 11]
```

jonathandung:
## How is this different than `duplicates_justseen` or `duplicates_everseen`?

As the above example shows, only the second occurrence of the duplicate item is yielded once it appears, or if `key` is passed, all elements in the returned iterable would map to different keys. (This is why the return value is not `[23, 11, 19, 18, 27, 27]`, as it would be under `duplicates_everseen`, or `[18, 27]` as in `duplicates_justseen`.)

## Additional remarks

The `key` parameter should be passable by position or keyword, mirroring the behaviour of other functions in this library. The items in the iterable should also be hashable, because the implementation I have in mind uses a dictionary.

jonathandung:
I would like to ask here if I am expected to write tests, and how comprehensive they should be.

jonathandung:
~~I just read through the tests directory, and it appears that my doctest is sufficient.~~

I made this remark since I recall having found no tests for at least two recipes I chose at random, but that appears to have been fixed (or I remembered wrong). I will add tests tomorrow if there is more feedback to determine the fate of my PR.

JamesParrott:
Hi Jonathan.  Have you been told this feature request is accepted?  The unique items are being stored in the dict so personally I'd use a `collections.Counter` (or if using  a key,  `defaultdict(list)`) for this.  `duplicates` could be of use though, when there are so many repeated items it is not desirable to cache them all.

If you want to plough on regardless, at risk of rejection, then I liked the readability of `duplicated[i] = False`, and other than the `.get`, it's essentially the same code a lot of users would write for themselves.  But:
 -  firstly, do users really expect to see the second duplicated item, not the first one (or last)?
[...]

PR #1205, opened 2026-07-06
Implement `duplicates`

jonathandung:
### Issue reference
more-itertools/more-itertools#1204


bbayles:
Deferred - see the comment [here](https://github.com/more-itertools/more-itertools/issues/1204#issuecomment-5132087130). Thanks for the PR all the same.

PR #1209, opened 2026-07-12
Raise ValueError for invalid group length in grouper

vidigoat:
### Changes

`grouper` silently returns an empty iterator when given a group length below one:

```pycon
>>> list(grouper('ABCDEF', -1))
[]
>>> list(grouper('ABCDEF', 0))
[]
```

This happens because `[iter(iterable)] * n` is an empty list for any `n < 1`, and `zip`/`zip_longest` over no iterators yields nothing. The invalid argument is swallowed rather than reported.

Its sibling recipe `batched` already raises `ValueError` for `n < 1`:

```pycon
>>> list(batched('ABCDEF', -1))
ValueError: n must be at least one
```

This PR makes `grouper` do the same, so an unusable group length is signalled instead of quietly producing wrong output. The valid path is unchanged. Added a regression test and a changelog entry.

### Checks and tests

`make check` (ruff format/check, stubtest) and the full `unittest` suite pass locally; coverage on `recipes.py` remains at 100%.

### DIFF
--- a/docs/versions.rst
+++ b/docs/versions.rst
@@ -10,6 +10,7 @@ Unreleased
 
 * Changes to existing functions:
     * :func:`iter_index` was fixed to accept negative *start* and *stop* with general iterables (thanks to gaoflow)
+    * :func:`grouper` now raises ``ValueError`` for a group length below one, matching :func:`batched`, instead of silently returning an empty iterator (thanks to vidigoat)
 
 11.1.0
 ------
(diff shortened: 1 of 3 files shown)


bbayles:
I'd like it if you could follow these instructions when opening PRs:

```
### Issue reference
more-itertools/more-itertools#XXXX
```

rhettinger:
I don't think this is worth it.  As the OP points out, swallowing a negative repeat value is common in Python. `[0] * -5` returns an empty list.  The `repeat()` itertool also intensionally swallows negative inputs.

Unless the OP can show a real-world case where grouper raising an exception would have been helpful, I would drop this as a useless edit.

Note the test for `batched` is really checking for the zero case as is not primarily targeted at negative inputs.

PR #1212, opened 2026-07-16
Fix iter_index crashing on range inputs

chuenchen309:
## Problem

`iter_index` chooses its sequence fast path whenever the iterable has an `.index()` method:

```python
seq_index = getattr(iterable, 'index', None)
...
yield (i := seq_index(value, i + 1, stop))   # always called with start/stop
```

`range` *has* an `.index()` method, but unlike `str`/`bytes`/`list`/`tuple` it accepts only the value — no `start`/`stop`. So the fast path crashes:

```python
>>> from more_itertools import iter_index
>>> list(iter_index(range(10), 5))
TypeError: range.index() takes exactly one argument (3 given)
```

This happens even on the simplest call (default `start=0`, `stop=None` → `stop` becomes `len(iterable)` and is still passed as a third argument). `range` is an ordinary iterable and the docstring contract ("yield the index of each place in *iterable* that *value* occurs") gives an expected result of `[5]`.

## Fix

Route `range` through the general-iterable path instead of the sequence fast path. That path already produces correct results for `range` (and already materializes to a `tuple` for negative `start`/`stop`, so `tuple.index`'s 3-arg form is used there). `range` is the one stdlib sequence whose `.index()` can't drive the fast path.

## Test plan

Added `IterIndexTests.test_range` covering positive/stepped ranges, `start`/`stop` (incl. negative), and a not-found value:

```
range(10), 5            -> [5]
range(0, 20, 2), 4      -> [2]
range(10), 3, 1, 8      -> [3]
range(10), 100          -> []
range(10), 5, start=-6  -> [5]
range(20), 5, stop=-2   -> [5]
```

- On `master` every non-empty `range` case raises `TypeError`; with this change they pass and the `str`/`list`/`tuple` fast path is unchanged.
- `pytest tests/` → 733 passed (+1), 19944 subtests.
- `ruff check` / `ruff format --check` clean.
- Added an Unreleased entry to `docs/versions.rst`.

---
Authored with assistance from Claude Code (Anthropic); the bug, fix, and test were reviewed and verified locally by the contributor.

### DIFF
--- a/docs/versions.rst
+++ b/docs/versions.rst
@@ -10,6 +10,7 @@ Unreleased
 
 * Changes to existing functions:
     * :func:`iter_index` was fixed to accept negative *start* and *stop* with general iterables (thanks to gaoflow)
+    * :func:`iter_index` no longer raises ``TypeError`` on :class:`range` inputs (thanks to chuenchen309)
 
 11.1.0
 ------
--- a/more_itertools/recipes.py
+++ b/more_itertools/recipes.py
@@ -889,21 +889,11 @@ def iter_index(iterable, value, start=0, stop=None):
     associated with particular values.
 
     """
-    if isinstance(iterable, range):
-        # range.index() takes only the value -- no start/stop -- so it cannot
-        # drive the fast path below. It does locate the value arithmetically,
-        # and a range never repeats one, so at most one index can match and
-        # finding it needs no scan.
-        try:
-            i = iterable.index(value)
-        except ValueError:
-            return
-        lo, hi, _ = slice(start, stop).indices(len(iterable))
-        if lo <= i < hi:
-            yield i
-        return
-
     seq_index = getattr(iterable, 'index', None)
+    if isinstance(iterable, range):
+        # The fast path resumes past each hit with index(value, i, stop), but
+        # range.index() takes only the value and reports just the first match.
+        seq_index = None
     if seq_index is None and (start < 0 or (stop is not None and stop < 0)):
         # islice() rejects negative indices, but the fast path (below) accepts
         # them with the usual from-the-end semantics. Materialize so that both
--- a/more_itertools/recipes.py
+++ b/more_itertools/recipes.py
@@ -890,6 +890,11 @@ def iter_index(iterable, value, start=0, stop=None):
 
     """
     seq_index = getattr(iterable, 'index', None)
+    if isinstance(iterable, range):
+        # range has an index() method, but unlike other sequences it accepts
+        # only the value (no start/stop), so it cannot drive the fast path
+        # below. Treat it as a general iterable instead.
+        seq_index = None
     if seq_index is None and (start < 0 or (stop is not None and stop < 0)):
         # islice() rejects negative indices, but the fast path (below) accepts
         # them with the usual from-the-end semantics. Materialize so that both
--- a/more_itertools/recipes.py
+++ b/more_itertools/recipes.py
@@ -889,12 +889,21 @@ def iter_index(iterable, value, start=0, stop=None):
     associated with particular values.
 
     """
-    seq_index = getattr(iterable, 'index', None)
     if isinstance(iterable, range):
-        # range has an index() method, but unlike other sequences it accepts
-        # only the value (no start/stop), so it cannot drive the fast path
-        # below. Treat it as a general iterable instead.
-        seq_index = None
+        # range.index() takes only the value -- no start/stop -- so it cannot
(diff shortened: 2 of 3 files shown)


pochmann:
Let's see if Claude can figure this out: Why not (try to) find the index and yield it if it's between start and stop?

chuenchen309:
You're right, and it's better than what I had. Done in 2447767.

The part I'd missed: **a range never repeats a value**, so there's at most one index to yield — and `range.index()` computes it arithmetically rather than searching. My version threw that away and walked the range element by element, and for negative `start`/`stop` it fell into the branch that materializes a tuple first. On `range(10**9)` that's a scan (and, for the negative case, a tuple of a billion ints) to produce a single index that arithmetic gives immediately:

```python
list(iter_index(range(10**9), 999_999_999))   # [999999999], instant
```

The bounds check is just `slice(start, stop).indices(len(iterable))`, which gives the same clamping and from-the-end semantics the sequence path gets from `list.index(value, start, stop)`.

```python
if isinstance(iterable, range):
    try:
        i = iterable.index(value)
    except ValueError:
        return
    lo, hi, _ = slice(start, stop).indices(len(iterable))
    if lo <= i < hi:
        yield i
    return
```

Two things I checked rather than assumed:

- **`range.index` matches `==` semantics**, so it agrees with the general path on the values where that could plausibly diverge: `5.0`, `True`/`False`, `complex(5, 0)` all find their index; `'5'` and `None` raise `ValueError` where the general path yields nothing.
- **Exhaustive agreement with the existing sequence path** — every `(start, stop)` pair from `-len-2` to `len+2` plus `stop=None`, across `range(10)`, `range(0, 20, 2)`, `range(10, 0, -1)`, `range(0)`, `range(1)`, `range(-5, 5)`, over in-range and out-of-range values: **39486 combinations, zero divergences** against `iter_index(tuple(r), ...)` as the reference.

The test now compares against the equivalent tuple rather than hard-coded expectations, plus a `range(10**9)` case that pins the no-scan property — it times out on the previous commit, which is what I wanted from it. Full suite: 734 passed.

pochmann:
So I guess you're saying (or Claude is saying? Your comment feels like AI) that Claude *can't* figure out why not to do that.

What I had in mind is stuff like this:

```
class Even:
    def __eq__(self, other):
        return not other % 2

print(*iter_index(range(10), Even()))
```

Your earlier version found them all, output was:

```
0 2 4 6 8
```

Your new version only finds the `0`. And with `tuple(range(10))` (and even with `iter(range(10))`) it does find all, so there's a divergence.

chuenchen309:
You're right, and the miss is worse than the diverging output shows.

Your question was "why *not*" and I answered it as if it were "why not just". That's the actual failure here, before any of the code.

The answer I should have given: `.index()` only ever reports the *first* match. The fast path gets away with that only because it loops — it re-calls `seq_index(value, i + 1, stop)` to resume past each hit. `range.index()` takes no `start`, so it can't be resumed; call it once and you get at most one index. That's the same fact that made the fast path raise on `range` in the first place, so it can't be the basis for the fix.

My commit message said "a range never repeats a value, so at most one index can match". That's the error in one line. `iter_index` never compares elements *to each other* — it compares each element to the probe, and equality against an arbitrary probe is the probe's to define. It needn't be injective, or even an equivalence relation. `Even()` equals 0, 2, 4, 6 and 8 at once. Distinct elements bound nothing.

I should also retract something from that comment while I'm at it: the "39486 combinations, zero divergences" was real but blind — every probe I generated was int-like (`5.0`, `True`, `complex(5, 0)`), so a non-injective `__eq__` was never in the input set and the oracle could not have fired. Right oracle, wrong inputs. Your five-line class found it immediately.

Pushed: `range` goes through the general path again, with your `Even()` case as a test (verified red before, green after) and a differential against the `tuple`/`iter` oracles with `Even`/`Always`/`Never` in the probe set this time. Also fixes a `ruff format` failure commit 2 introduced that I never re-checked — CI hasn't run on this PR, so nothing caught it but you.

One thing I'll mention rather than do: an O(1) path is still recoverable if it's gated on `type(value) is int` instead of on the container, since there `==` really is int equality and really is injective. I left it out — this is a crash fix, `master` raises today, so there's no perf regression to defend — but say the word if you'd rather have it.

And to answer the part you didn't quite ask: yes, I use Claude Code for this. You read the tone correctly, and the overclaiming you're picking up on is a real failure mode of how I've been working, not a stylistic quirk.

bbayles:
Superseded by #1214

PR #1194, opened 2026-06-30
Raise for negative tail sizes on sized iterables

CodingFeng101:
## What changed

`tail()` now rejects negative `n` values before choosing the sized or iterator implementation path.

## Why

The iterator path already raises `ValueError` for negative sizes via `deque(..., maxlen=n)`, but the sized-iterable path returned an empty iterator instead. That made `tail(-1, iter(seq))` and `tail(-1, seq)` behave differently for the same invalid argument.

## Validation

- `python -m pytest tests\test_recipes.py::TailTests -q`
- `python -m pytest tests\test_recipes.py -q`
- `python -m ruff check more_itertools\recipes.py tests\test_recipes.py`
- `git diff --check`

PR #1223, opened 2026-07-19
Raise a clear ValueError for negative n in chunked()

uttam12331:
### Issue reference
No issue — this is a small bug fix, in the same spirit as the recently merged negative-size guards for `sliced` (#1200) and `tail` (#1194).

### Changes
`chunked()` was the only sizing helper without an `n < 0` guard. A negative `n` leaked `islice`'s internal message:

```
>>> list(chunked([1, 2, 3], -1))
ValueError: Stop argument for islice() must be None or an integer: 0 <= x <= sys.maxsize
```

This validates `n` up front and raises `ValueError('n must be at least 0')`, matching `sliced()` and `tail()`. `n=None` (single chunk) and `n=0` (empty result) are deliberately preserved and unchanged.

### Checks and tests
Added `ChunkedTests.test_negative`, mirroring `IchunkedTests.test_negative`. `ruff format --check`, `ruff check more_itertools tests`, and the full `tests/test_more.py` suite all pass locally.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -230,6 +230,9 @@ def chunked(iterable, n, strict=False):
     list is yielded.
 
     """
+    if n is not None and n < 0:
+        raise ValueError('n must be at least 0')
+
     iterator = iter(partial(take, n, iter(iterable)), [])
     if strict:
         if n is None:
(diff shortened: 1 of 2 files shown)


JamesParrott:
Uttam, please can you enable Github Actions in your fork, so we can all see the results of the tests on PRs, without the maintainer having to manual approve them to be run on the main project?

Otherwise LGTM.  Tests pass on 3.10-3.14:  https://github.com/JamesParrott/more-itertools/actions/runs/29726881302/job/88301936363

bbayles:
Thanks - this behavior matches `itertools.batched`, which I probably ought to recommend for 3.12+.

PR #1235, opened 2026-08-07
Close the last coverage gaps, fix a no-op equality assertion

darrenhuai:
### Issue reference

No issue - this is tests only, apart from a one-line change to the coverage
gate in the Makefile.

### Changes

Coverage was sitting at 99% with six uncovered lines. Chasing them turned up a
test that wasn't testing anything. `NumericRangeTests.test_eq` had:

```python
# numeric_range objects are equal to themselves
range_1 = mi.numeric_range(2, 4, 1)
self.assertTrue(range_1, range_1)
```

The second argument to `assertTrue` is the failure message, so that line only
checked that the range was truthy - it never called `__eq__` at all, which is
why the identity shortcut in `numeric_range.__eq__` showed as uncovered. If you
make `__eq__` return `False` unconditionally, the old assertion still passes.
Switched it to `assertEqual` and added a second case comparing distinct objects
over the same values.

The other gaps:

`minmax`'s keyed branch compares items pairwise and one of its four arms was
never reached. Rather than adding a single hand-picked case I cross-checked
against `min`/`max` with a key over the short permutations of a fixed set, which
reaches all four. The key is `x % 7` over `1..7` so every key is distinct and
there are no ties to make the comparison ambiguous.

`callback_iter.result` had no test for the window where the function has started
but hasn't returned - `test_no_result` only reaches the case where there's no
future yet, so the `TimeoutError` branch was dead. The new test gates the worker
on an `Event` rather than a sleep, so it isn't a timing race.

`nth_prime` tightens its upper bound from the 688,383rd prime onwards, and the
exact path sieves up to that bound, so tightening it too far would leave the
sieve short and `nth()` would return `None` instead of a prime. Nothing
exercised it. One thing worth flagging: at the threshold the bound clears the
prime it has to reach by 3, so there isn't much room there. I checked every n in
the window where the tightened bound feeds the exact path (688,382 through
1,000,000) and it holds throughout, but it's tighter than I expected. The
`approximate=True` path above 1,000,000 was also completely untested.

That gets it to 100%, so the Makefile gate goes to `--fail-under=100`. Happy to
drop that line if you'd rather keep the gate at 99 and leave headroom.

### Checks and tests

`make all-checks` passes on the branch: 902 tests OK, coverage 100% at
`--fail-under=100`, `ruff format --check` and `ruff check` clean, `stubtest`
clean on both modules.

Since this PR is almost entirely tests, I also checked each new test actually
fails for the right reason rather than just passing - broke each behaviour under
test in turn and confirmed the corresponding test caught it, then reverted. The
`minmax` cross-check fails 511 subtests when that branch is disabled, and the
`nth_prime` bound test fails if the bound is tightened by a further 4.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ format:
 coverage:
 	python -m pip install --upgrade -r requirements/testing.txt
 	coverage run --include="more_itertools/*.py" -m unittest
-	coverage report --show-missing --fail-under=99
+	coverage report --show-missing --fail-under=100
 
 .PHONY: test
 test:
--- a/tests/test_more.py
+++ b/tests/test_more.py
@@ -32,7 +32,7 @@ from pickle import loads, dumps
 from random import Random, choices, random, randrange, seed
 from statistics import mean, median
 from string import ascii_letters
-from threading import Thread, Lock
+from threading import Event, Thread, Lock
 from time import sleep
 from typing import NamedTuple
 from unittest import TestCase, mock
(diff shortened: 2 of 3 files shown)


bbayles:
Thanks. I'm keeping `fail-under` at 99 for now; historically it's been annoying to have it at 100 when there are different code paths for different Python versions.

darrenhuai:
Perfect thank you!

PR #1242, opened 2026-08-30
docs: Clarify output order for sortable input elements

areiche2:
Add note about output order for sortable elements.

### Issue reference

No issue. This is a documentation-only clarification of existing behavior.

### Changes

Document that when the elements of the input iterable are sortable, distinct_permutations produces its output tuples in sorted order.

The current implementation already produces permutations in sorted order when the input elements are sortable. This is a useful property for callers and seems worth making part of the documented behavior. [`itertools.permutations`](https://docs.python.org/3/library/itertools.html#itertools.permutations) explicitly documents its corresponding ordering behavior, so I think this behavior should be documented here as well.

### Checks and tests

Documentation-only change.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -755,6 +755,9 @@ def distinct_permutations(iterable, r=None):
     duplicates are not generated and thrown away. For larger input sequences
     this is much more efficient.
 
+    If the elements of the input iterable are sortable, the output tuples are
+    produced in sorted order.
+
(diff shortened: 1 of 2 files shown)


bbayles:
Could you add a test that ensures that we don't regress on this new promise?

areiche2:
Added test. Confirmed `make all-checks` passed.

bbayles (approved):

PR #1248, opened 2026-09-02
Issue 1247: Keep numeric_range consistent with the items it produces

### LINKED ISSUE #1247: numeric_range: len(), `in` and index() disagree with the items produced, for float steps
`numeric_range` derives `len()`, `__contains__` and `index()` by dividing by the step, but produces its items by repeated multiplication. With floats the two disagree, so a range can report a length that does not match what it yields, and can deny membership to its own items.

On 11.1.0 and on current master (`2fe1b2e`):

### `len()` disagrees with iteration, in both directions

```pycon
>>> from more_itertools import numeric_range

>>> r = numeric_range(-8.732, -13.532, -2.4)
>>> list(r)
[-8.732, -11.132]
>>> len(r)
3
>>> r[2]                      # counted, but it is the excluded stop value
-13.532

>>> r = numeric_range(-69.37276490678778, -29.8339073344664, 4.942357196540172)
>>> len(list(r))
9
>>> len(r)
8
>>> r[8]                      # an item iteration yields, but indexing cannot reach
IndexError: numeric range object index out of range
```

### Items are not `in` their own range

```pycon
>>> r = numeric_range(0.0, 1.0, 0.1)
>>> r[3]
0.30000000000000004
>>> r[3] in r
False
>>> r.index(r[3])
ValueError: 0.30000000000000004 is not in numeric range
>>> sum(1 for v in r if v not in r)
5
```

`count()` follows `__contains__`, so it returns 0 for those items too.

### Cause

`__iter__` and `_get_by_index` compute items as `start + n * step`. Membership and length instead reason about the step arithmetically:

- `__contains__` and `index` test `(value - start) % step == 0`. `0.30000000000000004` is what the range yields at index 3, but it is not an exact multiple of `0.1`, so the remainder is not zero.
- `_len` computes `divmod(distance, step)` and rounds up. That measures the distance to cover rather than counting the items actually produced, and lands one over or one under when the multiplication does not agree with the division.

Exact types are unaffected — `Decimal`, `Fraction`, `int`, and `datetime`/`timedelta` all give a zero remainder for their own items, so they behave the same before and after. This is specific to binary floats.

The `__contains__` shortcut dates to #363 (2020); `_len` has had the same shape since it was introduced.

### Note

This is not only about surprising float repr, which the docstring already warns about. `numeric_range` is a `Sequence`, and here `r[i] not in r` and `len(r) != len(list(r))` — the object disagrees with itself, so `.index()`, `.count()`, `in`, `len()` and indexing give inconsistent answers for the same range.

I have a fix that settles all four against the items the range actually produces, keeping the exact-type behavior and the current performance for them. Happy to open a PR if you would like it.

onk3sh:
### Issue reference

more-itertools/more-itertools#1247

### Changes

`numeric_range` produces its items by repeated multiplication (`start + n * step`, see `__iter__` and `_get_by_index`), but derived `len()`, `__contains__` and `index()` from division by the step. With floats the two disagree:

```pycon
>>> r = numeric_range(0.0, 1.0, 0.1)
>>> r[3]
0.30000000000000004
>>> r[3] in r
False

>>> r = numeric_range(-8.732, -13.532, -2.4)
>>> len(list(r)), len(r)
(2, 3)
```

This settles all of them against the items themselves:

- `_len` still divides to get a starting count, then trims or extends it until it agrees with the values the range produces. The old formula could land one over (counting an item at or past the stop) or one under (missing an item that iteration yields).
- `index()` uses the quotient to locate the value on the grid and then confirms it by comparing against the item at that position, instead of testing the remainder for zero.
- `__contains__` defers to `index()`, and `count()` already defers to `__contains__`, so the four cannot drift apart again.

Exact types are unaffected. `Decimal`, `Fraction`, `int` and `datetime`/`timedelta` produce a zero remainder for their own items, so they resolve on the first candidate and take the same number of operations as before; the adjustment loops in `_len` do not run for them.

Both directions of the `len()` error are in the regression tests, including the harder under-count case, which needs a step whose accumulated product drifts below the divided estimate.

### Checks and tests

`ruff format --check .`, `ruff check more_itertools tests` and `stubtest more_itertools.more more_itertools.recipes` all pass. The suite is 905 tests, all passing, and `coverage report --fail-under=99` reports 100% on `more_itertools/more.py`.

I checked the new tests fail without the change: `test_contains_items_with_inexact_step` fails on membership, and `test_len_matches_iteration` fails with `AssertionError: 2 != 3`.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -2331,14 +2331,11 @@ class numeric_range(Sequence):
             return self._start > self._stop
 
     def __contains__(self, elem):
-        if self._growing:
-            if self._start <= elem < self._stop:
-                return (elem - self._start) % self._step == self._zero
-        else:
-            if self._start >= elem > self._stop:
-                return (self._start - elem) % (-self._step) == self._zero
-
-        return False
+        try:
+            self.index(elem)
+        except ValueError:
+            return False
+        return True
 
     def __eq__(self, other):
         # numeric_range object equality is intended to mirror the built-in range
@@ -2417,7 +2414,20 @@ class numeric_range(Sequence):
             return 0
         else:  # distance > 0 and step > 0: regular euclidean division
             q, r = divmod(distance, step)
-            return int(q) + int(r != self._zero)
+            n = int(q) + int(r != self._zero)
+            # The division above measures the distance to cover, but the items
+            # are produced by repeated multiplication (see `__iter__`). With
+            # inexact arithmetic the two disagree at the boundary, so settle
+            # the count against the items themselves.
+            while n and not self._before_stop(n - 1):
+                n -= 1
+            while self._before_stop(n):
+                n += 1
+            return n
+
+    def _before_stop(self, i):
+        value = self._start + i * self._step
+        return value < self._stop if self._growing else value > self._stop
 
     def __reduce__(self):
         return numeric_range, (self._start, self._stop, self._step)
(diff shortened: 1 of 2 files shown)


bbayles:
Nice work, thank you.

PR #1254, opened 2026-09-04
split_before/split_after/split_when: make maxsplit=0 agree with other values on empty input

### LINKED ISSUE #1252: split_before/split_after/split_when disagree with themselves on an empty iterable when maxsplit=0
`split_before`, `split_after` and `split_when` return a different result for an empty iterable when `maxsplit=0` than they do for any other `maxsplit`, even though no split can occur either way.

```python
>>> pred = lambda x: x == 0
>>> {ms: list(mi.split_before([], pred, maxsplit=ms)) for ms in (-1, 0, 1, 2)}
{-1: [], 0: [[]], 1: [], 2: []}
```

`split_after` and `split_when` behave the same way. `split_at` is consistent (`[[]]` at every `maxsplit`, matching `''.split(',', 0) == ['']`).

The `[]` answer is the one the library already pins: `tests/test_more.py:1548` asserts `list(mi.split_before([], ...)) == []` and `tests/test_more.py:1689` asserts the same for `split_when`. Neither test passes `maxsplit`, so neither reaches the fast path that produces `[[]]`.

The fast path is:

```python
if maxsplit == 0:
    yield list(iterable)
    return
```

added to all four functions in f77823f (2020) as a speed shortcut, when `split_before` still ended with a bare `yield buf` and `[[]]` was its correct empty-input answer. 2e81a56 (2021) later changed the terminal yield to `if buf: yield buf` and added `test_empty_collection`, but did not reach behind the fast path.

Happy to open a PR guarding the fast path the same way the slow path already is, leaving `split_at` alone.

seven7763:
## Summary

Fixes #1252.

With `maxsplit=0` the fast path yielded `[[]]` for an empty iterable, while every other `maxsplit` value yielded `[]` — although no split can happen either way:

```python
>>> pred = lambda x: x == 0
>>> {ms: list(mi.split_before([], pred, maxsplit=ms)) for ms in (-1, 0, 1, 2)}
{-1: [], 0: [[]], 1: [], 2: []}   # before this change
```

The three fast paths now yield nothing for empty input, matching their own slow paths and the expectations already pinned by `test_empty_collection`. `split_at` is untouched — its `[[]]` at every `maxsplit` is self-consistent and mirrors `"".split(",", 0) == [""]`.

## Changes

- `more_itertools/more.py`: `split_before`, `split_after`, `split_when` — the `maxsplit == 0` fast paths now only yield a chunk when the input is non-empty
- `tests/test_more.py`: one regression test per class asserting `[]` for empty input across `maxsplit` in `(-1, 0, 1, 2)`

## Validation

- [x] Reproduced the issue's dict on `master` before the change
- [x] After the change: all three functions return `[]` for every `maxsplit` on empty input; non-empty `maxsplit=0` behavior unchanged (`[[0, 1, 2]]` etc.)
- [x] `python -m pytest tests/ -q` — 745 passed, 21049 subtests passed
- [x] `doctest.testmod(more_itertools)` — 0 failures

## Notes

`split_at` deliberately keeps its current behavior; changing it would be a separate discussion (its `[[]]` matches `str.split` semantics).

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -1618,7 +1618,9 @@ def split_before(iterable, pred, maxsplit=-1):
         [[0, 1, 2], [3, 4, 5], [6, 7, 8, 9]]
     """
     if maxsplit == 0:
-        yield list(iterable)
+        buf = list(iterable)
+        if buf:
+            yield buf
         return
 
     buf = []
@@ -1654,7 +1656,9 @@ def split_after(iterable, pred, maxsplit=-1):
 
     """
     if maxsplit == 0:
-        yield list(iterable)
+        buf = list(iterable)
+        if buf:
+            yield buf
         return
 
     buf = []
@@ -1694,7 +1698,9 @@ def split_when(iterable, pred, maxsplit=-1):
 
     """
     if maxsplit == 0:
-        yield list(iterable)
+        buf = list(iterable)
+        if buf:
+            yield buf
         return
 
     it = iter(iterable)
--- a/tests/test_more.py
+++ b/tests/test_more.py
@@ -1550,6 +1550,13 @@ class SplitBeforeTest(TestCase):
         expected = []
         self.assertEqual(actual, expected)
 
+    def test_empty_collection_max_split_zero(self):
+        # maxsplit=0 took a fast path that yielded [[]] while every other
+        # maxsplit yielded [] for an empty iterable (issue #1252)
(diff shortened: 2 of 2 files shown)


seven7763:
Closing — this is the same fix as #1253 (opened a day earlier for the same issue), which also covers the changelog entry. Nothing to add on my side.

PR #1258, opened 2026-09-08
Fix None handling in combination_with_replacement_index

### LINKED ISSUE #1257: combination_with_replacement_index mistakes None values for exhaustion
`combination_with_replacement_index` can return the wrong index when the requested element contains `None`. It also accepts some combinations containing `None` that are absent from the iterable.

Reproduction on current master:

```python
from itertools import combinations_with_replacement
from more_itertools import combination_with_replacement_index

pool = [1, None, 2]
print(list(combinations_with_replacement(pool, 1)).index((None,)))  # 1
print(combination_with_replacement_index((None,), pool))          # 2

print(combination_with_replacement_index((None,), [1, 2]))         # 1
# Expected: ValueError, because (None,) is not a combination of [1, 2].
```

The outer scan treats `y is None` as exhaustion even when `None` is an actual, not-yet-matched element. Checking whether all element positions have been matched avoids confusing data with the exhaustion marker. A focused fix and regression tests are ready in a follow-up PR.

Reproduced on Windows with Python 3.12.

jokrasno:
### Issue reference

Fixes #1257.

### Changes

`combination_with_replacement_index((None,), [1, None, 2])` currently returns `2` instead of `1`. It can also return an index when the requested `None` value is absent from the pool, rather than raising `ValueError`.

Stop scanning the pool only after all element positions have matched, instead of using `None` as an exhaustion signal. Add regressions comparing results against `itertools.combinations_with_replacement`, including `None` at different positions, repeated values, iterator inputs, and invalid combinations.

### Checks and tests

- `python -B -m unittest tests.test_more.CombinationWithReplacementIndexTests -v` — all 10 tests pass. The two new regression tests produced 14 failing subcases against the original implementation.
- Ruff lint and formatting checks pass; `git diff --check` passes.
- Tested with Python 3.12 on Windows. Full `make all-checks` was not completed; no full-suite coverage, documentation, or packaging success is claimed.

### DIFF
--- a/more_itertools/more.py
+++ b/more_itertools/more.py
@@ -4533,7 +4533,7 @@ def combination_with_replacement_index(element, iterable):
                 break
             else:
                 k = tmp
-        if y is None:
+        if len(indexes) == l:
             break
     else:
         raise ValueError(
--- a/tests/test_more.py
+++ b/tests/test_more.py
@@ -5231,6 +5231,36 @@ class CombinationIndexTests(TestCase):
 
 
 class CombinationWithReplacementIndexTests(TestCase):
+    def test_none_values(self):
+        for iterable in (
+            [None, 1, 2],
+            [1, None, 2],
+            [1, 2, None],
+            [None, 1, None],
+        ):
+            for r in range(4):
+                first_index = {}
+                for index, element in enumerate(
+                    combinations_with_replacement(iterable, r)
+                ):
+                    with self.subTest(iterable=iterable, element=element):
+                        actual = mi.combination_with_replacement_index(
+                            iter(element), iter(iterable)
+                        )
+                        expected = first_index.setdefault(element, index)
+                        self.assertEqual(actual, expected)
+
+    def test_invalid_none_values(self):
+        for element, iterable in (
+            ((None,), [1, 2]),
+            ((1, None), [1, 2]),
+            ((1, None), [None, 1]),
+            ((None, 2), [1, None]),
+        ):
+            with self.subTest(element=element, iterable=iterable):
+                with self.assertRaises(ValueError):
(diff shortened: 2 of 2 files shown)


JamesParrott:
Hi Joshua, 

Thanks for this.  I can understand the general issue (without having delved into it).  

Is there some reason you haven't enabled the Github Actions workflows on your fork, so we can see the tests?  It's free for public repos and public forks.

Assuming it works, this PR looks like a really nice elegant fix, so well done.  

But I don't understand why `len(indexes) ==1` indicates the subsequence is found.  Personally, wouldn't it be more intuitive to use the normal way of avoiding similar `None`-related bugs - replacing the `None` in the `next` with a different (not-None) unique sentinel, e.g. `end = object()` (it's early days for `sentinel`, but if it gets back ported with a `__future__` import , that's perfect), and testing `y is end`?

Also doesn't changing the break condition to `len(indexes) ==1` break some subsequent code later on, populating `occupations`?  The intention there to me clearly feels like it's handing situations wher `indexes` could be longer than 1 (e.g. repeated elements?).

Or is the change more cunning than I've realised?  Am I missing something else?

JamesParrott:
Despite my quibble, this PR does [pass the test suite](https://github.com/JamesParrott/more-itertools/actions/runs/34214427637).

pochmann:
> But I don't understand why `len(indexes) ==1` indicates the subsequence is found.

And that's why the variable name `l` is bad :-)

JamesParrott:
> And that's why the variable name `l` is bad :-)

Yep.  This took me far too long!

jokrasno:
Closing this since #1261 includes the fix and my tests. Thanks for keeping the tests and crediting us both!

PR #1259, opened 2026-09-08
Fix zero and negative repeat in product helpers

jackwalkerlabs:
Product helpers currently accept `repeat=-1` as an empty product, and `repeat=0` exhausts supplied iterators even though `itertools.product` leaves them untouched. Validate repeat through the index protocol and skip materializing inputs for a zero repeat in `nth_product`, `product_index`, `gray_product`, and `partial_product`.

Regression tests failed on the base for all four helpers. Validation: all 914 unittest tests pass on CPython 3.12.14; Ruff lint and formatting pass. Tests also cover index-protocol objects and noninteger repeats.

Implemented and locally validated with OpenAI Codex.

Issue tracking: Found during source inspection; no matching open issue was identified for this fix.


rhettinger:
* The `index` calls are redundant. That is already checked by the `* repeat` code. The new tests `test_repeat_supports_index_protocol` and `test_noninteger_repeat` already pass without the code being modified.

* The `repeat == 0` behavior for `itertools.product` is a non-guaranteed CPython implementation detail, one that wasn't specifically intended, that isn't tested, and that doesn't match the pure-Python rough equivalent in the docs. I don't want to recreate that behavior because 1) it doesn't match the pure-Python equivalent, 2) because it optimizes a case that we don't care about, and 3) because it is nice that the current code raises a `TypeError` for non-iterable inputs.

That leaves only the `repeat < 0` check. That seems reasonable to me because it better matches a guaranteed behavior for `itertools.product`.

PR #1260, opened 2026-09-08
Validate running window sizes through the index protocol

jackwalkerlabs:
With `maxlen=2.5`, `running_min` and `running_max` never expire the oldest candidate and silently compute an unbounded result; `running_mean` silently uses a different window size. Integer-like objects implementing `__index__` also fail despite working with `running_median`.

Normalize `maxlen` through the index protocol in these functions and in `running_statistics`, which also needs integer `Stats.size` values. Add regressions for invalid floats, index-protocol objects, and nonpositive sizes, plus a changelog entry.

Validation: regressions fail before the change; all 913 unittest tests pass on CPython 3.12.14; Ruff lint and formatting pass.

Implemented and locally validated with OpenAI Codex.

Issue tracking: Found during source inspection; no matching open issue was identified for this fix.


rhettinger:
I don't think this is something we care about.  AIs frequently generate PRs to solve problems that people don't actually have.

Note the `recipes.pyi` file already requires a `int | None`, so people running static typing already get prodded.  If other people use `maxlen=2.0`, I don't feel like breaking their code for something so pedantic.

bbayles:
> Automatically generated pull requests that don't follow these instructions will be closed summarily.

more_itertools/recipes.py
```diff
@@ -1589,6 +1592,9 @@ def running_statistics(iterable, *, maxlen=None):
     but not complex numbers which are unorderable.
     """
 
+    if maxlen is not None:
```
rhettinger: `running_statistics` doesn't need a test because that is tested downstream in other calls.

Issue #1017, opened 2025-06-22
Faster `sliding_window`?


pochmann3:
For window size n up to 20, it uses a zip-of-tees iterator. Above that it uses a deque. I benchmarked a bit more on three systems using the [test case](https://github.com/more-itertools/more-itertools/issues/871) from the issue that introduced the tees solution. I also included a list solution which was always faster than the deque, up to three times faster. On these systems, the deque solution became faster than the tees solution around n=200, n=800, and n=100. At n=20, the tees solution was still quite a bit faster. They're all Linux systems, though, I don't have others.

@rhettinger showed times for n=400 with Python 3.13b3 on an Apple M1 Max, where the deque solution was still a bit slower:

```
 1.2481  'for w in sliding_window1(data, 400): pass'
 1.2173  'for w in sliding_window2(data, 400): pass'
```

Is the n=20 cutover point maybe too low? I'd be interested to see what this looks like on a system where the deque solution is already faster for n below 20, and what that system is. (Raymond [mentioned](https://github.com/more-itertools/more-itertools/issues/1015#issuecomment-2993900098) 20 was in the middle of the pack.)

And would the faster list solution be better than the deque solution? Or is the deque better because the maxlen feature makes it interesting and because that's what the recipe on the itertools page uses?

First system with Python 3.11.10:

![Image](https://github.com/user-attachments/assets/45fdf7e3-1c7d-48ba-adb9-61e0b3298b95)

Second system with Python 3.13.0:

![Image](https://github.com/user-attachments/assets/cb9bae96-b445-43db-8284-01c0f2a322af)

Third system with Python 3.13.4:

![Image](https://github.com/user-attachments/assets/b36c00ed-a0af-423f-a97c-15d16934b06d)

<details><summary>script with solutions, correctness checks, benchmarking, and plotting</summary>

```python
def with_deque(iterable, n):
    iterator = iter(iterable)
    window = deque(islice(iterator, n - 1), maxlen=n)
    for x in iterator:
        window.append(x)
        yield tuple(window)

def with_list(iterable, n):
    iterator = iter(iterable)
    window = list(islice(iterator, n - 1))
    for x in iterator:
        window.append(x)
        yield tuple(window)
        del window[0]

def with_tees(iterable, n):
    iterators = tee(iterable, n)
    for i, iterator in enumerate(iterators):
        next(islice(iterator, i, i), None)
    return zip(*iterators)

funcs = [
    with_deque,
    with_list,
    with_tees,
[...]

PR #1265, opened 2026-09-08
Use a list rather than a deque for large sliding_window sizes

LD-RW:
### Issue reference

more-itertools/more-itertools#1017

### Changes

`sliding_window` sends every window size above 20 to `_sliding_window_deque`. On CPython that is the slowest of the available strategies at every size I measured.

The deque updates in O(1) and the list's `del window[0]` is O(n), so the deque looks like the right structure. But the window is converted to a tuple once per iteration, and CPython has no fast deque-to-tuple path while list-to-tuple is a straight copy. That conversion dominates, and it more than pays for the O(n) delete. @pochmann3 identified this in #1017; the credit for the variant and the explanation is theirs.

This swaps the implementation and leaves the `n > 20` cutover exactly as it is. The test covering the long-window path was named `test_deque_version`, so it is renamed to `test_long_window_version` to match.

Measured through the public API, CPython 3.14.7, 100,000-item stream, min of 7 repeats:

| n | before | after | |
|---:|---:|---:|---|
| 3 | 2.91 ms | 2.95 ms | unchanged (tee path) |
| 10 | 5.57 ms | 5.04 ms | unchanged (tee path) |
| 20 | 9.20 ms | 8.83 ms | unchanged (tee path) |
| 21 | 20.63 ms | 8.24 ms | 2.50x |
| 25 | 22.83 ms | 8.35 ms | 2.73x |
| 50 | 38.11 ms | 13.59 ms | 2.80x |
| 100 | 66.12 ms | 18.93 ms | 3.49x |
| 400 | 190.29 ms | 58.41 ms | 3.26x |
| 800 | 342.37 ms | 113.25 ms | 3.02x |

Sizes at or below 20 are untouched - they take the same tee/zip path as before, and the small differences there are measurement noise.

### Checks and tests

- `make test` - 909 tests pass, 5 skipped
- Doctests - 145 pass
- `ruff format --check .` and `ruff check more_itertools tests` - clean
- Differential check: the old and new implementations produce identical output across 2,372 combinations of window size and input length, covering sequences, one-shot generators, and unhashable payloads

Two caveats I would rather state than leave for review to find:

**My benchmarks are CPython 3.14 only.** The reasoning rests on a CPython implementation detail (no `PyDeque_AsTuple` shortcut), and CI also runs pypy-3.10, which I could not test locally. If PyPy behaviour matters here I am happy to hold this until someone can check, or to adjust.

**I did not touch the `n > 20` cutover.** My numbers suggest the list variant also beats tee/zip from around n=8-10, so the threshold may be a little high, but that is a separate judgement call and I would rather raise it in #1017 than fold it into this change.


bbayles:
Apologies; I mixed this PR up with another. Will review soon!

bbayles:
I think Raymond's comment here remains accurate: https://github.com/more-itertools/more-itertools/issues/1017#issuecomment-3002043652

So we'll pass on this PR. Thanks for the contribution 🤖

Issue #1268, opened 2026-09-09
seekable(..., maxlen=0): peek() and bool() drop items


otiscuilei:
`seekable(..., maxlen=0)` is supposed to disable the history cache (`elements()` stays empty) while still iterating the source. `peek()` and `bool()` (which calls `peek()`) currently consume an item that they cannot put back, because a `deque(maxlen=0)` discards on append.

```python
from more_itertools import seekable

s = seekable([10, 20, 30], maxlen=0)
s.peek()
list(s)  # [20, 30]  — 10 is gone

s = seekable([10, 20, 30], maxlen=0)
bool(s)  # True
list(s)  # [20, 30]  — 10 is gone
```

Default `maxlen=None` is fine (`peek` is stable; `bool` does not consume). `test_maxlen_zero` only checks `list(s)` and does not cover peek/truthiness.

Happy to send a PR.

PR #1269, opened 2026-09-09
Issue 1268: seekable.peek/bool no longer drop items when maxlen is 0

otiscuilei:
more-itertools/more-itertools#1268

### Changes

`seekable.peek()` advances with `next(self)` then rewinds `_index` into `_cache`. With `maxlen=0` the cache is a `deque` that discards on append, so the rewind is a no-op and the peeked item is lost. `bool()` calls `peek()`, so `if s:` also drops the head.

When the cache cannot store the item, put it back on `_source` so the following `next()` returns it. Default `maxlen=None` is unchanged. `elements()` stays empty, which is the documented `maxlen=0` behaviour.

```python
from more_itertools import seekable

s = seekable([10, 20, 30], maxlen=0)
s.peek()
list(s)  # was [20, 30]; now [10, 20, 30]
```

### Checks and tests

- `python -m unittest tests.test_more.SeekableTest tests.test_more.PeekableTests` — OK
- `ruff check` / `ruff format --check` on the touched files — OK


more_itertools/more.py
```diff
@@ -3131,6 +3131,11 @@ def peek(self, default=_marker):
             if default is _marker:
                 raise
             return default
+        # maxlen=0 cannot store the item we just consumed, so put it back.
+        if isinstance(self._cache, deque) and self._cache.maxlen == 0:
```
bbayles, line 3135: Could you change this to `if getattr(self._cache, 'maxlen', None) == 0`? I prefer that to the type check.

bbayles (commented):
Thanks - I have one change request, then I'll merge this.

more_itertools/more.py
otiscuilei: Done — switched to `getattr(self._cache, 'maxlen', None) == 0` as requested.

PR #1274, opened 2026-09-11
Preserve source ValueErrors in monotonic extract

### LINKED ISSUE #1273: Monotonic extract replaces exceptions raised by its input iterator
`extract(..., monotonic=True)` converts a `ValueError` raised by the input iterator into `ValueError("Invalid index: ...")`, even when the index is valid. This also replaces subclasses such as a caller's domain-specific decoding exception, preventing callers from catching that exception type. The default buffered mode preserves the original error.

```python
from more_itertools import extract

class RecordError(ValueError):
    pass

def records():
    yield "first"
    raise RecordError("record decoding failed")

for monotonic in (False, True):
    try:
        list(extract(records(), [1], monotonic=monotonic))
    except ValueError as error:
        print(monotonic, type(error).__name__, str(error))
```

Actual output:

```text
False RecordError record decoding failed
True ValueError Invalid index: 1
```

Expected: both modes propagate the original `RecordError`. Index `1` is valid; the data source fails while producing that item.

Reproduced on Python 3.12.10 / Windows against current master `9ed3dbb0ae527230cd156d91d0af305478558fba` (11.1.0). The `ValueError` handler in `_extract_monotonic` covers both construction of `islice` and advancing the source; only argument-validation failures need that translation.

Prepared with OpenAI Codex assistance; the reproduction was run locally.

Gonghan-Princess:
### Issue reference

Fixes #1273.

### Changes

Construct the slice inside the index-validation handler, then advance it outside that handler. This keeps errors from the input iterator intact while retaining the existing handling of repeated indices, invalid indices, and exhaustion.

The regression test checks exception identity for plain `ValueError` and a subclass, including errors on the first requested item, during a skip, after a successful value, and after a repeated index. It covers both extraction modes; all eight monotonic-mode cases fail before the fix. Added an unreleased changelog entry.

### Checks and tests

Verified on Windows with Python 3.12.10 using the checks from `make all-checks`:

- 915 unit tests pass; package coverage is 100% (99% required).
- Ruff formatting and lint pass; stubtest passes for both modules.
- Sphinx builds with warnings treated as errors.
- Flit builds the wheel and sdist; both pass `twine check`.

Prepared with OpenAI Codex assistance for investigation, implementation, tests, and drafting. All reported checks were run locally.


rhettinger:
Perhaps just remove the `from None` so that upstream ValueErrors are visible.

rhettinger:
IMO this doesn't add any real value.  A `ValueError` is still raised. Only the error message is changed. Perhaps just remove the `from None` so that the whole chain is displayed.

bbayles:
I'm going to try tweaking the README to require descriptions of production issues for automated issues and PRs to see if it cuts down on this stuff. A few useful suggestions, but mostly a waste of time.

PR #1278, opened 2026-09-12
Reuse input iterators in zip_broadcast

### LINKED ISSUE #1277: zip_broadcast opens input iterables twice
`zip_broadcast()` calls `iter(obj)` to decide whether an argument is a scalar, discards that iterator, then calls `iter(obj)` again to consume the input. This opens an iterable twice. An input that permits only one opening works with `zip()` but fails with `zip_broadcast()`:

```python
from more_itertools import zip_broadcast

class SingleUseIterable:
    def __init__(self):
        self.opened = False

    def __iter__(self):
        if self.opened:
            raise RuntimeError('input already opened')
        self.opened = True
        return iter([1, 2])

print(list(zip(SingleUseIterable(), ['label', 'label'])))
# [(1, 'label'), (2, 'label')]

print(list(zip_broadcast(SingleUseIterable(), 'label')))
# RuntimeError: input already opened
```

Expected: `[(1, 'label'), (2, 'label')]`, without reopening the source. This also happens with empty inputs and with `strict=True`.

Reproduced on Linux with CPython 3.14.2 at commit `9ed3dbb0ae527230cd156d91d0af305478558fba`. Retaining the iterator obtained during scalar detection fixes the reproduction while preserving the configured scalar types and strict length checks. A patch with regression coverage is ready.

vitalivo:
### Issue reference

Fixes #1277.

### Changes

`zip_broadcast()` currently discards the iterator obtained while detecting scalars, then opens the iterable again. Inputs that allow only one opening fail with `RuntimeError` even though they work with `zip()`.

Reuse that first iterator. Scalar detection and strict length checks retain their existing behavior. The regression test covers empty and non-empty single-use inputs with both values of `strict`.

### Checks and tests

All four regression subcases fail on the original code and pass with the fix. `make all-checks` passes on Linux with CPython 3.14.2: 916 tests run, 5 skipped, 100% measured coverage; formatting, lint, stubtest, documentation, and package checks pass. `git diff --check` passes as well.


rhettinger:
I don't know if the problem being solved is a real-world problem, but there is no downside to making this edit.

+1

PR #1241, opened 2026-08-29
fix `one` to respect custom error handlers

andjf:
### Issue reference
resolves more-itertools/more-itertools#1240

### Changes
Fixes a bug in `one()` where custom exception handlers (`too_long` and `too_short`) were not always respected due to the use of the boolean `or` fallback. 

Specifically, this patch:
* Prevents a crash caused by eagerly evaluating the `__repr__` of elements to build the default `ValueError` message when a custom `too_long` exception has already been provided.
* Fixes an issue where custom exceptions that evaluate to `False` (via a custom `__bool__` method) would be bypassed, incorrectly raising the default `ValueError` instead. 
* Adds test coverage for both of these edge cases (items with a broken/unsupported `__repr__`, and exception classes with `__bool__` returning `False`).

### Checks and tests
Checks ran successfully

<details>
<summary>full <code>make all-checks</code> output</summary>

```
(.venv) @andjf ➜ /workspaces/more-itertools (master) $ make all-checks
python -m pip install --upgrade -r requirements/development.txt
Requirement already satisfied: coverage in ./.venv/lib/python3.12/site-packages (from -r /workspaces/more-itertools/requirements/testing.txt (line 1)) (7.16.0)
Requirement already satisfied: ruff in ./.venv/lib/python3.12/site-packages (from -r /workspaces/more-itertools/requirements/testing.txt (line 2)) (0.16.5)
Requirement already satisfied: mypy in ./.venv/lib/python3.12/site-packages (from -r /workspaces/more-itertools/requirements/typechecks.txt (line 1)) (2.3.1)
Requirement already satisfied: Sphinx in ./.venv/lib/python3.12/site-packages (from -r /workspaces/more-itertools/docs/requirements.txt (line 1)) (9.1.0)
Requirement already satisfied: furo>=2025.07.19 in ./.venv/lib/python3.12/site-packages (from -r /workspaces/more-itertools/docs/requirements.txt (line 2)) (2025.12.19)
Requirement already satisfied: flit<5,>=4.0.2 in ./.venv/lib/python3.12/site-packages (from -r /workspaces/more-itertools/requirements/packaging.txt (line 1)) (4.0.2)
Requirement already satisfied: flit_core<5,>=4.0.2 in ./.venv/lib/python3.12/site-packages (from -r /workspaces/more-itertools/requirements/packaging.txt (line 2)) (4.0.2)
[...]

Issue #1240, opened 2026-08-29
`one` handling of custom opinionated exceptions and objects


andjf:
`one()` fails when:
- certain custom exception objects are passed for `too_short` or `too_long`
- the objects being iterated have faulty `__repr__`s

Examples (tested with `more-itertools-11.1.0`):
```py
from itertools import count
from more_itertools import one

class OpinionatedError(Exception):
    def __bool__(self):
        return False

class OpinionatedObject:
    def __repr__(self):
        raise NotImplementedError("No representation without taxation!")

it = count()
one(it, too_long=OpinionatedError())
# raises the following: ValueError: Expected exactly one item in iterable, but got 0, 1, and perhaps more.
# desired behavior: raise the OpinionatedError

it = []
one(it, too_short=OpinionatedError())
# raises the following: ValueError: too few items in iterable (expected 1)
# desired behavior: raise the OpinionatedError

it = (OpinionatedObject() for _ in count())
one(it, too_long=OverflowError)
# raises the following: NotImplementedError: No representation without taxation!
# desired behavior: raise the OverflowError
```

Additionally, if an object's `__repr__` was somehow computationally expensive (which it really shouldn't be), `one()` currently pays the penalty of calling that `__repr__` even if `too_long` was provided.

I acknowledge that this is an extreme and weird edge case

bbayles:
> I acknowledge that this is an extreme and weird edge case

Indeed, I agree. To extreme and weird to handle.

andjf:
That's fair.

Just out of curiosity for future contributions (to this project and others): I assumed handling these types of edge cases was desirable since the fix is small and doesn't add real complexity. Since custom exceptions with `__bool__` and objects with custom `__repr__` are fully valid, is the concern mostly with accommodating those niche behaviors, or do you see these scenarios as being out of scope of the function's responsibility in the first place?

JamesParrott:
My 2p.  Python is so powerful, that it is always possible to find an edge case.  At some point, edge cases like these, are more about hacking the library, than they are about reflecting realistic use cases.

PR #1279, opened 2026-09-13
Fix one() and only() dropping a falsy user-supplied exception

itzzdev09:
### Summary
`one(it, too_long=exc)` and `only(it, too_long=exc)` used `raise too_long or ValueError(msg)` (and similar for `too_short`), which caused any custom exception instance whose boolean evaluation is falsy (e.g. an exception class defining `__bool__` returning `False`) to be discarded, erroneously raising the default `ValueError`.

### Changes
- Check `if too_long is not None:` and `if too_short is not None:` instead of truthiness.
- Only construct the default exception message when no custom exception is supplied, avoiding unnecessary string formatting or potential errors in item `__repr__`.
- Add unit tests verifying custom falsy exceptions are preserved and raised as expected.


rhettinger:
+1 This fixes an anti-pattern using truthiness to check for default arguments instead of checking for object identity.

Even though the fix should be made for clean coding, I do wonder whether this scenario ever arises in practice. It isn't normal (or Pythonic) exceptions to be false.

itzzdev09:
Thanks! Agreed it's rare. The realistic way it happens is by accident rather than on purpose: an exception class that defines __len__ (or __bool__) for some other reason, like a validation error that carries a list of problems, and gets raised with that list empty:

```
class ValidationError(ValueError):
    def __init__(self, errors=()):
        super().__init__(errors)
        self.errors = list(errors)
    def __len__(self):
        return len(self.errors)
```

one([], too_short=ValidationError())  # previously raised the generic ValueError instead

Mostly it's just the identity-vs-truthiness cleanup though, as you said.

andjf:
@rhettinger @bbayles Out of curiosity, how is this different than https://github.com/more-itertools/more-itertools/pull/1241 (issue https://github.com/more-itertools/more-itertools/issues/1240)?

bbayles:
It's the same, but Raymond argued persuasively that the PR fixes an anti-pattern. I will credit you in the release notes.

PR #1281, opened 2026-09-13
Preserve empty columns in sort_together

### LINKED ISSUE #1280: sort_together drops output columns when an input is empty
### Problem

`sort_together()` drops all output columns when the shortest input column is empty. This means a caller unpacking a fixed number of aligned columns fails on empty data, even though the number of input columns has not changed.

Reproduced on current master, `9ed3dbb0ae527230cd156d91d0af305478558fba` (11.1.0), using CPython 3.14.4:

```python
from more_itertools import sort_together

print(sort_together([[], []]))              # []
print(sort_together([[], []], strict=True)) # []
print(sort_together([[], [1]]))             # []

xs, ys = sort_together([[], []])
# ValueError: not enough values to unpack (expected 2, got 0)
```

I would expect `[(), ()]` for each of the three calls above: one sorted (possibly truncated) tuple per input column. With no input columns, `sort_together([])` should still return `[]`. With `strict=True`, unequal lengths such as `[[], [1]]` should still raise `ValueError`.

### Cause and proposed fix

The first transpose, `zip(*iterables, strict=strict)`, has no rows. After sorting, the inverse transpose is `zip(*[])`, which cannot recover the original number of columns.

The proposed fix retains the outer iterable as a tuple and returns one empty tuple per column when there are no sorted rows. The strict length check is still evaluated before this fallback; individual input columns are not materialized just to count them.

I searched the issue and PR history for `sort_together` and empty columns and did not find a matching report or fix. This was found during source inspection, not an incident in a production system. Investigation, patch preparation, and this report were AI-assisted.

rastagan-git:
### Issue reference

Fixes #1280.

### Changes

Preserve one output tuple per input column when `sort_together()` has no rows to sort. For example, `sort_together([[], []])` now returns `[(), ()]`, so fixed-column unpacking also works on empty data.

- Retain the outer iterable as a tuple to remember its column count, including when the outer input is an iterator.
- Apply the empty-row fallback after `sorted(zip(..., strict=strict))`, preserving strict-mode errors for unequal lengths.
- Add an empty-column docstring example and regression tests for zero through three columns, both strictness modes, an empty shortest column in either position, and nested one-shot iterators.

Nonempty sorting, key functions, reverse ordering, and the no-column result are unchanged. The intentional compatibility change is that one or more columns with zero sorted rows return a list of empty tuples instead of `[]`. The retained outer tuple uses O(number of columns) references; no new eager copy of the individual columns is made. No performance speedup is claimed.

### Checks and tests

The new tests produced eight failing subtests on the unmodified implementation. After the fix:

- Focused `SortTogetherTest`: 9 tests pass.
- CPython 3.14.4: full suite, 918 tests run, 5 skipped; 100% line coverage (2,213 statements).
- CPython 3.10.0 and 3.11.9: full suite, 917 tests pass on each.
- Ruff whole-repository format check and source/test lint pass.
- `stubtest more_itertools.more more_itertools.recipes` passes.
- `sphinx-build -W -b html docs docs/_build/html` passes.
- Flit wheel/sdist build and Twine checks pass.
- `git diff --check` passes.

The `make all-checks` targets were executed through their equivalent native commands on Windows. Python 3.12, 3.13, PyPy, and remote CI have not been independently validated here.

AI assistance was used for source inspection, implementation, tests, and this description.


bbayles:
New guideline for coding assistants: issues need to be real things encountered in actual code.

Issue #1282, opened 2026-09-16
iequals() masks ValueError exceptions raised while comparing or iterating


AbhiramMandala:
`iequals()` currently wraps its comparison in `try/except ValueError` and returns `False` for *any* `ValueError` — not just the "iterables differ in length" `ValueError` raised by `zip(..., strict=True)`. As a result it silently swallows `ValueError`s that genuinely originate from the input iterables, which is incorrect.

### Reproducible example

Case 1: a `ValueError` raised while comparing elements is swallowed:

```pycon
>>> from more_itertools import iequals

>>> class BadComparer:
...     def __eq__(self, other):
...         raise ValueError('comparison failed')

>>> iequals(iter([BadComparer()]), iter([BadComparer()]))
False
```

Case 2: a `ValueError` raised while iterating is swallowed:

```pycon
>>> def second_item_raises():
...     yield 1
...     raise ValueError('iteration failed')

>>> iequals(second_item_raises(), [1])
False
```

A plain equality check on the same data raises, which is the behavior users expect:

```pycon
>>> [BadComparer()] == [BadComparer()]
Traceback (most recent call last):
  ...
ValueError: comparison failed
```

### Expected behavior

`iequals()` should return `False` only when the iterables actually differ — i.e. they have different lengths or a pair of elements compares unequal. A `ValueError` raised by the iterables themselves (during iteration or comparison) should propagate, not be masked.

### Root cause

In `more_itertools/more.py`:

```python
try:
    return all(map(all_equal, zip(*iterables, strict=True)))
except ValueError:
    return False
```

The `except ValueError` cannot distinguish the length-mismatch `ValueError` from `zip(strict=True)` from unrelated `ValueError`s.

This is the same class of error-masking bug that has recently been fixed in `value_chain` (#1251) and `monotonic extract` (#1273). There is currently no open issue or PR covering `iequals`.

Note: the investigation, reproduction, and this report were prepared with the assistance of an AI coding agent.

bbayles:
I'm inclined to think this is a non-issue. If you're comparing two sequences, and one of them raises a `ValueError`, were they equal? I say `False` is a fine response.

I would be open to a note in the docstring that says this function swallows `ValueError`.

It would be nice if CPython `zip` raised a more specific exception, though.

rhettinger:
[...]

Issue #1275, opened 2026-09-12
numeric_range reverse iteration changes float values and overflows at datetime limits


emme1t:
`reversed(numeric_range(...))` can change the values and length of a floating-point range. It also raises `OverflowError` for valid ranges that start at a datetime limit.

Reproduced on Python 3.13.13 / Windows with master at `9ed3dbb0ae527230cd156d91d0af305478558fba`:

```python
from more_itertools import numeric_range

values = numeric_range(0.0, 1.0, 0.1)
forward = list(values)
backward = list(reversed(values))

print(len(forward), len(backward))
# 10 11

print(backward[-1])
# -0.09999999999999998

assert backward == forward[::-1]  # fails
```

Reverse iteration should produce exactly the original ten values in reverse order. The additional negative value lies outside the range, and several of the other values differ from those produced by forward iteration.

A related endpoint problem appears with datetimes:

```python
from datetime import datetime, timedelta
from more_itertools import numeric_range

values = numeric_range(
    datetime.min,
    datetime.min + timedelta(days=2),
    timedelta(days=1),
)
print(list(values))
# [datetime.datetime(1, 1, 1, 0, 0), datetime.datetime(1, 1, 2, 0, 0)]

print(list(reversed(values)))
# OverflowError: date value out of range
```

The expected reverse sequence contains January 2 followed by January 1. Descending ranges starting at `datetime.max` have the corresponding overflow.

`__reversed__` currently constructs a new range using the final value as its start and `self._start - self._step` as its stop. That arithmetic can change floating-point values and move the stop beyond datetime's representable limits. Reading the original range's indexes in descending order appears to address both cases while preserving lazy iteration.

This report was prepared and reproduced with OpenAI Codex at the account owner's request.

pochmann:
Your fix looks alright, but out of curiosity and for "completeness", here are some alternatives and a benchmark.

### Your version "emme1t"

```python
    def __reversed__(self):
        return map(self._get_by_index, reversed(range(self._len)))
```

### Alternative "None": Delete the current `__reversed__`

Then `reversed` will create an iterator using the sequence protocol. That's pretty much the same as what your version does, except you skip `__getitem__` and use `_get_by_index` directly.

### Alternative "pochmann_1"

Same as yours but just directly using the "start + i * step" from `_get_by_index`.

```python
    def __reversed__(self):
        for i in reversed(range(self._len)):
            yield self._start + i * self._step
```

### Alternative "pochmann_2"

[...]

PR #1283, opened 2026-09-16
Fix iequals() masking ValueError exceptions raised by input iterables

AbhiramMandala:
### Issue reference
more-itertools/more-itertools#1282

### Changes
- `iequals()` now uses `zip_longest` with an identity-based sentinel instead of `zip(strict=True)` with a broad `except ValueError`. This distinguishes the length-mismatch `ValueError` (return `False`) from unrelated `ValueError`s raised by the input iterables (propagate the exception), following the same pattern as the `value_chain` fix in #1251 and the `monotonic extract` fix in #1273.
- Added regression test `test_value_error_not_masked` covering both comparison `ValueError`s and iteration `ValueError`s.

### Checks and tests
Ran `make all-checks` locally: `ruff format --check`, `ruff check`, `stubtest`, `coverage` (100%, 921 tests), `sphinx-build -W`, `flit build` + `twine check` all pass.


AbhiramMandala:
## Verification Report

**PR:** https://github.com/more-itertools/more-itertools/pull/1283  
**HEAD:** `9c79c89`  
**Issue:** https://github.com/more-itertools/more-itertools/issues/1282  
**State:** OPEN, MERGEABLE  
**Files changed:** 2 (`more_itertools/more.py`, `tests/test_more.py`)  
**Net diff:** +26 / −4

---

### 1. Issue and problem
`iequals()` wrapped all comparison logic in `try/except ValueError`, returning `False` for any `ValueError` — not only the length-mismatch error raised by `zip(strict=True)`. This masked `ValueError`s originating from the input iterables during iteration or comparison.

**Before (current master):**
```pycon
>>> class BadComparer:
...     def __eq__(self, other):
...         raise ValueError('comparison failed')
>>> iequals(iter([BadComparer()]), iter([BadComparer()]))
False  # masked
```

### 2. Root cause
The `except ValueError: return False` block cannot distinguish zip's length-mismatch error from unrelated errors inside `all_equal`/`all()`.

### 3. Fix summary
Replaced the `try/except ValueError` with a `zip_longest` + identity-based sentinel approach: only the sentinel check distinguishes length mismatches (return `False`) from real exceptions (propagate). This matches the same error-unmasking pattern applied to `value_chain` in #1251 and `extract` in #1273.

The `object()` sentinel cannot be faked by user objects (identity check, not equality), so the `mock.ANY` fakeability edge case from issue #900 is also preserved.

### 4. Regression test added
`test_value_error_not_masked` in `IequalsTests` covers both cases:
- `ValueError` raised by `__eq__` → must propagate
- `ValueError` raised mid-iteration → must propagate

### 5. Reproducing the fix
```
git checkout AbhiramMandala:fix-iequals-valueerror-masking
```
```python
>>> from more_itertools import iequals
>>> class BadComparer:
...     def __eq__(self, other):
...         raise ValueError('comparison failed')
>>> iequals(iter([BadComparer()]), iter([BadComparer()]))
# raises ValueError('comparison failed')
```

### 6. Full test suite
```
python -m coverage run --include="more_itertools/*.py" -m unittest
# 921 tests, 0 failures, 100% coverage
```

### 7. Checks and tests (all pass locally)
| Check | Status |
|---|---|
| `ruff format --check .` | 13 files already formatted |
| `ruff check more_itertools tests` | All checks passed |
| `stubtest more_itertools.more more_itertools.recipes` | Success: no issues |
| `coverage --fail-under=99` | 100% (2219 stmts, 0 miss) |
| `sphinx-build -W` | build succeeded |
| `flit build` + `twine check` | PASSED |

### 8. Implementation review
- No new dependencies introduced
- No `.pyi` signature changes required (public API unchanged)
- No documentation changes needed (bug fix only, semantics now match user expectation)
- `zip_longest` already imported at line 25; no new imports
- No functional change to correct usage (0-length mismatch → `False`, equal iterables → `True`)

### 9. Commits in this branch
- `9c79c89` — Don't mask ValueError raised while comparing in iequals

### 10. Outstanding notes
- All four current open bug issues (#1280, #1275, #1273, #1268) remain unmerged; this PR does not conflict
- No changelog file is updated; maintainer to advise on target version

PR #1285, opened 2026-09-17
Issue 1284: don't let a bucket lookup invent a key

### LINKED ISSUE #1284: bucket: a failed lookup or `in` test adds a phantom key to the keys returned by iteration
### Description

Testing a key against a `bucket` — with `in`, or by selecting a bucket that turns out to be empty — permanently adds that key to the keys returned by iterating the bucket. The key is reported even though no item ever had it, and even though `in` says it is not there.

```python
>>> from more_itertools import bucket
>>> s = bucket(['a1', 'b1', 'a2'], key=lambda x: x[0])
>>> 'c' in s
False
>>> sorted(s)
['a', 'b', 'c']          # expected ['a', 'b']
>>> list(s['c'])
[]
```

So `'c' in s` is `False` while `'c' in list(s)` is `True`, for the same object at the same moment.

Selecting the bucket directly does it too, with no membership test involved:

```python
>>> s = bucket(['a1', 'b1'], key=lambda x: x[0])
>>> list(s['zzz'])
[]
>>> sorted(s)
['a', 'b', 'zzz']
```

A `validator` does not help — it only filters which keys can be invented:

```python
>>> s = bucket(['a1', 'b1'], key=lambda x: x[0], validator=lambda k: k in {'a', 'b', 'c'})
>>> 'c' in s
False
>>> sorted(s)
['a', 'b', 'c']
```

The docstring's own example is affected. `sorted(list(s))` is documented as "Get the keys", and one membership test changes the answer:

```python
>>> iterable = ['a1', 'b1', 'c1', 'a2', 'b2', 'c2', 'b3']
>>> s = bucket(iterable, key=lambda x: x[0])
>>> 'd' in s
False
>>> sorted(list(s))
['a', 'b', 'c', 'd']     # documented as ['a', 'b', 'c']
```

Each miss is sticky and they accumulate, so a lookup-heavy loop can leave the key list mostly phantoms:

```python
>>> s = bucket(['a1'], key=lambda x: x[0])
>>> for k in ('q', 'r', 's'):
...     _ = k in s
>>> sorted(s)
['a', 'q', 'r', 's']
>>> {k: list(s[k]) for k in sorted(s)}
{'a': [], 'q': [], 'r': [], 's': []}
```

### Cause

`_cache` is a `defaultdict(deque)`, and `_get_values` begins by *reading* it:

```python
def _get_values(self, value):
    while True:
        # If we've cached some items that match the target value, emit
        # the first one and evict it from the cache.
        if self._cache[value]:          # <- a read that creates the entry
```

On a miss that read inserts an empty `deque` under `value`. `__iter__` then reports whatever keys `_cache` holds:

```python
return iter(self._cache)
```

so the empty entry is indistinguishable from a real key. `__contains__` routes through `_get_values` as well, which is why a plain `in` test is enough to trigger it.

`__iter__` was added later than the cache (#371, for #370), so the two have simply never agreed on what a key is.

### Expected behaviour

Iterating a `bucket` should yield only keys that at least one item actually had. `k in s` and `k in list(s)` should agree.

I have a fix and tests ready and will open a PR referencing this issue. It makes `_cache` a plain `dict` so a lookup cannot insert, and registers a key only where an item is actually seen. The existing `test_list` case — where a key whose items were fully consumed through `s[key]` must still be listed — keeps working, since that key is registered when its first item is produced rather than by the lookup itself.

Found by machine analysis while probing for invariant mismatches, not by hitting it in production.

onk3sh:
### Issue reference

Fixes #1284.

### Description

Testing a key against a `bucket` permanently added that key to the keys returned by iterating it, so `'c' in s` and `'c' in list(s)` disagreed for the same object at the same moment:

```python
>>> s = bucket(['a1', 'b1', 'a2'], key=lambda x: x[0])
>>> 'c' in s
False
>>> sorted(s)
['a', 'b', 'c']          # expected ['a', 'b']
```

`_cache` was a `defaultdict(deque)` and `_get_values` opened by *reading* it:

```python
if self._cache[value]:      # inserts an empty deque when value is a miss
```

`__iter__` reports `iter(self._cache)`, so the inserted entry was indistinguishable from a real key, and `__contains__` routes through `_get_values`, so a plain `in` test created one. Misses were sticky and accumulated. The docstring's own "Get the keys" example returned `['a', 'b', 'c', 'd']` after a single `'d' in s`.

### The change

`_cache` becomes a plain `dict`, so a lookup cannot insert, and a key is registered only where an item with that key is actually seen — `setdefault` on the two paths that store an item, plus one on the path that hands an item straight to the caller.

That last one is the case worth flagging. When `_get_values` finds a match it yields the item without ever putting it in the cache, so without an explicit registration the key would go unreported — which would break the existing `test_list`, where all three buckets are drained through `s[key]` first and `set(D)` must still be `{10, 20, 30}`. Registering at the point the first matching item is produced keeps that behaviour: the key is recorded because an item had it, not because someone asked about it.

The `defaultdict` import stays; it is used by three other functions.

### Testing

- `python -m unittest discover` — 924 tests, OK (5 skipped).
- doctests: `more.py` 588 attempted / 0 failed, `recipes.py` 145 / 0.

Four tests added to `BucketTests`, all confirmed to fail without the `more.py` change and pass with it:

```
FAIL: test_in_does_not_add_key
FAIL: test_getitem_miss_does_not_add_key
FAIL: test_in_does_not_add_key_with_validator
FAIL: test_keys_are_only_those_seen
Ran 9 tests — FAILED (failures=4)
```

No change to `more.pyi`; the stub does not describe `_cache`.

I have no strong view on whether the empty-lookup behaviour should also be mentioned in the docstring — the existing note about exhausting the iterable is still accurate — so I have left the prose alone. Happy to add a line if you would like one.


bbayles (approved):

bbayles:
Thanks for the report and the fix - this will go in the next release.

PR #1286, opened 2026-09-17
Fix sort_together dropping output columns when an input is empty

cristianchiriac:
more-itertools/more-itertools#1280

### Changes

`sort_together()` drops all output columns when the shortest input column is empty, so a caller unpacking a fixed number of aligned columns fails even though the input column count didn't change:

```python
>>> sort_together([[], []])
[]
>>> xs, ys = sort_together([[], []])
ValueError: not enough values to unpack (expected 2, got 0)
```

The cause: the first transpose, `zip(*iterables, strict=strict)`, produces no rows when the shortest input is empty. After sorting, `reordered` is `[]`, and `zip(*reordered, strict=strict)` (i.e. `zip(strict=strict)`) has no way to recover the original number of columns — it just returns nothing.

The fix keeps the original `iterables` around as a `tuple` (needed anyway since it's consumed once by `zip(*iterables, ...)`), and returns `[()] * len(iterables)` when there are no sorted rows to transpose back, instead of `[]`. The `strict` length check is unaffected — it's still evaluated while building `transposed`/`reordered`, before the empty-rows fallback, so `sort_together([[], [1]], strict=True)` still raises `ValueError`.

### Checks and tests

- Added `test_empty_shortest_column` to `SortTogetherTest` in `tests/test_more.py`, covering `sort_together([[], []])`, `strict=True` with equal empty lengths, `sort_together([[], [1]])`, `sort_together([])`, and that `strict=True` with unequal lengths (one empty, one not) still raises.
- Verified the new test fails against the unmodified code and passes with the fix.
- `python -m unittest discover -s tests`: 926 tests, all passing.
- `ruff format --check .` and `ruff check more_itertools tests`: clean.
- `stubtest more_itertools.more more_itertools.recipes`: no issues.


pochmann:
There's already a PR from the issue author, so what's the reason for your PR?

Tatamis:
Fair question — #1281 came first (Sep 13) and lands essentially the same fix: `tuple(iterables)` plus the `if not reordered: return [()] * len(iterables)` fallback. I opened this independently from the issue without spotting it, so the only real difference is test layout, and #1281 additionally documents the behaviour in the `sort_together` docstring.

I don't think there's anything here worth keeping over #1281. Happy to close this in its favour — or feel free to close it yourself.

PR #1288, opened 2026-09-17
Issue #1287: Less aggressive fix for issue #1284

### LINKED ISSUE #1287: Do you really want setdefault calls in the bucket class?
ISTM that PR #1285 was too aggressive in fixing the spurious creation of cache keys by a `defaultdict`. It removed the `defaultdict` entirely and replaced it with a regular dict and `setdefault` calls.

In general, `setdefault` should be avoided because it greedily creates a new default value object on every call even if that object is not needed.  In this case, we get a new `deque` on every call.  If you really want a regular dict, the workarounds are to apply LBYL or EAFP logic instead of `setdefault`.

However, it would be cleaner and more performant to restore the `defaultdict` and let it do its job. For sections that need to avoid creating spurious keys, just make a `get` call.

rhettinger:
Restore `defaultdict`. Where needed, prevent spurious key creation with an LBYL check or call to `get`. Otherwise, let `defaultdict` do its job and avoid making unnecessary `deque` objects.


bbayles (approved):

bbayles:
This is better, yes - thanks.

PR #1289, opened 2026-09-18
fix: clear ValueError for ichunked(n < 1) on non-empty input

Pitchfork-and-Torch:
## Summary
`ichunked(non_empty, 0)` (and other `n < 1`) failed inside `itertools.islice` with a cryptic bounds error. Empty iterables historically yielded nothing for `n == 0`.

## Fix
Once a chunk would be produced, raise `ValueError('n must be at least one')`. Empty input still yields an empty iterator.

## Test plan
- [x] `IchunkedTests` including new `test_zero_nonempty_raises` (8 passed)


bbayles:
Closing for not following PR rules for coding assistants.

PR #1292, opened 2026-09-20
Support the integer index protocol in numeric_range indexing

### LINKED ISSUE #1291: numeric_range indexing rejects objects implementing __index__
`numeric_range` currently rejects objects implementing `__index__`, although ordinary Python sequences accept them as integer indices.

```python
from more_itertools import numeric_range

class Index:
    def __index__(self):
        return 1

print(range(4)[Index()])          # 1
print(numeric_range(4)[Index()])  # TypeError
```

The same mismatch affects negative and out-of-range custom indices. Converting non-slice keys with `operator.index` would preserve integer indexing semantics and reject floats/strings. I have a focused patch and regression test ready, with the full 929-test suite passing.

agammann:
Closed by the maintainer because the report did not include the required affected production-code example. The example in #1291 is synthetic. I have not established a production use case that meets the contribution requirements, and this PR remains closed.

### Issue reference

Fixes #1291.

### Changes

Use `operator.index` for non-slice keys so `numeric_range` accepts objects implementing `__index__`, like Python sequences do. Keep the existing error message for invalid keys and the existing slice path. Update the key annotation to `SupportsIndex` and add a changelog entry.

The regression test covers positive and negative custom indices, out-of-range indices, and invalid index values. Its six valid/out-of-range subtests fail on the original implementation.

### Original local checks and tests

- `python -m unittest`: 929 tests passed.
- Strict Sphinx HTML build passed.
- Ruff lint and formatting checks passed on changed Python files.
- Coverage run: 929 tests passed, 100% coverage (99% required).
- Full Ruff checks and stubtest passed.
- Wheel/source distribution build and Twine checks passed.

These are the original local results, not a passing fork GitHub Actions run. The integer-access performance question was not measured before closure; no performance conclusion is claimed.


more_itertools/more.py
```diff
@@ -2378,20 +2379,21 @@ def __eq__(self, other):
...
+        if isinstance(key, slice):
             start_idx, stop_idx, step_idx = key.indices(self._len)
             return numeric_range(
                 self._start + start_idx * self._step,
                 self._start + stop_idx * self._step,
                 self._step * step_idx,
             )
-        else:
+        try:
+            index = _index(key)
```
JamesParrott, line 2390: How about getting rid of the `_index` alias, and just using `idx` (as per the slice handling) or even `i` instead?

JamesParrott (changes requested):
Another formulaic bot PR.

The test workflows have not been run on the fork https://github.com/agammann/more-itertools/actions , and the author appears to me to be an enthusiastic user of AI, and another incredibly prolific creator of PRs on Github in general.

That said, although I'm not up to date with the latest developments to `numeric_range`, and I've never needed to implement  `__index__` on a class myself, this particular PR seems quite small and focussed (and the tests look good, if the author would care to run them in Github Actions themselves, instead of just effectively saying "Trust me bro").

pochmann:
How much slower does this make int accesses?

bbayles:
> * If you're using a coding assistant or other automated tool to identify bugs, issues you create need to show examples of production code is affected by the problem. Sanitized code is OK; contrived examples aren't.
> * Issues and pull requests from coding assistants or other automated tools that don't follow these instructions will be closed summarily.

Closing summarily.

